PR #117, opened 2026-04-27
Preserve SQL comments through pist fmt

winebarrel:
## Summary
- `pist fmt` keeps `--` and `/* */` comments that appear between top-level statements.
- Comments are attached to the entity that owns the surrounding region: a comment inside an entity's ownership range (its `CREATE` plus associated `CREATE INDEX` / `ALTER TABLE ADD CONSTRAINT` / `COMMENT ON`) becomes that entity's trailing; otherwise it becomes the next entity's leading.
- The auto-generated `-- public.foo` header is no longer emitted by `fmt` (dump output is unchanged).

## How it works
- `parser.ScanComments` uses `pg_query.Scan` to recover comment tokens (`SQL_COMMENT` / `C_COMMENT`) with byte offsets, since they are not present in the parse tree.
- `EntityRef` now tracks `Location` (start of the `CREATE` keyword), `StmtEnd`, and `OwnershipEnd`. Non-entity stmts (`CREATE INDEX`, `ALTER TABLE ADD ...`, `COMMENT ON ...`) extend their target entity's `OwnershipEnd`.
- `formatWithComments` walks entities in source order, emitting leading + bare SQL + trailing for each, plus any file-trailing comments at the end.
- `model.{Table,View,Enum,Domain}ToSQLBare` are new variants without the `-- FQN` header; `*ToSQL` still produces the header for `dump`.

## Limitations
- Comments inside a `CREATE` body (e.g. inline column comments) are dropped because pg_query discards them before producing the parse tree.
- Within an entity's ownership range, comments are emitted at the end of the entity block rather than at their original position between sub-statements.

## Test plan
- [x] `make test` (parser, model, diff, dump, apply, plan, root, toposort)
- [x] `make lint`
- [x] `bash test/scenario/fmt.test.sh`
- New fixtures: comments_basic, comments_block, comments_inline_dropped, comments_with_directives, comments_around_index, comments_around_alter, comments_around_comment_on, comments_around_column_comment, comments_around_matview, comments_multiple_children, comments_only, comments_each_kind
- New unit tests: ScanComments / CodeStart / filterTopLevelComments / commentStmtTargetFQN

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,9 @@ pist fmt --check schema.sql
 > [!NOTE]
 > `dump` output uses PostgreSQL's own formatting (e.g. `pg_get_viewdef`), while `fmt` normalizes through the pg_query parser. This means `dump` output may not pass `fmt --check` directly. Run `fmt -w` once after the initial `dump` to normalize, then use `--check` in CI going forward.
 
+> [!NOTE]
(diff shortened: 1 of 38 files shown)


winebarrel:
__

PR #124, opened 2026-04-29
Validate column references in desired schema at plan time

winebarrel:
## Summary
- Resolve unqualified column references in each table's index, constraint, and foreign-key definitions against the table's desired column set; report unresolved references as a single aggregated error (`errors.Join`) before any DDL runs.
- Catches the common mistake of renaming a column via `-- pist:renamed-from` while forgetting to update a dependent definition (previously surfaced only at apply time as `column "X" does not exist`).
- Scope intentionally limited to same-table references — FK `PkAttrs` (parent-table columns) and partition children are skipped, mirroring the rename-rewrite policy.

## Test plan
- [x] `make test` (parser unit tests + new error YAML fixtures across index / CHECK / UNIQUE / FK / EXCLUDE)
- [x] `make lint`
- [x] Coverage check — `walkExprColumnRefs` 100%, `ValidateColumnRefs` 100%; remaining sub-100% lines in helpers are paranoid pg_query AST guards skipped per CLAUDE.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Validate column references in desired schema at plan time: indexes, constraints (CHECK / UNIQUE / PK / EXCLUDE), and foreign keys (local side) whose definitions reference columns absent from the owning table's desired column set are reported as a single aggregated error before any DDL is executed. Catches the common mistake of renaming a column via `-- pist:renamed-from` while forgetting to update the dependent definition.
+
 ## [1.2.0] - 2026-04-29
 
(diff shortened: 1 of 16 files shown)


winebarrel:
Addressed both review comments in bc0c47c:

**Code duplication.** Extracted the shared pg_query helpers — `WalkExprColumnRefs`, `ParseConstraintDef` / `ParseConstraintDefStrict`, `DeparseConstraintDef`, and the `ALTER TABLE _t ADD CONSTRAINT _c` wrapper prefix — into a new `internal/pgast` package. Both `diff/rename.go` (rewrite) and `parser/validate.go` (collect) now go through the same walker and parse/deparse helpers via a visitor callback (`func(*pg_query.String)`). Direct unit tests for the package live in `internal/pgast/pgast_test.go`.

**README error example.** Fixed: `model.Ident` only quotes identifiers that aren't safe unquoted, so the actual error reads `column name referenced in index idx_users_name does not exist on table public.users` (no quotes for safe identifiers). Updated the README to match.

winebarrel:
Addressed both review comments in a76a90e:

1. **INHERITS children.** Switched the skip condition from `PartitionOf != nil && PartitionBound != nil` to just `PartitionOf != nil` so INHERITS-style children (PartitionBound nil) are also skipped. Added a regression test confirming a child with an empty `Columns` map and a `PRIMARY KEY (id)` constraint no longer reports a false positive.

2. **Duplicate errors.** Added a per-object `reportMissing` helper that dedupes missing column names while preserving first-encounter order, so `CHECK ((qty > 0 AND qty < 100))` with a missing `qty` produces a single error line instead of two. Test asserts the count is exactly 1.

PR #125, opened 2026-04-29
Add fixtures for partitions, generated columns, domain rename + GENERATED bug fixes

winebarrel:
## Summary
Fill plan/apply test gaps for several features that previously only had dump or unit-level coverage. Adding tests surfaced two real bugs in stored generated column handling, which are fixed in the same PR.

### Bug fixes (commit b4fc9cb)
- **Parser**: pg_query reports \`GeneratedWhen=\"a\"\` (ALWAYS keyword) for STORED generated columns, not \`\"s\"\`. The parser's \`== \"s\"\` check never matched, so emitted SQL used \`DEFAULT <expr>\` instead of \`GENERATED ALWAYS AS (<expr>) STORED\`.
- **Diff**: \`alterColumnSQL\` emitted \`ALTER COLUMN ... SET DEFAULT\` for no-op diffs on generated columns (the catalog and parsed-desired expressions differ by pg_get_expr-added type casts). PostgreSQL doesn't allow altering generated expressions via SET DEFAULT, so the default-diff branch is skipped when either side is a stored generated column.

### New fixtures (commit ee126bb)
- Partitioned tables: \`create_partitioned_table\`, \`add_partition_child\`, \`drop_partition_child\` (plan); apply for create.
- Generated columns: \`create_table_with_generated_column\`, \`add_generated_column\`, \`no_diff_generated_column\` (plan); apply for create.
- \`rename_domain\` (plan + apply) — previously only the error-case fixture existed.
- \`no_diff_unlogged_table\`, \`no_diff_deferrable_constraint\` to lock in round-trip behavior.

INHERITS plan/apply is intentionally deferred — \`model.Table.SQL\` drops the child's own columns in the INHERITS branch, which needs separate design work.

## Test plan
- [x] \`make test\` (full suite)
- [x] \`make lint\`


winebarrel:
Addressed both review comments in 9c935cf:

1. **GENERATED toggle no longer silently skips.** When a column is generated on one side and non-generated on the other, plan now errors out at parse time with `column public.products.total: cannot toggle GENERATED — DROP COLUMN + ADD COLUMN is required`. Added `testdata/plan/error_toggle_generated_column.yml` covering this case.

   Expression-only changes between two stored generated columns are still skipped — catalog renders the expression with pg_get_expr-added type casts (e.g. `price * (quantity)::numeric`) which don't reliably compare with the desired-side raw expression. TODO.md updated to reflect the remaining work.

2. **YAML quoting consistency.** Switched the three new no-diff fixtures from `plan: ''` to `plan: ""` to match the established convention.

PR #126, opened 2026-04-29
Update stale rename NOTE and expand TODO.md

winebarrel:
## Summary
- Reword the outdated NOTE on \`detectColumnRenames\` (it claimed dependents weren't updated, which hasn't been true since #123) and point at TODO.md for the genuine remaining gap.
- Add six new deferred items uncovered by a post-#125 audit:
  - \`Table.TableSpace\` / \`Index.TableSpace\` changes are silently dropped
  - \`Table.Unlogged\` toggles are silently dropped
  - \`Column.StorageType\` is declared but never read/written
  - \`Constraint.Deferrable\` / \`Deferred\` go through DROP+ADD instead of \`ALTER CONSTRAINT\`
  - Standalone sequences are not first-class (parser/diff don't handle \`CREATE/ALTER/DROP SEQUENCE\`)
  - Table rename: cross-table dependents (FK \`REFERENCES\` and view definitions in *other* tables)

Documentation only.

## Test plan
- [x] \`make test\`
- [x] \`make lint\`


winebarrel:
Addressed both review nits in b76129b:

1. Listed the three rewrite helpers individually with backticks (`rewriteColumnRefsInIndexes`, `rewriteColumnRefsInConstraints`, `rewriteColumnRefsInForeignKeys`) so the slash-separated abbreviation no longer reads as a single function name.

2. Spelled out `ALTER INDEX ... SET TABLESPACE <new>` explicitly in the tablespace TODO, mirroring the table example.

winebarrel:
Addressed all four review comments in 9c3869c:

1. Kept the `CREATE UNLOGGED TABLE` inline code span on a single line.
2. Kept the `FOREIGN KEY ... REFERENCES old_table(...)` inline code span on a single line.
3. **Corrected the deferrable DDL form.** I had written `ALTER CONSTRAINT ...` standalone, but PostgreSQL only supports it as `ALTER TABLE ... ALTER CONSTRAINT ... [NOT] DEFERRABLE [INITIALLY ...]`. Good catch — the original wording would have sent someone down the wrong path.
4. Switched "DROP/ADD" to "drop/recreate" in the column rename comment since indexes use DROP/CREATE and only constraints/FKs use DROP/ADD.

PR #140, opened 2026-05-04
Detect and emit DDL for IDENTITY column transitions

winebarrel:
## Summary

- `alterColumnSQL` previously ignored `ColumnIdentity`, so toggling a column between a plain column and `GENERATED ... AS IDENTITY` produced an empty plan while the schema actually drifted. This was asymmetric with the `GENERATED ... STORED` toggle path which errors out.
- Emit `ALTER TABLE` statements for the three identity transitions:
  - `none → identity`: `ADD GENERATED { ALWAYS | BY DEFAULT } AS IDENTITY` (preceded by `DROP DEFAULT` and `SET NOT NULL` when the current column is not yet identity-compatible).
  - `identity → none`: `DROP IDENTITY IF EXISTS` (followed by `DROP NOT NULL` when desired is nullable).
  - `always ↔ by default`: `SET GENERATED { ALWAYS | BY DEFAULT }`.
- The NOT NULL diff guard now skips only when the desired side is identity (which implies NOT NULL), so an `identity → nullable` transition now correctly emits `DROP NOT NULL` after `DROP IDENTITY`.

## Test plan

- [x] New plan fixtures cover all four transitions plus the two preconditions (`DROP DEFAULT` before ADD, `DROP NOT NULL` after DROP).
- [x] New apply fixtures with `verify_no_drift: true` confirm the plan applies cleanly and produces no follow-up diff.
- [x] `make vet`, `make lint`, `make test` pass.


winebarrel:
Addressed both review comments:

1. **`serial` → identity** — switched to unconditional `DROP DEFAULT` (idempotent in Postgres) so the hidden `nextval()` default that catalog reports as `nil` for `serial`/`bigserial`/`smallserial` is also dropped before `ADD GENERATED ... AS IDENTITY`. New fixture pair `add_identity_from_serial.yml` (plan + apply with `verify_no_drift`) covers the path end-to-end.
2. **Apply ordering fixtures** — added `add_identity_drops_default`, `add_identity_to_nullable_column`, `drop_identity_column_to_nullable` apply fixtures with `verify_no_drift: true` so each precondition (DROP DEFAULT before ADD, SET NOT NULL before ADD, DROP NOT NULL after DROP) is exercised against a real Postgres instance.

For the codecov failure: CI runs coverage per-package without `-coverpkg`, so the YAML plan/apply fixtures in the root test package didn't bump `diff/tables.go`'s patch coverage. Added direct unit tests in `diff/tables_test.go` covering all transitions; `alterColumnSQL` and `identityKind` are now at 100% in the diff package's own profile.

winebarrel:
Reverted the unconditional `DROP DEFAULT` and made the emit conditional on `current.Default != nil` OR `isSerialType(current.TypeName)`. `catalog/columns.go` renders `serial`/`bigserial`/`smallserial` explicitly as `TypeName`, so the serial case is detected without an extra catalog flag, and plain integer columns no longer pay the cost of a redundant `ALTER TABLE` lock under `Apply` without `--with-tx`. Updated fixtures and unit tests; `alterColumnSQL` / `identityKind` / `isSerialType` are at 100% in the diff package profile.

winebarrel:
Reused `serialBaseTypes` in `isSerialType` so the serial pseudo-type set lives in one place.

PR #142, opened 2026-05-04
Drop raw-form fallback in schema-replacer pair list

winebarrel:
## Summary

\`buildDefReplacer\` added two pairs to the replacer whenever \`model.Ident(from) != from\`: the canonical quoted form (e.g. \`\"a.b\".\`) and the raw concatenated form (e.g. \`a.b.\`).

All inputs to this replacer come from canonical SQL — catalog \`pg_get_*def\` output and parser pg_query deparse — and never contain the raw form for identifiers that require quoting, so the raw-form pair could not legitimately match anything. It could, however, collide with a three-part column reference: with \`SchemaMap{\"a.b\": \"x\"}\`, the raw pair \`a.b.\` rewrote \`a.b.col\` (a three-part name meaning schema \`a\`, table \`b\`, column \`col\`) to \`x.col\`, silently corrupting an unrelated reference.

Drop the raw-form fallback. Only the \`model.Ident\` form is added, so the replacer matches exactly the canonical quoted reference for schemas whose name needs quoting and never overlaps with three-part references.

## Note on the original ‘determinism’ framing

This PR previously framed the bug as a non-determinism issue arising from \`a.\` and \`a.b.\` overlapping in the replacer trie. Copilot reviewers correctly pointed out that valid SQL for a schema named \`a.b\` uses the quoted form \`\"a.b\".foo\`, never the raw \`a.b.foo\`, and that the only way \`a.b.\` ends up in the replacer at all is via the unsafe raw fallback — which is itself the actual bug. With the fallback removed, only quoted-form pairs remain, and \`.\` always sits at the trie boundary, so two distinct schemas can never produce trie-prefix-overlapping pairs and ordering is irrelevant.

## Test plan

- [x] New \`TestBuildDefReplacer_PreservesThreePartReference\` exercises \`SchemaMap{\"a.b\": \"x\", \"a\": \"y\"}\` against both the unsafe form (\`a.b.col\` → must rewrite as schema \`a\` only, yielding \`y.b.col\`) and the canonical form (\`\"a.b\".col\` → must rewrite the dotted-name schema, yielding \`x.col\`). Verified the test fails on the previous raw-fallback code and passes after the fix.
- [x] New \`TestBuildDefReplacer_QuotedSchema\` covers ordinary quoted-name rewrites.
- [x] Existing \`TestDump_WithSchemaMap_QuotedSchema\` (schema \`\"My Schema\"\`) continues to pass — catalog output uses the quoted form, which the remaining pair handles.
- [x] \`make vet\`, \`make lint\`, \`make test\` pass.

### DIFF
--- a/export_test.go
+++ b/export_test.go
@@ -11,6 +11,7 @@ var (
 	ExtractObjectName         = extractObjectName
 	OrderStatements           = orderStatements
 	CompareTaggedPos          = compareTaggedPos
+	BuildDefReplacer          = buildDefReplacer
 )
 
 func DumpResultTables(r *DumpResult) *orderedmap.Map[string, *model.Table] {
--- a/remap.go
+++ b/remap.go
@@ -9,18 +9,18 @@ import (
 
 // buildDefReplacer builds a strings.Replacer that replaces schema-qualified
 // prefixes in raw SQL definitions (e.g. "staging." → "public.").
-// It handles both unquoted and quoted schema identifiers.
+//
+// All inputs to this replacer come from canonical SQL — pg_get_*def output
+// from the catalog or pg_query deparse output from the parser — so any
+// identifier that requires quoting is already wrapped in double quotes.
+// Only the model.Ident form is added to the pair list. Adding an unquoted
+// fallback (e.g. raw `a.b.` for a schema literally named `a.b`) would
+// dangerously collide with three-part references like `a.b.col` (schema
+// `a`, table `b`, column `col`).
 func buildDefReplacer(schemaMap map[string]string) *strings.Replacer {
 	var pairs []string
 	for from, to := range schemaMap {
-		fromIdent := model.Ident(from)
-		toIdent := model.Ident(to)
-		// Always add the Ident form (handles quoting)
-		pairs = append(pairs, fromIdent+".", toIdent+".")
-		// If Ident differs from raw name, also add the raw form
-		if fromIdent != from {
-			pairs = append(pairs, from+".", to+".")
-		}
+		pairs = append(pairs, model.Ident(from)+".", model.Ident(to)+".")
 	}
 	return strings.NewReplacer(pairs...)
 }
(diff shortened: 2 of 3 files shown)


winebarrel:
Addressed both review comments by constructing a fresh `schemaMap` inside each loop iteration. This makes the runtime use a new per-map hash seed each time, so iteration order genuinely varies across iterations rather than relying on a single map's incidental behaviour.

Verified by temporarily reverting the longest-first sort in `buildDefReplacer`: the determinism test now fails (`-count=10` reliably hits a mismatch within ~200 iterations), confirming the test exercises the bug it's meant to guard against.

winebarrel:
Reframed and rebuilt. The reviews are right: `a.b.foo` isn't a valid reference for a schema named `a.b` (canonical form is `"a.b".foo`), and the only reason `a.b.` ever ended up in the replacer was the unsafe raw-form fallback — which is itself the actual bug because it collides with three-part column references like `a.b.col`.

Replaced both prior commits with a single commit that drops the raw-form fallback. With only the canonical `model.Ident` form in the pair list, two distinct schemas can never produce trie-prefix-overlapping pairs (`.` always sits at the trie boundary), so the longest-first sort and the dotted-schema regression test are no longer needed and have been removed.

New `TestBuildDefReplacer_PreservesThreePartReference` directly checks the actual hazard: `SchemaMap{"a.b": "x", "a": "y"}` with input `a.b.col` must rewrite the schema `a` only (→ `y.b.col`), while the canonical `"a.b".col` rewrites the dotted-name schema (→ `x.col`). Verified the test fails on the prior raw-fallback code and passes after the fix.

PR title and description updated to match the new scope.

winebarrel:
Reworded the doc comment so it matches the actual assertion: the schema portion is intentionally rewritten by SchemaMap (`a.` → `y.`); only the table/column segments (`b.col`) must remain intact.

PR #144, opened 2026-05-04
Validate non-empty Schemas at boundary in Plan/Apply/Dump

winebarrel:
## Summary

- The CLI defaults `Options.Schemas` to `["public"]` via kong, but library callers who construct `Options` directly can leave it empty. Today `catalog.NewCatalog` already errors on empty `Schemas`, but `Schemas[0]` is also indexed in `diff_all.go` before reaching that catalog call would short-circuit under any future refactor that re-orders or skips the catalog construction.
- Add an explicit boundary check at the top of `Plan`, `Apply`, and `Dump` that returns a clear pistachio-prefixed error message instead of relying on a downstream component to surface the misuse.

## Test plan

- [x] Existing `TestPlan_EmptySchemas` / `TestApply_EmptySchemas` / `TestDump_EmptySchemas` are tightened to assert the error message contains `"at least one schema must be specified"`.
- [x] `make vet`, `make lint`, `make test` pass.

### DIFF
--- a/apply.go
+++ b/apply.go
@@ -29,6 +29,9 @@ type ApplyResult struct {
 }
 
 func (client *Client) Apply(ctx context.Context, options *ApplyOptions, w io.Writer) (*ApplyResult, error) {
+	if err := client.validateSchemas(); err != nil {
+		return nil, err
+	}
 	conn, err := client.connect(ctx)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


winebarrel:
Extended `validateSchemas` to also reject empty / whitespace-only entries. `model.Ident` would otherwise drop the component silently and either emit malformed DDL or route changes through `search_path` to an unintended schema — both silent failure modes the bare `len() == 0` check missed.

New `TestPlan_BlankSchemaName` covers `""` and `"   "` cases.

PR #211, opened 2026-05-15
Fix view drift on IN clauses and current-side typed-literal casts

winebarrel:
## Summary

Follow-up to the #201/#203/#204/#205/#206/#207 cast-asymmetric series, this time for view body comparisons. `equalViewDef` (`diff/views.go`) previously only normalised whitespace and stripped schema / column qualifications, so two `pg_get_viewdef` canonicalisations showed false drift on every `pista apply`:

1. `WHERE x IN ('a','b')` is stored as `WHERE x = ANY (ARRAY['a','b'])`.
2. `WHERE status = 'published'` against an enum-typed column is stored as `WHERE status = 'published'::post_status` (and analogous text-cast / typed-literal cases).

Every view containing either shape emitted a redundant `CREATE OR REPLACE VIEW` on every plan/apply cycle.

## Approach

`equalViewDef` now feeds both sides into two new walkers that descend into every expression-bearing position of a view's `SELECT` body — target list, `WHERE`, `HAVING`, `GROUP BY`, `ORDER BY`, `LIMIT` / `OFFSET`, `JOIN ... ON`, CTE bodies, `UNION` arms, and sub-SELECTs in `FROM`:

- **`normalizeViewExprs`** (symmetric, both sides): calls the existing `normalizeCheckExpr` at each expression position, converting `= ANY(ARRAY[...])` back to `IN (...)` and stripping text-like casts.
- **`alignViewCasts`** (asymmetric, current side only): calls the existing `alignCurrentCasts` at the same positions, stripping `TypeCast` nodes that appear on the catalog form but not in the desired SQL. A cast the user explicitly wrote in desired but missing from current still surfaces as a real change (regression-tested).

`normalizeCheckExpr` and `alignCurrentCasts` themselves gained a `SubLink` case that recurses into the contained `SELECT` via the new walkers. CHECK constraints, defaults, generated columns, and index predicates cannot contain sub-queries in PostgreSQL, so this is a no-op for them. RLS policy `USING` / `WITH CHECK` *can* contain sub-queries, so they benefit from the same improvement — semantically-equivalent `USING (id IN (SELECT ...))` forms no longer surface as drift.

## Test plan

- [x] 30+ new unit tests in `diff/views_test.go` cover each walker position with both the IN↔ANY and current-only TypeCast normalisations (target list, WHERE, HAVING, GROUP BY, ORDER BY, LIMIT, OFFSET, CTE, UNION, JOIN ON, RangeSubselect, SubLink) — plus regression guards (`notStrippedFromDesired`, `realChangeStillDetected`).
- [x] Three end-to-end YAML plan fixtures against a real PostgreSQL:
  - `no_diff_view_in_clause.yml` — view with `IN ('published', 'pinned')` → no drift.
  - `no_diff_view_enum_filter.yml` — view filtering on an enum column → no drift after `'x'::post_status` cast is added by catalog.
  - `no_diff_view_in_clause_sublink.yml` — `EXISTS (SELECT ... WHERE status IN (...))` → no drift (guards the `SubLink` recursion in `normalizeCheckExpr` / `alignCurrentCasts`).
- [x] `go test -p 1 ./...` — all packages pass.
- [x] `make lint` — 0 issues.
- [x] Coverage: `diff/` is 96.5 % (+0.2 pt over main baseline 96.3 %). `normalizeViewExprs` 97.6 %, `alignViewCasts` 96.6 %. Remaining uncovered lines are defensive nil / parse-error guards.

## Notes

- Dead SubLink branches at the top of `normalizeViewExprs` and `alignViewCasts` were removed once tests confirmed every real SubLink path goes through `normalizeCheckExpr` / `alignCurrentCasts` (which pass the SubLink's `Subselect` — a `SelectStmt` — never the SubLink itself).


winebarrel:
Note on the Codecov "5 missing / 5 partials" in `diff/views.go`: these are the same 5 defensive sites counted twice (the `if` condition as a "partial" and its `return` body as a "miss").

| `if` (partial) | `return` body (miss) | Guard |
|---|---|---|
| L62 `if errCur \|\| errDes != nil` | L63 | `pg_query.Parse` fallback in `equalNormalisedViewSelect` — input was already Parse+Deparse'd successfully by `normalizeViewDef`, so a 2nd Parse failure is essentially impossible |
| L78 same (post-Deparse) | L79 | Deparse fallback after our in-place AST walk |
| L89 `if node == nil` | L90 | nil node guard in `normalizeViewExprs`, matches the pre-existing style of `stripQualifications` |
| L163 `if desired == nil \|\| current == nil` | L164 | nil guard in `alignViewCasts` |
| L176 `if cs == nil` | L177 | structural-mismatch guard (desired is a `SelectStmt` but current isn't — both come from the same source so this is defensive) |

These are unreachable through natural inputs (every caller's contract rules them out). Per CLAUDE.md's "don't write unnatural tests for unreachable defensive code", I left them in rather than removing the guards or chasing the number with synthetic tests. Patch coverage (92.95 %) is well above the project's `codecov.yml` patch target of 75 %.

winebarrel:
Addressed both Copilot review comments in dbd81f9:

1. **`stripQualifications` recursing only into `SubLink.Subselect`** — confirmed the gap with a failing test (`TestEqualViewDef_inSubquery_testexprQualified`): `users.id IN (SELECT user_id FROM orders)` vs `id IN (SELECT user_id FROM public.orders)` reported drift because the LHS lives in `Testexpr`. Fixed by also recursing into `sl.Testexpr`, matching the shape of the `SubLink` handling already in `normalizeCheckExpr` / `alignCurrentCasts`. Test now passes.
2. **Naming consistency** — renamed `equalNormalisedViewSelect` → `equalNormalizedViewSelect` to match the rest of the package (`normalizeViewDef`, `normalizeCheckExpr`, `normalizeViewExprs`, `normalizeIndexStmt`, etc.). Updated the related comments in `views.go` and one comment in `views_test.go` to use American spelling. Left the pre-existing British spellings in `tables.go` comments (lines 1135, 1204, 1322) untouched — those predate this PR.

`go test -p 1 ./...` and `make lint` both clean.

winebarrel:
Addressed both Copilot review comments in 5ee7b7d:

Renamed `normalizeViewExprs` → `normalizeSelectExprs` and `alignViewCasts` → `alignSelectCasts`. Both walkers are now reachable from outside view comparison (`normalizeCheckExpr` / `alignCurrentCasts` SubLink cases route sub-queries inside CHECK / RLS expressions through them), so the `View` prefix was misleading. `Select` reflects that they walk an arbitrary `SelectStmt` tree. Also updated the docstring on `normalizeSelectExprs` to mention both call sites.

`go test -p 1 ./...` and `make lint` both clean.

winebarrel:
Addressed both Copilot review comments.

**Re: line 113 (symmetric text-cast strip in target list) — fixed in f6b7b46.** Confirmed the concern with two probes:
- `equalViewDef("SELECT id FROM t", "SELECT id::text FROM t")` returned `true` (user-added cast hidden by `normalizeCheckExpr`'s symmetric text-cast strip).
- `equalViewDef("SELECT id::text FROM t", "SELECT id FROM t")` also returned `true` (user-removed cast hidden by `alignCurrentCasts`' current-only strip).

Both shapes change the view's output column type and must surface as a diff. Fix routes target-list expressions through two new helpers (`normalizeTargetExpr` / `alignTargetCasts`) that preserve a top-level `TypeCast` and only recurse into `tc.Arg` below it. Paren stripping and `IN`↔`ANY` normalization still apply below the cast, so the existing `_targetList` tests for those still pass. Three new tests pin the three top-level-cast shapes (`_added` / `_removed` / `_bothPresent`).

Other walker positions (WHERE / HAVING / GROUP BY / ORDER BY / LIMIT / OFFSET / JOIN-ON / CTE / UNION / RangeSubselect / SubLink) are unaffected: a top-level cast in a predicate doesn't change the view's output column shape, so the existing pipeline is still correct there.

**Re: line 164 (`normalizeSelectExprs` / `alignSelectCasts` duplicate traversal) — keeping as-is.** The duplication is intentional structural parallelism: one walker mutates a single tree, the other walks two trees in parallel and conditionally aligns. A shared traversal helper would need different callback shapes for the two modes (single-node mutation vs paired-node alignment) and would obscure the cases where alignment requires a length check or per-position branch (`if dtc != nil && ctc != nil` etc.). Per CLAUDE.md's "don't add abstractions beyond what the task requires," I'd rather rely on test coverage to detect drift: each walker position has matching `inVsAnyArray_<pos>` (exercises `normalizeSelectExprs`) and `currentOnlyTypeCast_<pos>` (exercises `alignSelectCasts`) tests, so changing one walker without the other would fail a paired test.

`go test -p 1 ./...` and `make lint` both clean.

winebarrel:
Addressed in 7480ba1. Confirmed the double pg_query work: `normalizeViewDef` ran a parse + deparse on each side, then `equalNormalizedViewSelect` re-parsed both and deparsed both again — drift-prone views (the exact case this branch fixes) paid 2× the pg_query cost.

Collapsed the whole flow into a single roundtrip per side:

```go
func equalViewDef(current, desired string) bool {
    if current == desired { return true }
    curResult, errCur := pg_query.Parse("CREATE VIEW _v AS " + current)
    desResult, errDes := pg_query.Parse("CREATE VIEW _v AS " + desired)
    if errCur != nil || errDes != nil { return current == desired }
    for _, stmt := range curResult.Stmts { stripQualifications(stmt.Stmt); normalizeSelectExprs(stmt.Stmt) }
    for _, stmt := range desResult.Stmts { stripQualifications(stmt.Stmt); normalizeSelectExprs(stmt.Stmt) }
    if len(curResult.Stmts) == 1 && len(desResult.Stmts) == 1 {
        alignSelectCasts(desResult.Stmts[0].Stmt, curResult.Stmts[0].Stmt)
    }
    curStr, errCur := pg_query.Deparse(curResult)
    desStr, errDes := pg_query.Deparse(desResult)
    if errCur != nil || errDes != nil { return current == desired }
    return curStr == desStr
}
```

The intermediate `normCur == normDes` short-circuit goes away — the merged path produces the same final string for whitespace-only differences too, so behaviour is preserved. The `normalizeViewDef` helper and its unit test are dropped since the merged `equalViewDef` is the only caller of that pipeline.

`go test -p 1 ./...` and `make lint` clean. Sanity-checked end-to-end against pagila: round-trip on the dump plus `CREATE VIEW v AS SELECT … FROM film WHERE rating IN ('G','PG')` against the catalog-stored `= ANY (ARRAY['G'::mpaa_rating, 'PG'::mpaa_rating])` form still produces `No changes`.

winebarrel:
Addressed both Copilot review comments in c94b585.

**Re: line 87 / line 199 (ResTarget.Val nil guards).** Confirmed the project convention — `toposort/schema.go:228` already does `if rt := target.GetResTarget(); rt != nil && rt.Val != nil` before walking. The protobuf getter on a nil receiver happens to return nil today (so `normalizeTargetExpr(nil).GetTypeCast()` is currently safe and `normalizeCheckExpr` has its own `node == nil` guard at entry), but matching the explicit `rt.Val != nil` / `dt.Val != nil && ct.Val != nil` form keeps this loop in step with the rest of the codebase and documents the invariant for future readers.

**Re: line 167 (stale doc).** Good catch — the doc still pointed at the removed `equalNormalizedViewSelect`. Updated to name the actual call sites: `equalViewDef` (the merged pipeline from 7480ba1) and `alignCurrentCasts`' SubLink case in `tables.go` (where sub-queries inside CHECK / RLS expressions feed back through the same walker).

`go test -p 1 ./...` and `make lint` clean.

winebarrel:
Addressed in 2259a02. Confirmed the bug with a probe:

```go
equalViewDef(
    "SELECT lower(users.name::text) FROM users",
    "SELECT lower(name) FROM users",
)
```

returned `false` before the fix. `stripQualifications` had no `TypeCast` case, so when iterating the FuncCall args it hit `users.name::text`, didn't recurse into `tc.Arg`, and left the inner `users.name` qualified. After `normalizeSelectExprs` ran and collapsed the text-like cast, the now-bare `users.name` deparsed asymmetrically against the desired side's unqualified `name`.

Fix adds a `TypeCast` case at the end of the `stripQualifications` switch that recurses into `tc.Arg` (mirroring the other recursive cases). Tests:

- `TestEqualViewDef_qualifiedColumnInsideTextCast` — pins the exact shape above.
- All existing tests continue to pass; the new case only adds recursion at TypeCast nodes that previously short-circuited.

`go test -p 1 ./...` and `make lint` clean.

PR #291, opened 2026-07-18
Fix deparse dropping key column named "value" in UNIQUE/PK constraints

winebarrel:
## Problem

libpg_query drops a single key column named `value` from the deparsed
column list of a UNIQUE/PRIMARY KEY constraint, even though the parse
tree keeps it (upstream [pganalyze/pg_query_go#148](https://github.com/pganalyze/pg_query_go/issues/148)).

This produced a broken constraint `Definition` (`UNIQUE` with no columns), causing:

- invalid `CREATE TABLE` / `ALTER TABLE ... ADD CONSTRAINT` DDL that fails on apply with a syntax error, and
- false drift: a table already carrying `UNIQUE (value)` was reported as changed on every `plan`.

```
-- desired schema
CREATE TABLE api_keys (
    id bigint PRIMARY KEY,
    value text NOT NULL,
    CONSTRAINT api_keys_value_unique UNIQUE (value)
);

-- before: pista plan (broken)
ALTER TABLE public.api_keys ADD CONSTRAINT api_keys_value_unique UNIQUE;
--> ERROR: syntax error at or near ";"
```

It affects single-column `UNIQUE`/`PRIMARY KEY` on a column named `value` (table- and column-level). Multi-column keys, other column names, and `pista dump` (which reads `pg_get_constraintdef`) are unaffected.

## Fix

`deparseConstraintDef` now swaps each key column name for a collision-free
placeholder that deparses reliably, then substitutes the real identifiers
back into the output. The parse tree keeps the correct keys, so this
restores the column list without depending on the upstream fix. Only
UNIQUE/PRIMARY KEY constraints populate `Keys`, so other constraint types
are unaffected.

## Tests

- Parser unit tests: `Definition` for `UNIQUE (value)` and `PRIMARY KEY (value)`.
- Regression test for an 11-column composite key including `value`, guarding against placeholder substring collisions (`_1` vs `_10`).
- Plan fixtures: ALTER ADD, UNIQUE no-drift, PK no-drift.

Verified with `make lint` (0 issues), `make vet`, the full test suite, and end-to-end `pista plan`/`apply` across the create, drift, and alter paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.16.1] - 2026-07-18
+
+* Fix invalid DDL for a single-column `UNIQUE` / `PRIMARY KEY` constraint on a column named `value`. libpg_query dropped the column from the deparsed definition, so `plan` / `apply` produced a broken `ADD CONSTRAINT ... UNIQUE` and reported false drift. The key columns are now restored from the parse tree. ([#291](https://github.com/winebarrel/pistachio/pull/291))
+
 ## [1.16.0] - 2026-07-13
 
 * Open the `plan` and `dump` connections in read-only mode, so those commands cannot write to the database even by accident. `apply` still applies DDL. Pass `--no-read-only` (env `$PISTA_NO_READ_ONLY`) for a read-write connection.
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -1329,6 +1329,35 @@ func deparseConstraintDef(con *pg_query.Constraint) (string, error) {
 	con.SkipValidation = false
 	defer func() { con.SkipValidation = origSkipValidation }()
 
+	// Work around a libpg_query deparse bug: a single key column named "value"
+	// is dropped from the deparsed column list, even though the parse tree
+	// keeps it. Swap each key column name for a collision-free placeholder that
+	// deparses reliably, then substitute the real identifiers back into the
+	// output. Only UNIQUE/PRIMARY KEY constraints populate Keys, so other
+	// constraint types are unaffected.
+	swapped := make([]*pg_query.String, 0, len(con.Keys))
+	origKeys := make([]string, 0, len(con.Keys))
+	repl := make(map[string]string, len(con.Keys))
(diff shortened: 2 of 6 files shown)


winebarrel:
Reviewed. The fix is well-scoped and the workaround is documented clearly (upstream link + revert note + defer-based restore). A few notes, none blocking.

**Verified independently**

- Restricting the swap to `con.Keys` (leaving `con.Including` untouched) is correct: `UNIQUE (email) INCLUDE (value)` deparses to `UNIQUE (email) INCLUDE (value)` with no drop, so the bug is specific to the key-column list. No gap on the INCLUDE side.
- The `_e` delimiter reasoning holds: `..._1_e` is never a substring of `..._10_e`, and the 11-column regression test exercises the `_10` boundary. Good catch to guard that.
- Drift comparison stays safe because `equalConstraintDef` re-parses and re-deparses both sides, so routing key idents through `model.Ident` instead of libpg_query's own quoting cannot introduce false drift.

**Suggestion (minor)**

- The description says the bug affects column-level constraints too ("table- and column-level"), and it does work: `value text UNIQUE` -> `UNIQUE (value)` and `value text PRIMARY KEY` -> `PRIMARY KEY (value)`, since column-level constraints fill `con.Keys` at parser.go:539/548 and then hit the same swap. But the tests only cover the table-level `CONSTRAINT ... (value)` form. Since the description explicitly claims the column-level case, adding a one-line assertion (or a plan fixture) for `value text UNIQUE` would lock that path in against regressions.

Nothing else. `con.Keys`-only scoping, placeholder collision handling, and restore-on-return all look right.

winebarrel:
Thanks for the review. Addressed the suggestion: added `TestParseSQL_ColumnLevelUniqueOnValueColumn` and `TestParseSQL_ColumnLevelPrimaryKeyOnValueColumn` to lock the column-level `value text UNIQUE` / `value text PRIMARY KEY` paths (they fill `con.Keys` from the column name and hit the same workaround). Pushed in 336a166. lint is clean and the parser suite passes.

PR #296, opened 2026-07-21
Add declarative management of standalone sequences

winebarrel:
## Summary

Adds declarative management of standalone sequences (`CREATE SEQUENCE`) through the plan/apply/dump pipeline: create, alter, drop, comment, and rename.

**Core rule:** only standalone sequences are managed. A sequence owned by a `serial` or identity column is handled as part of that column and excluded from the sequence diff, so it is never double-managed. The distinction comes from `pg_depend.deptype` — `'a'` for serial, `'i'` for identity; a sequence with neither is standalone.

## What it does

- **CREATE** — emits `CREATE SEQUENCE` filling in PostgreSQL's implicit defaults (type/increment sign decide min/max/start), so a bare `CREATE SEQUENCE s;` round-trips with no drift.
- **ALTER** — a single `ALTER SEQUENCE` combining only the changed options (`AS`, `INCREMENT BY`, `MINVALUE`, `MAXVALUE`, `START WITH`, `CACHE`, `CYCLE` / `NO CYCLE`).
- **DROP** — gated by `--allow-drop sequence`; skipped drops are surfaced as `-- skipped:`.
- **COMMENT** — `COMMENT ON SEQUENCE` add/remove, round-tripped by dump.
- **RENAME** — via the `-- pista:renamed-from` directive.
- **dump** — standalone sequences are emitted (respects `--omit-schema` / `--split`).

## Design notes

- **catalog** — `ListSequences` (all, incl. owned) is split from `Sequences` (standalone map). The owner-detection join now covers `deptype IN ('a','i')`, so identity-owned sequences are correctly excluded (previously only serial was). The sequence comment is now read too.
- **ordering** — the toposort learns a table→sequence dependency when a column default references the sequence via `nextval`, so the sequence is created before the table (and dropped after) regardless of name order.
- **defaults** — the parser reproduces PostgreSQL's defaults exactly; bigint bounds arrive from pg_query as `Float` nodes and are parsed to `int64`.

## Behavioral notes

- Changing `START WITH` does not reset the sequence's current value (`RESTART` is not part of declarative diffing).
- Shrinking `MAXVALUE` below the current value can make `ALTER SEQUENCE` error at apply time; the error propagates as-is.

## Tests

- 25 YAML fixtures across plan/apply/dump (create/alter/drop/rename/comment/ordering, serial+identity exclusion, `OWNED BY` exclusion, descending/smallint/`NO MINVALUE` defaults, disallowed drop, rename errors).
- Unit tests for `model`, `parser` (default table), `diff` (incl. cross-schema rename error), `toposort` (nextval edge), and `catalog` (owner + comment).
- `go vet`, `golangci-lint` (0 issues), full `make test`, and `make test-scenario` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.17.0] - 2026-07-21
+
+* Manage standalone sequences. `plan`, `apply`, and `dump` handle `CREATE SEQUENCE`, `ALTER SEQUENCE`, `DROP SEQUENCE`, `COMMENT ON SEQUENCE`, and rename via `-- pista:renamed-from`. Only standalone sequences are managed. A sequence owned by a `serial` or identity column stays with that column and is excluded from the diff, classified by `pg_depend.deptype`. A sequence referenced by a column default via `nextval` is created before that table and dropped after. `--enable`, `--disable`, `--include`, `--exclude`, and `--allow-drop` accept `sequence`. The `UNLOGGED` attribute is not tracked. ([#296](https://github.com/winebarrel/pistachio/pull/296))
+
 ## [1.16.1] - 2026-07-18
 
 * Fix invalid DDL for a single-column `UNIQUE` / `PRIMARY KEY` constraint on a column named `value`. libpg_query dropped the column from the deparsed definition, so `plan` / `apply` produced a broken `ADD CONSTRAINT ... UNIQUE` and reported false drift. The key columns are now restored from the parse tree. ([#291](https://github.com/winebarrel/pistachio/pull/291))
--- a/README.md
+++ b/README.md
@@ -502,15 +502,16 @@ pista dump --split ./schema/
 
 - Domain types (`CREATE DOMAIN`, `ALTER DOMAIN SET/DROP DEFAULT`, `SET/DROP NOT NULL`, `ADD/DROP CONSTRAINT`)
 - Enum types (`CREATE TYPE ... AS ENUM`, `ALTER TYPE ... ADD VALUE`)
+- Sequences (`CREATE SEQUENCE`, `ALTER SEQUENCE`, `DROP SEQUENCE`). Only standalone sequences are managed; sequences owned by a serial or identity column are handled as part of that column, not as separate objects.
 - Tables (including unlogged and partitioned tables)
 - Views
 - Materialized views
 - Columns (serial/bigserial/smallserial, identity, generated)
 - Constraints (primary key, unique, check, exclusion, foreign key)
 - Indexes (unique, partial, expression, hash, multi-column)
-- Comments (on tables, columns, views, types, domains)
+- Comments (on tables, columns, views, types, domains, sequences)
 - Row-level security (`ALTER TABLE ... ENABLE/DISABLE/FORCE/NO FORCE ROW LEVEL SECURITY`, policies via `CREATE POLICY` / `ALTER POLICY` / `DROP POLICY`)
-- Renaming (tables, views, enums, enum values, domains, columns, constraints, foreign keys, indexes, policies via `-- pista:renamed-from` directive)
+- Renaming (tables, views, enums, enum values, domains, sequences, columns, constraints, foreign keys, indexes, policies via `-- pista:renamed-from` directive)
 - Array, JSON, UUID, and other built-in types
 - Quoted identifiers
 
--- a/TODO.md
+++ b/TODO.md
@@ -86,17 +86,6 @@ for in-place changes; using it would avoid the round-trip.
 Origin: post-[#125](https://github.com/winebarrel/pistachio/pull/125) audit. Optimisation rather than a bug; current
 behaviour is correct, just heavier than necessary.
 
-## Standalone Sequences are not first-class
-
-`catalog/sequences.go` reads sequences (used for SERIAL/IDENTITY
-metadata), but parser and diff don't handle `CREATE SEQUENCE` /
-`ALTER SEQUENCE` / `DROP SEQUENCE` as schema operations. Standalone
-sequences in desired SQL therefore round-trip via execute directives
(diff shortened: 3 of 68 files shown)


winebarrel:
## Review

### Overview

Adds standalone sequence management (create/alter/drop/comment/rename) across catalog, parser, diff, toposort, plan/apply/dump, with the serial/identity ownership rule enforced symmetrically on both the catalog side (`pg_depend.deptype IN ('a','i')`) and the desired side (`standaloneSequences` filtering out `OWNED BY`). The implementation follows the existing per-object-type patterns closely (rename detection mirrors enums, remap mirrors domains, filter/ignore/drop-policy wiring matches the other types), so there is little novel surface to go wrong. Test coverage is thorough: 25 fixtures, unit tests for each layer, and a scenario test that keeps verifying the owned sequences never leak into the diff.

### Findings

**1. Quote-requiring sequence names miss the nextval dependency edge (minor, edge case)**

`extractSeqDeps` in `toposort/schema.go` extracts the literal from `nextval('...')` and, for the unqualified case, calls `resolveUnqualified(model.Ident(lit), ...)`. When the sequence name requires quoting (e.g. `CREATE SEQUENCE public."MySeq"`), the default expression is `nextval('"MySeq"'::regclass)`, so `lit` arrives already quoted as `"MySeq"`. `model.Ident` re-quotes it to `"""MySeq"""`, the lookup misses, and the edge is dropped. Ordering then falls back to name order, so a table whose default references a quote-requiring sequence name can be emitted before the `CREATE SEQUENCE` and fail at apply time.

The qualified case (`'"MySchema".seq'`) happens to work because `pg_get_expr` quotes identifiers the same way `Ident` does. Suggest treating an already-quoted literal as-is for the unqualified lookup (e.g. try `resolveUnqualified(lit, ...)` without re-quoting, or strip the surrounding quotes before `Ident`). A fixture like `create_sequence_before_table.yml` with a `"Zzz_seq"`-style name would lock it in.

**2. Redundant condition in `detectSequenceRenames` (nit)**

In `diff/rename.go`, the inner `if oldKey != newKey` guard is always true because `oldKey == newKey` already hit `continue` above. This faithfully mirrors `detectEnumRenames`/`detectTableRenames`, so consistency is a fair reason to keep it; if you ever clean it up, all three could go together.

**3. Rename combined with option changes has no fixture (test gap, nit)**

`detectSequenceRenames` feeds the adjusted map into the modified-sequences loop, so a rename plus an `INCREMENT BY`/`MAXVALUE` change in the same plan should emit both the `RENAME` and the `ALTER`. The scenario test renames with identical options, so this path is currently untested. One small plan fixture would cover it.

### Notes (no action needed)

- The `deptype IN ('a','i')` fix means identity-owned sequences are now correctly excluded where previously only serial was; the new `identity sequence has owner` catalog test pins this.
- `ALTER SEQUENCE ... OWNED BY` on an existing standalone sequence creates a deptype `'a'` dependency, so the catalog side classifies it as owned, matching the desired-side `OWNED BY` exclusion. The symmetry holds.
- The documented behavioral caveats (START WITH not resetting the current value, MAXVALUE shrink failing at apply) are reasonable for declarative diffing, and the UNLOGGED limitation is recorded in TODO.md.

winebarrel:
Thanks for the review. Addressed in 4f6b190:

**1. Quote-requiring sequence names miss the nextval edge** — Fixed. `extractSeqDeps` was calling `resolveUnqualified(model.Ident(lit), ...)`, which double-quoted an already-quoted literal like `"MySeq"`. Since `lit` comes from `pg_get_expr` / deparse it is already in canonical (quoted-when-needed) form, so it now passes `lit` through unchanged. Confirmed end-to-end: `CREATE SEQUENCE public."Zzz_seq"` referenced by `public."Aaa"` (which sorts first) now orders the sequence before the table and applies cleanly. Locked in with `testdata/plan/create_sequence_before_table_quoted_name.yml` (both names quoted so only the edge, not alphabetical order, can produce the correct result) and a `toposort` unit test.

**2. Redundant `if oldKey != newKey` in `detectSequenceRenames`** — Left as-is for consistency with `detectEnumRenames` / `detectTableRenames`, as you suggested. If it's cleaned up later, the three should go together.

**3. Rename + option change had no fixture** — Added `testdata/plan/rename_sequence_with_option_change.yml`, which renames and changes `INCREMENT BY` / `MAXVALUE` in the same plan and asserts both the `RENAME` and the follow-up `ALTER` (on the new name) are emitted.

winebarrel:
## Follow-up review (b1323fb, 4f6b190)

The two new commits address the previous findings. Verified each against the code and ran the relevant tests locally against PostgreSQL (`TestPlan/create_sequence_before_table_quoted_name`, `TestPlan/rename_sequence_with_option_change`, and the toposort sequence tests all pass).

**Finding 1 (quoted nextval edge) — fixed.** `extractSeqDeps` now passes the literal to `resolveUnqualified` as-is instead of re-quoting it with `model.Ident`. This is the right fix: `resolveUnqualified` builds `Ident(defaultSchema) + "." + name`, and since `pg_get_expr`/deparse quote identifiers the same way `Ident` does, the already-quoted literal matches the `defined` keys directly. The updated doc comment on `resolveUnqualified` spelling out the name-normalization contract is a nice touch, and the fix is pinned at both levels (toposort unit test and the plan fixture with a quoted table name that sorts before the sequence).

**Finding 3 (rename + option change) — covered.** `rename_sequence_with_option_change.yml` locks in the rename-then-ALTER behavior, confirming the adjusted map from `detectSequenceRenames` flows into the modified-sequences diff correctly.

Finding 2 (redundant `oldKey != newKey` guard) was left as-is for consistency with the enum/table variants, which is the call I suggested.

The additional coverage beyond the findings is also welcome: the rename-already-applied idempotency test, the `defElemInt64` Float branch with explicit bigint bounds, the schema-map remap tests for both directions, and the `--disable`/`--exclude` dump fixtures.

LGTM.

PR #331, opened 2026-07-30
Add composite type support

winebarrel:
## Summary

Adds declarative management of PostgreSQL **composite types** (`CREATE TYPE ... AS (...)`), alongside the existing enum and domain support. You define the desired composite types in your schema files and pistachio generates the DDL diff to converge the database.

## What's covered

- **Create / drop** composite types, gated by `--allow-drop=composite_type`
- **ALTER TYPE** `ADD ATTRIBUTE` / `DROP ATTRIBUTE` / `ALTER ATTRIBUTE ... TYPE`, matched by attribute name (order-independent, since PostgreSQL cannot reorder attributes)
- **Rename**: type (`ALTER TYPE ... RENAME TO`) and attribute (`ALTER TYPE ... RENAME ATTRIBUTE`) via `-- pista:renamed-from`
- **Comments** on the type and on each attribute (`COMMENT ON TYPE` / `COMMENT ON COLUMN`)
- **Collation** on attributes (`COLLATE`)
- **Dependency ordering**: a composite type is created after the enums / domains / composite types its attributes reference, and before tables that use it
- Wired through **dump** (including `--split` and `--omit-schema`), **plan**, **apply**, `--include` / `--exclude`, `--enable` / `--disable` (`composite_type`), **schema-map** remap, and the `-- pista:ignore` directive

`--allow-drop=composite_type` also gates `DROP ATTRIBUTE` (attribute removal loses data). Whole-object and attribute-level drops are suppressed by default and surfaced as `-- skipped:` comments, matching the existing drop-policy behavior.

## Scope notes

- Attribute reordering produces no diff (PostgreSQL has no reorder operation; `ADD ATTRIBUTE` always appends).
- `--omit-schema` remains dump-only (unchanged); plan/apply resolve unqualified type names to the default schema.
- Roles / grants / functions / triggers stay out of scope (use `-- pista:execute`); this PR only adds the composite type object.

## Tests

- Unit: `model` (9), `diff` (18, 100% coverage), `catalog` (4), schema-map remap (2)
- Fixtures: `plan` (15), `apply` (10), `dump` (3) covering create / alter / drop / rename / comment / collation / dependency ordering / ignore / enable / disable / include / omit-schema / disallowed-drop / unqualified names / idempotency
- New-code coverage: `model` and `diff` 100%; `catalog` and `parser` at parity with the existing enum/domain code (only DB-error and nil-guard branches remain)
- `make test` (`-p 1`) all pass, `make lint` 0 issues

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * Manage composite types. `plan`, `apply`, and `dump` handle `CREATE TYPE ... AS (...)`, `ALTER TYPE ... ADD/DROP/ALTER ATTRIBUTE`, `RENAME TO`, `RENAME ATTRIBUTE`, and comments on the type and its attributes. A composite type is created after the enums, domains, and composite types its attributes reference, and before tables that use it. `--enable`, `--disable`, `--include`, `--exclude`, and `--allow-drop` accept `composite_type`; `--allow-drop=composite_type` also gates `DROP ATTRIBUTE`. Attribute reordering is not diffed (PostgreSQL cannot reorder attributes), and `ALTER ATTRIBUTE ... TYPE` fails at apply while a table column uses the type. ([#331](https://github.com/winebarrel/pistachio/pull/331))
 
-* Fix a schema-qualified user-defined type (enum, domain, or composite) used as a column or attribute type producing a no-op `SET DATA TYPE` on every plan. The column type is now compared with the type's own schema stripped. Also set `search_path` to the target schemas before applying managed DDL, so an unqualified user-type reference resolves for a non-public target schema instead of failing at apply. ([#331](https://github.com/winebarrel/pistachio/pull/331))
+* Fix a schema-qualified user-defined type (enum, domain, or composite) used as a column or attribute type producing a no-op `SET DATA TYPE` on every plan. The column type is now compared with the type's own schema stripped. Also set `search_path` to the target schemas plus `public` before applying managed DDL, so an unqualified reference to a user type or to a `public` object (such as an extension type) resolves for a non-public target schema instead of failing at apply. ([#331](https://github.com/winebarrel/pistachio/pull/331))
 
 ## [1.19.0] - 2026-07-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.20.0] - 2026-07-30
+
+* Manage composite types. `plan`, `apply`, and `dump` handle `CREATE TYPE ... AS (...)`, `ALTER TYPE ... ADD/DROP/ALTER ATTRIBUTE`, `RENAME TO`, `RENAME ATTRIBUTE`, and comments on the type and its attributes. A composite type is created after the enums, domains, and composite types its attributes reference, and before tables that use it. `--enable`, `--disable`, `--include`, `--exclude`, and `--allow-drop` accept `composite_type`; `--allow-drop=composite_type` also gates `DROP ATTRIBUTE`. Attribute reordering is not diffed (PostgreSQL cannot reorder attributes), and `ALTER ATTRIBUTE ... TYPE` fails at apply while a table column uses the type. ([#331](https://github.com/winebarrel/pistachio/pull/331))
+
 ## [1.19.0] - 2026-07-29
 
 * Add `--assume-validated` (env `$PISTA_ASSUME_VALIDATED`) to treat every constraint as validated. Pistachio ignores `NOT VALID` in the desired schema and never emits `NOT VALID` or `VALIDATE CONSTRAINT`. Constraint additions and definition changes still apply, without `NOT VALID`. Use it when `NOT VALID` is a one-off migration step you do not want in the desired state.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Manage composite types. `plan`, `apply`, and `dump` handle `CREATE TYPE ... AS (...)`, `ALTER TYPE ... ADD/DROP/ALTER ATTRIBUTE`, `RENAME TO`, `RENAME ATTRIBUTE`, and comments on the type and its attributes. A composite type is created after the enums, domains, and composite types its attributes reference, and before tables that use it. `--enable`, `--disable`, `--include`, `--exclude`, and `--allow-drop` accept `composite_type`; `--allow-drop=composite_type` also gates `DROP ATTRIBUTE`. Attribute reordering is not diffed (PostgreSQL cannot reorder attributes), and `ALTER ATTRIBUTE ... TYPE` fails at apply while a table column uses the type. ([#331](https://github.com/winebarrel/pistachio/pull/331))
 
+* Fix a schema-qualified user-defined type (enum, domain, or composite) used as a column or attribute type producing a no-op `SET DATA TYPE` on every plan. The column type is now compared with the type's own schema stripped. Also set `search_path` to the target schemas before applying managed DDL, so an unqualified user-type reference resolves for a non-public target schema instead of failing at apply. ([#331](https://github.com/winebarrel/pistachio/pull/331))
+
 ## [1.19.0] - 2026-07-29
 
 * Add `--assume-validated` (env `$PISTA_ASSUME_VALIDATED`) to treat every constraint as validated. Pistachio ignores `NOT VALID` in the desired schema and never emits `NOT VALID` or `VALIDATE CONSTRAINT`. Constraint additions and definition changes still apply, without `NOT VALID`. Use it when `NOT VALID` is a one-off migration step you do not want in the desired state.
--- a/README.md
+++ b/README.md
@@ -462,7 +462,7 @@ CREATE TYPE public.status AS ENUM (
 );
 ```
 
-**Composite types** (statement level) **and their attributes** (inside `CREATE TYPE ... AS (...)`, before the attribute). A type rename emits `ALTER TYPE ... RENAME TO`; an attribute rename emits `ALTER TYPE ... RENAME ATTRIBUTE`, which keeps stored data:
+**Composite types and their attributes**. The statement-level directive renames the type (`ALTER TYPE ... RENAME TO`). A directive inside `CREATE TYPE ... AS (...)`, before an attribute, renames that attribute (`ALTER TYPE ... RENAME ATTRIBUTE`) and keeps stored data:
 
 ```sql
 -- pista:renamed-from public.address
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ pista apply schema.sql                 # apply it
 Or split the schema across multiple files:
 
 ```bash
-pista dump --split ./schema/       # dump per table/view/enum/domain
+pista dump --split ./schema/       # dump per table/view/enum/domain/composite type
 pista plan ./schema/*.sql          # review the diff
 pista apply ./schema/*.sql         # apply it
 ```
@@ -245,7 +245,7 @@ pista plan --assume-validated schema.sql
 pista apply --assume-validated schema.sql
 ```
 
-By default, `plan` and `apply` do not drop tables, views, enums, domains, columns, constraints, foreign keys, or indexes. Use `--allow-drop` to enable dropping specific object types (`all`, `table`, `view`, `enum`, `domain`, `column`, `constraint`, `foreign_key`, `index`). Also available as `$PISTA_ALLOW_DROP`. `constraint` covers CHECK / UNIQUE / PRIMARY KEY / EXCLUSION; foreign keys are governed by `foreign_key` separately.
+By default, `plan` and `apply` do not drop tables, views, enums, domains, composite types, columns, constraints, foreign keys, or indexes. Use `--allow-drop` to enable dropping specific object types (`all`, `table`, `view`, `enum`, `domain`, `composite_type`, `column`, `constraint`, `foreign_key`, `index`, `policy`). Also available as `$PISTA_ALLOW_DROP`. `constraint` covers CHECK / UNIQUE / PRIMARY KEY / EXCLUSION; foreign keys are governed by `foreign_key` separately. `composite_type` also gates `DROP ATTRIBUTE` on a composite type.
 
 ```bash
 # Allow all drops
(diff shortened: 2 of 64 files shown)


winebarrel:
## Review

Reviewed against a live PostgreSQL 16 instance: full test suite (`go test -p 1 ./...`, all passing) plus hands-on `pista plan` / `apply` / `dump` runs.

The design follows the existing enum / domain patterns closely, and the wiring is complete across catalog / model / diff / parser / dump / filter / remap / toposort / drop policy. Dependency ordering (composite type before table on create, the reverse on drop) is correct against a real database. The dump → plan round trip is clean, including quoted identifiers, array types, and type modifiers.

One item is a bug that needs fixing; the rest are documentation and companion artifacts.

### Needs fixing

**1. Plan never converges with an unqualified `COLLATE` (reproduced on a live database)**

```sql
CREATE TYPE public.ct AS (name text COLLATE "C");
```

After applying, every subsequent plan emits the same statement forever:

```
ALTER TYPE public.ct ALTER ATTRIBUTE name TYPE text COLLATE "C";
```

The catalog side (`catalog/composite_types.go:107`) returns `pg_catalog.C` via `nspname || '.' || collname`, while the parser side (`parser/parser.go:956`) stores what the user wrote, `C`. So `equalPtr(wa.Collation, da.Collation)` at `diff/composite_types.go:152` never matches. `pista dump` emits `pg_catalog."C"`, so dump-derived schemas do not hit this, but `COLLATE "C"` is the ordinary way to write it by hand.

Domains have the same asymmetry, but they fail loudly with `cannot change collation of domain ...`. A composite type instead emits a silent no-op DDL on every run, which is worse. Resolving an unqualified collation name in the parser (defaulting built-ins to `pg_catalog`), or comparing only the trailing component, would close it.

`testdata/apply/composite_type_collation.yml` passes because it uses the qualified `pg_catalog."C"`. Adding one `verify_no_drift: true` fixture that uses `COLLATE "C"` would prevent a regression.

**2. `ALTER ATTRIBUTE ... TYPE` always fails while a table column uses the type**

PostgreSQL does not allow changing an attribute's type while a table column references the composite type, and `CASCADE` does not help. The plan looks fine but apply fails:

```
ALTER TYPE public.addr ALTER ATTRIBUTE city TYPE character varying(50);
ERROR: cannot alter type "addr" because column "p.a" uses it (SQLSTATE 0A000)
```

`ADD ATTRIBUTE` / `DROP ATTRIBUTE` / `RENAME ATTRIBUTE` all succeed under the same conditions, so the restriction is specific to type changes. Nothing is destroyed since the transaction rolls back, but this is not mentioned in the PR's scope notes or the README, and no fixture covers it. Detecting it at plan time with a clear error (as domain base-type changes already do), or at minimum documenting the limitation, would help.

### Missing companion artifacts

**3. No CHANGELOG.md entry.** Comparable feature additions (#296 sequence management, #291, etc.) all have one.

**4. No `test/scenario/*.test.sh`.** The closest precedent, #296, shipped `sequence.test.sh` alongside the feature.

**5. Stale sample output in the docs.** `ObjectCount.Summary()` now includes `composite type`, but `README.md:261`, `README.md:423`, `README.md:542`, and `getting-started.md:80,95,108` still show `(1 table, 0 views, 0 enums, 0 domains, 0 sequences)`.

**6. No TODO.md entry.** Per repository convention, the deferred items (attribute reordering producing no diff, and the restriction in 2 above) belong in TODO.md with the originating PR.

### Minor

- Renaming a composite type emits a redundant `ALTER TABLE ... SET DATA TYPE`, even though PostgreSQL updates the column type automatically on rename. Harmless, and the second plan is clean. Same class as "Table rename: cross-table dependents" in TODO.md.
- Writing both `CREATE TYPE public.x AS ENUM (...)` and `CREATE TYPE public.x AS (...)` in one file is not rejected, because `setUnique` operates on separate maps. Both `CREATE TYPE` statements are emitted and apply fails. The gap already exists between enums and domains, but another type kind widens it.
- Gating `DROP ATTRIBUTE` on `composite_type` rather than `column` reads as intentional, since the README states it explicitly.

Only item 1 needs a code change; 2 through 6 are documentation and companion artifacts.

---

winebarrel:
Thanks for the thorough review. All items addressed in `bd9a29d`.

**1. Unqualified `COLLATE` never converges (fixed).** Attribute collations are now compared by name only (`equalCollation` / `collationName` in `diff/composite_types.go`), so the catalog's `pg_catalog.C` matches a hand-written `C`. Added `testdata/apply/composite_type_collation_unqualified.yml` using `COLLATE "C"` with `verify_no_drift: true` to lock it in.

**2. `ALTER ATTRIBUTE ... TYPE` fails while a table column uses the type (documented).** Plan-time detection needs cross-object awareness the composite diff does not have, so for now it is documented in the README "Supported Objects" note and in TODO.md. Kept as a deferred item rather than adding the coupling.

**3. CHANGELOG.md (added).** New `## [1.20.0]` entry.

**4. Scenario test (added).** `test/scenario/composite_type.test.sh` walks create -> alter attributes -> rename -> comment, drift-free at each step. Passes in the full scenario run.

**5. Stale sample output (fixed).** `README.md:261/423/542` and `getting-started.md:80/95/108` now include `composite type` in the `Summary()` line.

**6. TODO.md (added).** Entries for the item-2 restriction and the attribute-reordering no-diff behavior.

**Minor.** The redundant `ALTER TABLE ... SET DATA TYPE` you saw on rename turned out to be a broader pre-existing bug: a schema-qualified user-type column (enum/domain/composite) drifts on every plan because `format_type` returns the type unqualified for a type in the search path while the parser keeps the qualified form. Fixed in `equalTypeName` by stripping the table's own schema from both sides, so `home public.addr` on a table in `public` no longer emits a no-op `SET DATA TYPE`. The rarer cross-schema case (type in a different search-path schema than its table) still drifts and is recorded in TODO.md. The duplicate `CREATE TYPE ... AS ENUM` vs `AS (...)` name collision is unchanged; noted as a pre-existing gap.

winebarrel:
## Re-review (bd9a29d)

Re-checked against a live PostgreSQL 16 instance: full test suite (`go test -p 1 ./...`, all passing), `go vet`, the new scenario test, and hands-on verification of the fixed behaviors. (`golangci-lint` could not run here — the installed binary is built with Go 1.25 against a 1.26 target.)

### Confirmed addressed

| Item | Status |
|---|---|
| 1. Unqualified `COLLATE` drift | Fixed. `equalCollation` (`diff/composite_types.go:81`) compares the trailing component. Verified on a live database that `COLLATE "C"` now settles to `-- No changes` |
| 2. `ALTER ATTRIBUTE ... TYPE` restriction | Documented in the README's Supported features section and in TODO.md |
| 3. CHANGELOG | `1.20.0` entry added |
| 4. Scenario test | `composite_type.test.sh` added. Ran it, 4/4 pass. `run_step` verifies plan → apply → no drift at every step, so the coverage is substantive |
| 5. Stale sample output | Updated in all three `README.md` spots and all three `getting-started.md` spots |
| 6. TODO.md | Three entries added |

Beyond the review, the qualified user-type drift on table columns was also fixed via `equalTypeName` + `stripTypeSchema` (`diff/tables.go:1336`). `schemaOf` handles quoted schema names correctly and `stripTypeSchema` preserves a trailing `[]`. It benefits enum and domain columns at the same time. Verified that `home public.addr` now settles.

### One issue remains

**The schema-qualification normalization was not applied to composite attributes.**

`diff/tables.go` was fixed, but `diff/composite_types.go:172` still does a bare `wa.TypeName != da.TypeName`. An attribute whose type is another user-defined type (composite, enum, or domain) written schema-qualified never converges.

```sql
CREATE TYPE public.inner_t AS (x text);
CREATE TYPE public.outer_t AS (i public.inner_t);
CREATE TYPE public.kind    AS ENUM ('home');
CREATE TYPE public.withenum AS (k public.kind);
```

After applying, every subsequent plan emits the same statements:

```
ALTER TYPE public.outer_t ALTER ATTRIBUTE i TYPE public.inner_t;
ALTER TYPE public.withenum ALTER ATTRIBUTE k TYPE public.kind;
```

And if a table column uses `outer_t`, that statement hits the restriction from item 2, so apply fails on every run. The loop is no longer just noise at that point.

`testdata/plan/create_composite_type_enum_chain.yml` does use `kind public.kind`, but it is plan-only from an empty database, so it cannot catch this.

The fix should be a drop-in: `equalTypeName(wa.TypeName, da.TypeName, schemaOf(fqcn))`, both already in package `diff`. A composite attribute cannot be `serial`, so the serial normalization inside `equalTypeName` is harmless here. For coverage, a `verify_no_drift: true` fixture mirroring `composite_type_qualified_column_no_drift.yml` would do it.

### Aside (pre-existing, out of scope for this PR)

Noticed while testing: apply sets `search_path` only just before `-- pista:execute` statements (`apply.go:151`), so an unqualified user-type reference in desired SQL fails at apply for any schema other than `public`.

```
-n app  ->  CREATE TABLE app.people (id integer, home addr);
            ERROR: type "addr" does not exist (SQLSTATE 42704)
```

This reproduces identically with an enum, so it predates this PR. Not something to fix here, but perhaps a TODO.md candidate.

---

With the remaining item fixed, this looks good to merge.

---

winebarrel:
Both items addressed.

**Composite attribute type normalization (the remaining issue).** `diffCompositeType` now compares attribute types with `equalTypeName(wa.TypeName, da.TypeName, schemaOf(fqcn))` (`diff/composite_types.go`), so an attribute whose type is a schema-qualified user type (composite/enum/domain) no longer emits a no-op `ALTER ATTRIBUTE` on every plan. Verified on a live database that the nested `outer_t` / `withenum` example settles to `-- No changes`. Added `testdata/apply/composite_type_qualified_attribute_no_drift.yml` (`verify_no_drift: true`) and a unit test.

**Aside (apply search_path) — fixed rather than deferred.** `apply` set `search_path` only just before `-- pista:execute`, so an unqualified user-type reference in a column/attribute definition failed at apply for a non-public target schema. Moved the `search_path` setup ahead of the managed DDL loop, so both the managed DDL and the execute statements resolve unqualified user types. Object names in the generated DDL are always schema-qualified, so this only affects name resolution inside definitions, not where objects are created; it is executed but not written to the output writer, so `applied_sql` is unchanged. `apply -n app` with `home addr` now succeeds and re-plans clean. Added `TestApply_UnqualifiedUserTypeInNonPublicSchema` (the YAML harness always targets public, so this needs a Go test).

Full suite (`go test -p 1 ./...`), scenario test, and `golangci-lint` all pass.

winebarrel:
## Re-review (2324721)

Full test suite (`go test -p 1 ./...`) and `go vet` pass, and I verified the fixed behaviors against a live PostgreSQL 16 instance.

### Confirmed addressed

- **Composite attribute schema normalization.** `diff/composite_types.go:172` now uses `equalTypeName(wa.TypeName, da.TypeName, schemaOf(fqcn))`. Verified no drift with composite, enum, domain, and array attribute types. The `composite_type_qualified_attribute_no_drift.yml` fixture and the unit test cover it.
- **TODO.md** — the cross-schema entry was broadened to include composite attributes.
- The **`search_path`** item I flagged as out of scope was fixed too. Dropping the `len(client.Schemas) > 0` guard is safe, since `validateSchemas()` at `apply.go:49` already rejects an empty list. Using a Go test (`TestApply_UnqualifiedUserTypeInNonPublicSchema`) rather than a fixture is the right call here, since the fixture harness always targets `public` — consistent with the guidance in CLAUDE.md.

### New: the search_path change introduces a regression

`SET search_path TO <target schemas>` *replaces* the session default (`"$user", public`) instead of extending it. For any non-public target schema, an unqualified reference to an object in `public` now fails at apply where it previously succeeded. Verified against both bd9a29d and 2324721.

With `-n app`, a column typed as a domain that lives in `public`:

```sql
CREATE TABLE app.t1 (id integer, c pubdm);
```
- bd9a29d: applies successfully
- 2324721: `ERROR: type "pubdm" does not exist (SQLSTATE 42704)`

With `-n app`, a `DEFAULT` calling a function in `public`:

```sql
CREATE TABLE app.t2 (id integer, c text DEFAULT pubfn());
```
- bd9a29d: applies successfully
- 2324721: `ERROR: function pubfn() does not exist (SQLSTATE 42883)`

This is the common real-world layout where extensions (citext, hstore, ltree, postgis, uuid-ossp) are installed into `public` while the managed schema is something else.

Extending rather than replacing looks like the natural fix: `SET search_path TO <schemas>, public`, or read the session's current `search_path` and prepend the target schemas. `pg_catalog` is implicitly searched first either way, so built-in types are unaffected. A test for "unqualified reference to a `public` object from a non-public target schema" would lock the behavior in.

If the replacing semantics are intentional and stay, the CHANGELOG needs to call it out as a breaking change — the current wording reads as a pure addition.

### Minor

The `SET search_path` is deliberately not written to the output writer, so piping `pista plan -n app` into `psql` no longer reproduces what `pista apply` does (it hits the errors above). Piping plan output is a common workflow, so a note in the README, or emitting it as a comment line, would help.

---

With that one item resolved, this looks good to merge.

---

winebarrel:
Good catch, fixed in `8d4e4cd`.

**search_path now extends instead of replacing.** `apply` sets `SET search_path TO <target schemas>, public` (public is appended unless it is already a target), so an unqualified reference to a `public` object from a non-public target schema resolves again. Verified on a live database with `-n app`: a column typed as a `public` domain (`c pubdm`) and a `DEFAULT public.pubfn()` both apply and re-plan clean. Since it extends rather than replaces, it is not a breaking change, and the CHANGELOG wording was updated to say "target schemas plus `public`". Added `TestApply_PublicObjectFromNonPublicSchema` covering both the domain and function references.

**Minor.** Added a README note under `plan`: `apply` sets `search_path` but `plan` output omits the `SET`, so piping `pista plan -n <schema>` into `psql` for a non-public schema may fail on an unqualified reference; qualify it or run `pista apply`. Kept `plan` output unchanged rather than emitting the `SET` line, since that would appear in every plan and change the existing output contract.

Full suite (`go test -p 1 ./...`) and `golangci-lint` pass.

winebarrel:
## Re-review (8d4e4cd) — LGTM

Everything raised is addressed, and I have nothing further.

### Verified

- **search_path regression fixed** (`apply.go:142-161`) — `public` is now appended rather than the path being replaced. Re-checked the two cases that broke on 2324721:
  - `-n app` with `c pubdm` (unqualified domain in `public`) — applies
  - `-n app` with `DEFAULT pubfn()` (function in `public`) — applies
  - `home addr` (unqualified composite type in the target schema) resolves in the same apply, and the follow-up plan reports `-- No changes`
- **No duplicate entry** — the branch skips the append when `public` is already among the target schemas, so the default `-n public` still emits `SET search_path TO public`.
- **CHANGELOG / README** — the entry now describes the extend semantics, and the README note near line 156 covers piping `plan` output into `psql`. That closes the minor point from the last round.
- **Tests** — `TestApply_PublicObjectFromNonPublicSchema` exercises both a domain and a function, which pins the regression.

### Extra edge cases checked (all fine)

- A database with no `public` schema at all — PostgreSQL ignores a missing schema in `search_path`, so apply succeeds.
- A schema name needing quotes (`-n 'My App'`) — quoted correctly via `model.Ident`; apply and re-plan are both clean.

### Overall

- `go test -p 1 ./...` all passing, `go vet` clean.
- Ran the full CLI scenario suite (19 scripts) — all pass, so the `search_path` change does not disturb the existing scenarios.

The six items from the first round, the composite attribute normalization from the second, and the `search_path` regression from the third are all resolved.

(Posted as a comment rather than an approving review — GitHub rejects a self-approval, and this session authenticates as the PR author.)

---

PR #336, opened 2026-07-31
Add --try-tx to wrap apply in a transaction when possible

winebarrel:
--with-tx fails when the diff contains CREATE/DROP INDEX CONCURRENTLY,
because PostgreSQL cannot run those inside a transaction. That leaves no
way to keep a per-index CONCURRENTLY opt-in and still get a transaction
on the runs that do not touch indexes.

--try-tx (env $PISTA_TRY_TX) behaves like --with-tx, except that such a
diff runs without a transaction instead of failing, and the output
records "-- Transaction skipped: plan contains CONCURRENTLY index DDL".
The decision follows the generated diff, so an index carrying
-- pista:concurrently that is not being changed still gets a
transaction.

--with-tx is unchanged and still errors, so an explicit all-or-nothing
request is never silently weakened. The two flags are mutually
exclusive; unlike --with-tx, --try-tx can be combined with
--force-index-concurrently.

Claude-Session: https://claude.ai/code/session_01V5GQnLhxyfKP28s5hiZE94

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.21.0] - 2026-07-31
+
+* Add `--try-tx` (env `$PISTA_TRY_TX`) to `apply`. It wraps the apply in a transaction like `--with-tx`, but a diff containing `CREATE/DROP INDEX CONCURRENTLY` runs without a transaction instead of failing, and the output records `-- Transaction skipped: plan contains CONCURRENTLY index DDL`. The decision follows the generated diff, so a `-- pista:concurrently` index that is not being changed still gets a transaction. `--with-tx` is unchanged and still errors on such a diff; the two flags are mutually exclusive. Unlike `--with-tx`, `--try-tx` can be combined with `--force-index-concurrently`.
+
 ## [1.20.0] - 2026-07-30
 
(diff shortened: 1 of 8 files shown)


winebarrel (commented):
Reviewed the full diff at e3064b8. The implementation looks correct and I found no blocking issues.

## What I verified locally

- `go build ./...` and `go vet ./...` pass.
- All `TestApply*` tests pass against PostgreSQL 16, including the 8 new `--try-tx` tests.
- `test/scenario/index_concurrently.test.sh` passes (7/7, including new steps 06 and 07).
- The xor groups work as intended, including via env vars: `PISTA_WITH_TX=1` + `--try-tx` and `PISTA_TRY_TX=1` + `PISTA_WITH_TX=1` both fail with "--with-tx and --try-tx can't be used together", while `--try-tx --force-index-concurrently` is accepted.

## Design notes

- The core change is minimal and easy to reason about: a single `withTx := options.WithTx || (options.TryTx && !result.HasConcurrentlyIndex)` decision point, with `--with-tx` semantics untouched. The explicit all-or-nothing contract of `--with-tx` is preserved.
- Basing the decision on `result.HasConcurrentlyIndex` (the generated diff) rather than directive presence is the right call, and `TestApply_TryTx_DirectiveIndexUnchanged_UsesTransaction` pins that behavior down.
- The `-- Transaction skipped` comment is only reachable after the early return for empty diffs (apply.go:91), so a no-op apply stays clean, as `TestApply_TryTx_NoChanges_OmitsTransactionComments` asserts.
- `HasConcurrentlyIndex` covers both table and materialized view indexes (diff_all.go:220), so `--try-tx` inherits the matview handling for free.
- Test coverage is thorough: skip path, tx path, rollback, no-op, unchanged directive index, concurrently-pre-SQL, both index-concurrently flag interactions, and xor enforcement. Docs (README, directives.md, CHANGELOG) are consistent with the behavior.

## Minor, non-blocking

- The skip message says "plan contains CONCURRENTLY index DDL" in `apply` output. "diff" might read slightly more naturally in that context, but "plan" matches the vocabulary used elsewhere, so fine either way.
- Optional: a `--try-tx` test with a materialized view index would pin the matview path explicitly, though it shares the `HasConcurrentlyIndex` flag with the already-tested `--with-tx` matview case, so the marginal value is small.

---

winebarrel (commented):
Re-reviewed at 29743fc. The new commit adds `TestApply_TryTx_MatviewIndex_SkipsTransaction`, which addresses the one optional suggestion from my previous review by pinning the materialized view index path (`viewDiff.HasConcurrently`) under `--try-tx` explicitly.

Verified locally: `go vet ./...` passes and the full `TestApply*` suite passes against PostgreSQL 16, including the new test. The test follows the same structure as the existing `--try-tx` tests and asserts the right three things (CONCURRENTLY DDL emitted, transaction skipped, no transaction started).

Nothing further from me. LGTM.

---

PR #338, opened 2026-07-31
Quote type_func_name keywords, build domain CHECK definitions from the parse tree

winebarrel:
## Quote `type_func_name` keywords used as identifiers

PostgreSQL documents 23 keywords as "reserved (can be function or type name)". They are valid as a type or function name but not as a table, column, index, constraint, sequence, type, or policy name:

```
authorization binary collation concurrently cross current_schema freeze full
ilike inner is isnull join left like natural notnull outer overlaps right
similar tablesample verbose
```

`quoteIdent` only quoted reserved keywords, so these were emitted bare. `dump` produced SQL that could not be parsed again, and `plan` / `apply` produced DDL that PostgreSQL rejected with a syntax error.

```
CREATE TABLE public.t (left integer)      => syntax error at or near "left"
CREATE INDEX full ON public.t (id)        => syntax error at or near "full"
CREATE TYPE binary AS ENUM ('a')          => syntax error at or near "binary"
```

Quoting now follows the grammar's `ColId` production, which accepts a bare identifier, an unreserved keyword, or a `col_name` keyword and nothing else. It is written as an allow list so a keyword category added upstream is quoted until it is reviewed.

### Tests

`model/testdata/keywords.txt` carries all 491 PostgreSQL keywords as a corpus. The test derives each category through `pg_query.Scan` at run time, so the file cannot drift out of sync with the parser, and checks that `Ident`'s output parses in every position an identifier is emitted: 47 bare positions, 29 schema-qualified ones, and the three-part form used by column comments. The position list was collected from the `Ident` call sites in `model/`, `diff/`, and `apply.go`.

The corpus size is pinned, so a pg_query_go upgrade that adds keywords fails the test rather than leaving them untested. `make keywords` regenerates the file from the `kwlist.h` bundled with the version in `go.mod`.

Fixtures cover CREATE, ALTER, DROP, RENAME, and `--bulk-alter` over enums, domains, sequences, tables, columns, constraints, indexes, policies, and comments. The two apply fixtures set `verify_no_drift`.

`none` is skipped in the policy role position only: PostgreSQL rejects `CREATE ROLE none` quoted or not, so no policy can name it.

## Build domain CHECK definitions from the parse tree

Found while following up on the review of the change above. `extractCheckDefFromDeparsed` read a domain's `CHECK` definition out of the deparsed statement and ended it at the next `" CONSTRAINT "`. That text also occurs inside string literals:

```sql
CREATE DOMAIN public.label AS text
    CONSTRAINT label_not_keyword CHECK (VALUE <> ' CONSTRAINT ');
```

came out as `CHECK (value <> '`, and the emitted DDL did not parse.

The function already had a fallback that deparses `con.RawExpr`, taken whenever the marker did not match. It produces the same definition in every case the suite covers and does not depend on the surrounding text, so it is now the only path; the whole-statement deparse it needed goes away with it. `testdata/plan/domain_check_with_constraint_literal.yml` and the matching apply fixture cover the truncation.

## Notes

#339 also adds its entry under `[1.21.1]`, so the second merge will need the bullets combined under one heading.


winebarrel:
## Review

### Overview

This PR fixes `quoteIdent` so that PostgreSQL's 23 `type_func_name` keywords (`left`, `right`, `full`, `inner`, `join`, ...) are quoted when used as identifiers. Previously only `RESERVED_KEYWORD` was quoted, so these names were emitted bare and the resulting DDL was rejected by PostgreSQL (and `dump` output could not be re-parsed). The new rule is an allow list mirroring the grammar's `ColId` production: bare identifiers, unreserved keywords, and `col_name` keywords stay unquoted; everything else gets quotes.

### Correctness

- The allow list matches `ColId` exactly, which is the right production for every position pistachio emits identifiers in (table, column, index, constraint, sequence, type, policy names). Quoting is always valid where an identifier is accepted, so quoting `type_func_name` keywords cannot regress the positions where they were previously legal.
- The allow-list direction is a good defensive choice: a keyword category added upstream defaults to quoted (safe) rather than bare (potentially broken).
- Verified locally: `go vet ./model/` and `go test ./model/` pass on the PR head.
- Side benefit worth noting: this also brings `Ident` closer to pg_query's deparser output for these keywords. For example, `extractCheckDefFromDeparsed` in `parser/parser.go` builds a `CONSTRAINT <Ident(name)> ` marker and searches it in deparsed SQL; before this change the marker (`CONSTRAINT left`) could not match the deparser's `CONSTRAINT "left"`, silently falling into the raw-expression fallback.

### Test quality

The test design is strong:

- I confirmed `model/testdata/keywords.txt` is byte-identical to the keyword list in pg_query_go v6.2.2's bundled `kwlist.h` (491 keywords).
- Deriving each keyword's category through `pg_query.Scan` at run time (instead of recording categories in the file) means classifications cannot drift from the parser. The exhaustive parse check of every keyword in every emit position (`identPositions` / `qualifiedIdentPositions` / `columnIdentPositions`) is what actually proves the `ColId` claim, and the "must not be quoted" assertions ensure the change is minimal rather than blanket-quoting.
- The plan/apply/dump YAML fixtures cover create, alter, drop, rename, and `--bulk-alter` paths, and the apply fixtures with `verify_no_drift: true` prove a real server accepts the quoted DDL and that it round-trips. This follows the repo convention of preferring fixtures over Go tests where possible.

### Minor suggestions (non-blocking)

1. **Corpus completeness on upgrade.** The runtime classification prevents *mis*classification drift, but a pg_query_go upgrade that adds new keywords would leave them silently missing from `keywords.txt`, and `require.Greater(t, len(keywords), 400)` will not catch a handful of additions. Consider pinning the exact count (`require.Len(t, keywords, 491)`) so an upgrade forces a corpus refresh, or adding a small generator/check against the module's bundled `kwlist.h`.
2. **`col_name` keywords vs. `quote_identifier`.** PostgreSQL's own `quote_identifier` (and pg_query's deparser) quotes `col_name` keywords too (`"int"`, `"char"`), while `Ident` leaves them bare. The tests prove bare is grammatically valid in every emitted position, so this is fine, but it means `Ident` output still differs from deparser output for those names (e.g. the `extractCheckDefFromDeparsed` marker misses for a domain constraint named `int` and takes the fallback path). Pre-existing behavior, just noting it since the new test now cements it.

### Verdict

Well-scoped fix with an unusually thorough, self-verifying test suite. LGTM.

---

winebarrel:
Thanks. Suggestion 1 is addressed in cabde28; suggestion 2 turns out to be harmless, but chasing it surfaced a separate bug.

## 1. Corpus completeness on upgrade

Both halves of the suggestion, since the count alone does not say how to fix a failure:

* `require.Len(t, keywords, 491, "keyword corpus is stale; run make keywords")`
* `scripts/gen-keywords.sh` and a `make keywords` target regenerate the corpus from the `kwlist.h` bundled with the pg_query_go version in `go.mod`, and record that version in the file header.

## 2. `col_name` keywords vs `quote_identifier`

The fallback path produces the same result, so nothing is lost by taking it. Parsing `CONSTRAINT "int" CHECK (VALUE > 0)` and `CONSTRAINT pos CHECK (VALUE > 0)` both yield `CHECK (value > 0)`, and the same holds with a regex operand, an `IN` list, and a second constraint following.

More than that, the marker path is the weaker of the two. It cuts the definition at the next occurrence of the literal `" CONSTRAINT "`, which also matches inside a string:

```sql
CREATE DOMAIN public.d AS text CONSTRAINT pos   CHECK (VALUE <> ' CONSTRAINT ');
CREATE DOMAIN public.d AS text CONSTRAINT "int" CHECK (VALUE <> ' CONSTRAINT ');
```

```
name="pos"  def="CHECK (value <> '"          <- truncated
name="int"  def="CHECK (value <> ' CONSTRAINT ')"
```

So the constraint that misses the marker is the one that comes out right. `extractCheckDefFromDeparsed` looks redundant: deleting it and always deparsing `con.RawExpr` passes the entire suite, including `make test-scenario`.

That is a separate defect in a different function, so I left this PR alone. Happy to open a PR removing `extractCheckDefFromDeparsed`, with a fixture for the truncation case, if you want it.

---

winebarrel:
Correction to my previous comment: the `extractCheckDefFromDeparsed` removal is in this PR after all, as 88e98e8, not a separate one.

* `extractCheckDefFromDeparsed` and the whole-statement deparse it needed are gone. Domain `CHECK` definitions are built from `con.RawExpr`, which is the path a definition that missed the marker already took.
* `testdata/plan/domain_check_with_constraint_literal.yml` and the matching apply fixture (`verify_no_drift`) cover the truncation. Without the change the plan emits `CHECK (value <> '`.
* Full suite, `make test-scenario`, and coverage (93.3%, unchanged) all pass.

The PR body has a section on it.

---

winebarrel:
## Re-review (cabde28, 88e98e8)

Both follow-up commits look good and fully address the points from the previous review.

### `cabde28` — Pin the keyword corpus size and add a regeneration target

- `require.Len(t, keywords, 491, ...)` now makes a pg_query_go upgrade that adds keywords fail the test instead of silently leaving them untested — exactly the concern raised before.
- Verified `scripts/gen-keywords.sh` locally: it reproduces the committed `keywords.txt` byte-for-byte from the bundled `kwlist.h` (491 keywords), and embedding the module version in the file header is a nice touch.
- Tiny nit, take it or leave it: after running `make keywords` on an upgrade, the pinned `491` in the test also needs bumping, but the failure message only says "run make keywords". Mentioning the constant in the message (or in the script's output) would save the future reader one round trip.

### `88e98e8` — Build domain CHECK definitions from the parse tree

- Good catch turning the deparser-divergence observation into a real bug fix: the old marker scan cut the definition at the next ` CONSTRAINT `, which also matches inside a string literal, producing unparseable DDL for `CHECK (VALUE <> ' CONSTRAINT ')`.
- Making the `RawExpr` deparse the only path is the right simplification — it was already the fallback, it cannot be confused by surrounding text, and it lets the whole-statement `Deparse` call and `extractCheckDefFromDeparsed` disappear entirely (net −38 lines in `parser/parser.go`). No existing fixtures changed, which supports the claim that both paths produced identical output for everything the suite covers.
- The new plan and apply fixtures cover the truncation case directly, and the apply fixture verifies drift-free round-tripping against a real server.

### Verification

- `go build ./...`, `go vet ./...`, and `go test ./model/ ./parser/` pass locally on the PR head.
- CI is green on the PostgreSQL 15/16/17/18 test matrices, `ascii`, and CodeQL; only the Windows job was still in progress at review time.
- The CHANGELOG note about combining the `[1.21.1]` bullets with #339 at merge time is a sensible heads-up.

LGTM.

---

PR #339, opened 2026-07-31
Split collation names on the quoting, not on every dot

winebarrel:
A collation name can contain a dot: `C.utf8` and `en_US.utf8` are the ones PostgreSQL ships. The catalog joined the schema and the name with a dot, and the model split the result back on every dot, so a domain or a composite attribute came out as `COLLATE pg_catalog."C".utf8`.

```
CREATE DOMAIN d AS text COLLATE pg_catalog."C".utf8;
ERROR:  cross-database references are not implemented: pg_catalog.C.utf8
```

Collation names are now carried in quoted SQL form, ready to follow `COLLATE`. The catalog quotes each part with `quote_ident`, the parser builds the same form with `model.Ident`, and the emitters print it verbatim. Comparison goes through `model.SplitQualifiedName` (moved from `parser/directive.go`), which respects quoting, and compares the schema only when both sides carry one, since the catalog always qualifies while a desired schema may not.

Three more defects fall out of the same change:

* Column collations lost their schema. `catalog/columns.go` selected `collname` alone, so a dump emitted `COLLATE mycoll`, which resolved only when the collation's schema was on `search_path`.
* A column collation written schema-qualified in the desired schema produced a no-op `ALTER COLUMN ... SET DATA TYPE` on every plan, because the parser kept the qualified name while the catalog kept the bare one.
* A domain collation written unqualified failed the plan with `cannot change collation of domain ...`, because the two forms were compared as plain strings. Composite attributes already had a name-only comparison; domains and columns now share it.

## Output change

Column collations are now schema-qualified: `COLLATE "C"` becomes `COLLATE pg_catalog."C"`. Domains and composite types were already qualified, so the three now agree, and it matches what `pg_dump` emits. Three fixtures and one catalog unit test were updated for it.

## Tests

Written first and confirmed failing against `main`: the dump fixture emitted `public.my.coll`, the apply fixture failed against PostgreSQL with the cross-database reference error, and the unqualified-collation plan fixture failed with `cannot change collation`.

Covered per object kind (table column, domain, composite attribute) and per operation (create, add, change, drop, no-drift):

* `testdata/dump/collation_dotted_name.yml`, `omit_schema_collation.yml`
* `testdata/apply/collation_dotted_name.yml`, `collation_dotted_name_change.yml` (both `verify_no_drift`)
* `testdata/plan/collation_dotted_name_no_drift.yml`, `collation_unqualified_no_drift.yml`, `collation_dotted_name_composite_change.yml`, `collation_domain_change_error.yml`
* `diff/collation_test.go` for `equalCollation`, `model/util_test.go` for `SplitQualifiedName`

`--omit-schema` keeps the collation's schema, since a collation is a separate object pistachio does not manage. `omit_schema_collation.yml` pins that.

The fixtures create `public."my.coll"` with `CREATE COLLATION ... FROM "C"` rather than relying on a system locale, so they run the same on PostgreSQL 15 through 18.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [1.21.1] - 2026-07-31
 
+* Fix broken DDL for a collation whose name contains a dot, such as the `C.utf8` and `en_US.utf8` family. The schema-qualified name was split on every dot, so a domain or composite attribute was emitted as `COLLATE pg_catalog."C".utf8`, which PostgreSQL rejects as a cross-database reference. Collation names are now split on the quoting instead.
+
+* Fix `dump` dropping the schema from a column collation. `COLLATE mycoll` resolved only when the collation's schema was on `search_path`; the dump now emits `COLLATE public.mycoll`, matching what domains and composite types already did. A column collation written schema-qualified in the desired schema also no longer produces a no-op `SET DATA TYPE` on every plan.
+
 * Fix a domain `CHECK` definition being truncated when the expression contains the text ` CONSTRAINT `, as in `CHECK (VALUE <> ' CONSTRAINT ')`. The definition was read out of the deparsed statement and cut at the next occurrence of that text, which also matches inside a string literal. It is now built from the parse tree, the same path a definition that did not match the marker already took.
 
 * Fix broken DDL for an identifier that is one of PostgreSQL's 23 `type_func_name` keywords (`left`, `right`, `full`, `inner`, `binary`, `natural`, and so on). They are valid as a type or function name but not as a table, column, index, constraint, sequence, type, or policy name. Only reserved keywords were quoted, so these were emitted bare: `dump` produced SQL that could not be parsed again, and `plan` / `apply` produced DDL that PostgreSQL rejected with a syntax error. Quoting now follows the grammar's `ColId` rule, which accepts a bare identifier, an unreserved keyword, or a `col_name` keyword.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 * Fix `dump` dropping the schema from a column collation. `COLLATE mycoll` resolved only when the collation's schema was on `search_path`; the dump now emits `COLLATE public.mycoll`, matching what domains and composite types already did. A column collation written schema-qualified in the desired schema also no longer produces a no-op `SET DATA TYPE` on every plan.
 
+* Fix `plan` failing with `cannot change collation of domain ...` when the desired schema writes a domain collation unqualified, such as `COLLATE "C"` against the catalog's `pg_catalog."C"`. The two forms are now recognized as the same collation, as they already were for composite attributes.
+
 * Fix a domain `CHECK` definition being truncated when the expression contains the text ` CONSTRAINT `, as in `CHECK (VALUE <> ' CONSTRAINT ')`. The definition was read out of the deparsed statement and cut at the next occurrence of that text, which also matches inside a string literal. It is now built from the parse tree, the same path a definition that did not match the marker already took.
 
 * Fix broken DDL for an identifier that is one of PostgreSQL's 23 `type_func_name` keywords (`left`, `right`, `full`, `inner`, `binary`, `natural`, and so on). They are valid as a type or function name but not as a table, column, index, constraint, sequence, type, or policy name. Only reserved keywords were quoted, so these were emitted bare: `dump` produced SQL that could not be parsed again, and `plan` / `apply` produced DDL that PostgreSQL rejected with a syntax error. Quoting now follows the grammar's `ColId` rule, which accepts a bare identifier, an unreserved keyword, or a `col_name` keyword.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 * Fix `plan` failing with `cannot change collation of domain ...` when the desired schema writes a domain collation unqualified, such as `COLLATE "C"` against the catalog's `pg_catalog."C"`. The two forms are now recognized as the same collation, as they already were for composite attributes.
 
+* Fix a no-op `SET DATA TYPE` on every plan for a column written with `COLLATE "default"`. The default collation is what a column gets implicitly and the catalog never reports it, so it is now ignored on columns as it already was on domains and composite attributes.
+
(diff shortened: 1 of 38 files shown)


winebarrel:
## Review

### Overview

This PR fixes DDL emission for collations whose names contain a dot (`C.utf8`, `en_US.utf8`): the catalog joined schema and name with a dot and the model split the result on every dot, producing `COLLATE pg_catalog."C".utf8`, which PostgreSQL rejects as a cross-database reference. The fix carries collation names in quoted SQL form end-to-end — the catalog builds it with `quote_ident`, the parser with `model.Ident`, and the emitters print it verbatim — with comparison going through a quote-aware split. Three adjacent defects (column collations losing their schema in dumps, a perpetual no-op `SET DATA TYPE` for qualified column collations, and a false `cannot change collation of domain` error for unqualified domain collations) fall out of the same change.

### Design

- Carrying the already-quoted SQL form is the right simplification: the old code quoted at emit time by re-splitting a joined string, which is exactly where the dot ambiguity lived. Now there is one canonical form and the split/join round trip disappears (`collateClause` and `Domain.SQL` both shrink).
- `equalCollation` comparing the schema only when both sides carry one is the correct tradeoff for a catalog that always qualifies against a desired schema that may not, and the previous composite-attribute comparison already established that convention. The test table covers the tricky cases, including `pg_catalog."C.utf8"` vs `"C"`, which the old dot-split would have wrongly matched as schema `"C"`.
- Moving `splitQualifiedName` to `model.SplitQualifiedName` removes a duplicate implementation; the remaining one-line wrapper in `parser/directive.go` could be dropped in favor of direct calls, but that is cosmetic.
- Keeping the collation's schema under `--omit-schema` is correct — the collation is a referenced object, not one of the dumped objects — and `omit_schema_collation.yml` pins that decision.
- The output change (column collations now schema-qualified, matching domains, composite types, and `pg_dump`) is clearly documented in the CHANGELOG and covered by updated fixtures.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` (clean), and `go test ./model/ ./parser/ ./diff/` all pass locally on the PR head.
- CI is fully green: PostgreSQL 15/16/17/18 matrices, windows, ascii, CodeQL, and both codecov checks.
- Fixture coverage is systematic — per object kind (column, domain, composite attribute) times per operation (create, add, change, drop, no-drift), with `verify_no_drift` on both apply fixtures, and `CREATE COLLATION ... FROM "C"` keeps them locale-independent across PostgreSQL versions.
- The branch is already rebased onto the merged #338, so the `[1.21.1]` CHANGELOG bullets combine under the existing heading and the merge-coordination note in the description is resolved.

### Minor observations (non-blocking)

1. `equalCollation` compares the quoted text, so it silently relies on `quote_ident` (catalog side) and `model.Ident` (parser side) agreeing. They differ in one known corner: `quote_ident` quotes `col_name` keywords (`"int"`, `"char"`) while `model.Ident` leaves them bare, so a collation or schema named after one of those would compare unequal despite naming the same object. Vanishingly unlikely for collation names; comparing dequoted parts would remove the dependency if you ever want to close it.
2. `parseColumnDef` does not skip a literal `COLLATE "default"` the way the domain and composite paths do (`lastPart != "default"`), so a desired schema writing it would plan a `SET DATA TYPE` no-op every run since the catalog never reports the default collation. Pre-existing behavior, untouched by this PR — noting it only because the two sibling paths now differ visibly.

### Verdict

Clean root-cause fix with the same disciplined test approach as #338 (tests written first and confirmed failing against `main`). LGTM.

---

winebarrel:
All three addressed in 62b1e83.

## 1. `equalCollation` depending on the two quoting rules agreeing

Closed. Parts are compared unquoted now, through `model.UnquoteIdent` (moved from `parser/directive.go`), so `quote_ident`'s `"int"` and `model.Ident`'s bare `int` match. Four cases added to the table, including that quoted case stays significant: `public."MyColl"` and `public.mycoll` are still different.

## 2. `COLLATE "default"` on a column

Fixed rather than left as is, since it was a real no-op `SET DATA TYPE` on every plan. The three `COLLATE` clauses in the parser now share `collationFromClause`, so they cannot drift apart again. `testdata/plan/collation_default_column_no_drift.yml` covers all three object kinds.

## 3. The `splitQualifiedName` wrapper

Dropped, along with `unquoteIdent`. `parser/directive.go` calls the model directly, and `TestSplitQualifiedName_SpacesAroundDot` moved into the model test table.

Full suite, `make test-scenario`, `go vet`, `gofmt` all pass. Coverage is 93.4% locally, up from 93.3%; `collationFromClause`, `equalCollation`, `splitCollation`, `SplitQualifiedName`, and `UnquoteIdent` are all at 100%.

---

winebarrel:
## Re-review (62b1e83)

The follow-up commit addresses both observations from the previous review, and cleans up more than asked.

### Unquoted comparison

`equalCollation` now compares parts through `model.UnquoteIdent`, so the catalog's `quote_ident` and the parser's `model.Ident` no longer need to agree on which names to quote — the `col_name`-keyword corner (`public."int"` vs `public.int`) is closed and pinned by a test. The case-folding semantics are right too: an unquoted identifier folds to lower case the way PostgreSQL reads it, so `public.MyColl` matches `public.mycoll` while `public."MyColl"` does not, and both directions are in the test table.

### Shared COLLATE parsing

Extracting `collationFromClause` replaces three near-identical blocks in `parseColumnDef`, `parseCreateDomainStmt`, and `parseCompositeTypeStmt`, and gives columns the `default`-skip the other two paths already had — fixing the perpetual no-op `SET DATA TYPE` for `COLLATE "default"` noted last round. The new `collation_default_column_no_drift.yml` covers all three object kinds, and the CHANGELOG entry is clear.

Consolidating `unquoteIdent` into `model.UnquoteIdent` and deleting the `splitQualifiedName` wrapper (with its test moved into `model/util_test.go`, keeping the spaces-around-dot case) also resolves the cosmetic duplication note. Nice that the directive code now shares the exact same split/unquote pair the diff uses.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt -l` (clean), and `go test ./model/ ./parser/ ./diff/` pass locally on the PR head.
- CI: PostgreSQL 15/16/17/18 matrices, ascii, CodeQL, and both Analyze jobs green; the windows job was still in progress at review time.

### One residual edge case, for the record

`collationFromClause` drops the collation whenever the last AST part is `default`, regardless of schema, so a hypothetical user-defined collation literally named `default` (`CREATE COLLATION public."default" FROM "C"` — the sval is indistinguishable from the unquoted keyword in the parse tree) referenced by a column would now plan a collation drop, where the pre-#339 column path happened to round-trip it. Domains and composite attributes have always behaved this way, so this makes the three paths consistent rather than introducing a new class of problem, and the name is about as pathological as they come. Not worth code, in my view — mentioning it only so the decision is on the record.

### Verdict

Both findings closed with tests, plus a genuine simplification (three copies of COLLATE parsing down to one). LGTM.

---

PR #340, opened 2026-07-31
Stop dropping constraints, partition parents and column defaults

winebarrel:
Five places lost information between the SQL that was read and the model
that was built. Each is covered by a fixture that fails without the fix.

parseAlterTableConstraint returned on its first AT_AddConstraint
subcommand, so only the first constraint of a comma-separated ALTER TABLE
survived. The rest were discarded silently, and the plan then offered to
drop them from the database. It now collects every subcommand. A rename
directive names one old object, so it is rejected when the statement adds
more than one constraint.

The partition CTE in ListTables selected the parent's relname without its
namespace, while the parser qualifies the parent through model.Ident.
Nothing matched: the toposort edge from child to parent was never added,
so both were dropped in declaration order and the parent took its children
with it, leaving the child's own DROP to fail. Dumping was worse. The
unqualified name was emitted verbatim, so a dump of a partitioned table
outside the search_path could not be reloaded, and reattached the child to
a same-named table in another schema when one existed. The parent is now
qualified on the way out of the catalog, and --omit-schema strips it again
along with the rest.

diffTable returned from its partition-child branch before reaching
diffComments, so comments on a partition never diffed. Comments are per
relation and PostgreSQL accepts them on a partition. Column comments also
needed the parser to record them: a partition child declares no columns of
its own, so the entry they attach to did not exist. Only true partition
children get one, since an INHERITS child still goes through the regular
column diff, where a bodyless entry would read as a new column.

A column counted as serial whenever pg_get_serial_sequence found a
sequence, which a plain ALTER SEQUENCE ... OWNED BY is enough to create.
Such a column was dumped as serial and its real default was lost, with no
drift reported. The default now has to draw from that sequence. Both names
go through regclass first, because pg_get_serial_sequence always qualifies
and pg_get_expr does not qualify what is on the search_path.

ALTER SEQUENCE was not parsed at all, so its OWNED BY never marked the
sequence unmanaged the way CREATE SEQUENCE ... OWNED BY does. The catalog
excludes owned sequences, so every plan proposed creating a sequence that
already existed and apply failed on it.

Detaching an owned sequence with OWNED BY NONE still plans a CREATE that
cannot run. The catalog hides owned sequences entirely, so the transition
is invisible; modelling it is left alone here.

Claude-Session: https://claude.ai/code/session_012wjHTFmHXpkJWXAGaZu7bD

### DIFF
--- a/dump.go
+++ b/dump.go
@@ -50,6 +50,12 @@ func (r *DumpResult) tables() *orderedmap.Map[string, *model.Table] {
 		fqtn := t.FQTN()
 		tableName := model.Ident(t.Name)
 		copied.Schema = ""
+		// --omit-schema is restricted to a single schema, so a partition
+		// parent always lives in the schema being stripped.
+		if copied.PartitionOf != nil {
+			parent := strings.TrimPrefix(*copied.PartitionOf, model.Ident(t.Schema)+".")
+			copied.PartitionOf = &parent
+		}
 		if copied.ForeignKeys.Len() > 0 {
 			fks := orderedmap.New[string, *model.ForeignKey]()
 			for _, fk := range copied.ForeignKeys.CollectValues() {
--- a/dump.go
+++ b/dump.go
@@ -50,8 +50,9 @@ func (r *DumpResult) tables() *orderedmap.Map[string, *model.Table] {
 		fqtn := t.FQTN()
 		tableName := model.Ident(t.Name)
 		copied.Schema = ""
-		// --omit-schema is restricted to a single schema, so a partition
-		// parent always lives in the schema being stripped.
+		// Strip the schema only when the parent is in the one being dumped.
+		// --omit-schema allows a single schema, but the parent may sit outside
+		// it, and such a reference has to keep its schema to resolve.
 		if copied.PartitionOf != nil {
 			parent := strings.TrimPrefix(*copied.PartitionOf, model.Ident(t.Schema)+".")
 			copied.PartitionOf = &parent
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -288,15 +288,20 @@ func parseSQLWithSchema(sql string, defaultSchema string) (*ParseResult, error) 
 
 			// RLS toggles and constraint subcommands can coexist in one
 			// ALTER TABLE statement. applyAlterTableRLS picks up only the RLS
-			// subtypes; parseAlterTableConstraint picks up AT_AddConstraint.
+			// subtypes; parseAlterTableConstraints picks up AT_AddConstraint.
 			// They walk the same cmd list independently, so run both.
 			applyAlterTableRLS(as, t)
 
-			con, fk, err := parseAlterTableConstraint(as, defaultSchema)
+			cons, fks, err := parseAlterTableConstraints(as, defaultSchema)
 			if err != nil {
 				return nil, err
 			}
-			if fk != nil {
+			// A rename directive names a single old object, so it cannot be
+			// applied when one statement declares several constraints.
+			if renameFrom != "" && len(cons)+len(fks) > 1 {
+				return nil, fmt.Errorf("renameFrom is ambiguous: ALTER TABLE %s adds %d constraints in one statement", fqtn, len(cons)+len(fks))
+			}
+			for _, fk := range fks {
 				if renameFrom != "" {
 					unquoted := normalizeUnqualifiedDirective(renameFrom)
 					fk.RenameFrom = &unquoted
@@ -304,7 +309,8 @@ func parseSQLWithSchema(sql string, defaultSchema string) (*ParseResult, error) 
 				if err := setUnique(t.ForeignKeys, fk.Name, "foreign key", fk); err != nil {
 					return nil, err
 				}
-			} else if con != nil {
+			}
+			for _, con := range cons {
 				if renameFrom != "" {
 					unquoted := normalizeUnqualifiedDirective(renameFrom)
 					con.RenameFrom = &unquoted
@@ -337,6 +343,9 @@ func parseSQLWithSchema(sql string, defaultSchema string) (*ParseResult, error) 
 			if err := setUnique(sequences, seq.FQN(), "sequence", seq); err != nil {
 				return nil, err
 			}
+
+		case node.GetAlterSeqStmt() != nil:
+			applyAlterSeqOwnedBy(node.GetAlterSeqStmt(), defaultSchema, sequences)
 
 		case node.GetCommentStmt() != nil:
 			cs := node.GetCommentStmt()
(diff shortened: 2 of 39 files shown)


dump.go
```diff
@@ -50,6 +50,12 @@ func (r *DumpResult) tables() *orderedmap.Map[string, *model.Table] {
 		fqtn := t.FQTN()
 		tableName := model.Ident(t.Name)
 		copied.Schema = ""
+		// --omit-schema is restricted to a single schema, so a partition
+		// parent always lives in the schema being stripped.
+		if copied.PartitionOf != nil {
+			parent := strings.TrimPrefix(*copied.PartitionOf, model.Ident(t.Schema)+".")
+			copied.PartitionOf = &parent
+		}
```
winebarrel: The comment overstates the invariant slightly: `--omit-schema` restricts the *dumped* schema to one, but a partition (or `INHERITS`) child in that schema can still have its parent in another schema:

```sql
CREATE TABLE other.parent (...) PARTITION BY RANGE (d);
CREATE TABLE public.child PARTITION OF other.parent FOR VALUES ...;
```

The code degrades gracefully — `TrimPrefix` is a no-op and the parent stays fully qualified, which is exactly the right output — so this is only a comment nit, plus a possible extra fixture (`omit_schema` dump with a cross-schema parent) to pin that behavior down.

---

parser/parser.go
```diff
@@ -288,23 +288,29 @@ func parseSQLWithSchema(sql string, defaultSchema string) (*ParseResult, error)
...
-			con, fk, err := parseAlterTableConstraint(as, defaultSchema)
+			cons, fks, err := parseAlterTableConstraints(as, defaultSchema)
 			if err != nil {
 				return nil, err
 			}
-			if fk != nil {
+			// A rename directive names a single old object, so it cannot be
+			// applied when one statement declares several constraints.
+			if renameFrom != "" && len(cons)+len(fks) > 1 {
+				return nil, fmt.Errorf("renameFrom is ambiguous: ALTER TABLE %s adds %d constraints in one statement", fqtn, len(cons)+len(fks))
```
winebarrel: Tiny wording nit: `renameFrom` is the internal variable name; the user-facing spelling is the `-- pista:renamed-from` directive. Something like `"pista:renamed-from is ambiguous: ..."` would be easier to map back to the schema file for someone who hits this error. (The `error_multiple_constraints_ambiguous_rename.yml` fixture would need the same tweak.)

---

winebarrel (commented):
## Review

Five distinct data-loss bugs fixed between parse and model, each with fixtures that pin the broken behavior. I read the full diff, checked out the branch, ran the whole test suite against a local PostgreSQL 16 (`go test -p 1 ./...` — all green), and probed the new catalog SQL directly against edge cases. Overall this looks solid and ready to land; the inline comments are minor nits only.

### What I verified

**Multi-constraint `ALTER TABLE` (parser/parser.go)** — `parseAlterTableConstraints` now collects every `AT_AddConstraint` subcommand instead of returning on the first, and FK/non-FK constraints can coexist in one statement. The ambiguity guard for `pista:renamed-from` on a multi-constraint statement is the right call, and the RLS coexistence fixture (`alter_table_constraints_with_rls.yml`) covers the two-walkers-one-statement interaction.

**Partition parent qualification (catalog/tables.go)** — joining `pg_namespace` and building `PartitionOf` via `model.Ident(schema, name)` makes the catalog side match the parser side (`parser.go:401` uses the same construction), which is what `toposort/schema.go:97` keys on. The three drop-ordering fixtures (including the reverse-name-order and three-level cases) demonstrate the toposort edge now fires. Quoted-name round-trip is covered too.

**Serial detection (catalog/columns.go)** — I ran the new `is_serial` expression directly against PostgreSQL 16:
- `ALTER SEQUENCE ... OWNED BY` with a `DEFAULT 42` column → correctly *not* serial (the bug being fixed);
- a serial column whose sequence was later **renamed** → still serial (`adbin` stores the OID, so both sides render the current name);
- a serial column in a schema **off `search_path`** → still serial (both sides render schema-qualified);
- the pg_dump adoption shape (`integer` + explicit `nextval` default + `OWNED BY`) → counts as serial, which is the trade-off the new TODO entry documents.

The double `regclass` round-trip with `quote_literal` is subtle but correct, and the comment explains why it is needed. Nice that the lateral also drops the duplicated `pg_get_serial_sequence` call from the two `CASE` arms.

**Partition-child comments (diff/tables.go, parser)** — `withInheritedColumns` is the right shape: bodyless entries only exist inside the comment diff, so they cannot leak into DDL emission, and `Table.SQL` ignores columns on the `PARTITION OF` branch anyway. Restricting the parser's synthetic column entry to true partition children (`PartitionOf != nil && PartitionBound != nil`) correctly avoids the INHERITS false-new-column trap. The pass-through of a comment on an unknown column (fail at apply, not silently vanish) is a good choice and is tested.

**`ALTER SEQUENCE ... OWNED BY` (parser)** — consistent with the `CREATE SEQUENCE ... OWNED BY` path (same `parseSeqOwnedBy` helper), and since unknown statements were already silently ignored by the parse loop, adding this case changes no error behavior. Fixtures cover unqualified names, unknown sequences, and non-`OWNED BY` options.

### Notes, non-blocking

- One residual gap the restriction leaves: `COMMENT ON COLUMN` for an inherited column of an **INHERITS** child is still silently dropped (the column lookup fails and the partition-child branch is not taken). That is pre-existing behavior and the code comment explains why it cannot use the same trick, but it may deserve a TODO.md entry alongside the other known gaps this PR documents.
- `applyAlterSeqOwnedBy` applies `OWNED BY NONE` by nilling the owner, so a desired file with `CREATE SEQUENCE s OWNED BY t.c; ALTER SEQUENCE s OWNED BY NONE;` plans the unusable `CREATE` — exactly the transition the new TODO entry scopes out, so just noting the parse-level path exists.
- The three new TODO entries (dump dependency ordering, sub-partition `PARTITION BY` loss, pg_dump-shape drift) are honest findings from the same audit; documenting them instead of half-fixing them here keeps the PR reviewable.

CHANGELOG and code comments are clear, and the fixture-first test style follows the repo conventions. LGTM.

---

dump.go
winebarrel: You are right, the comment was wrong. `--omit-schema` limits the dumped schema to one, but the parent can sit outside it, so "always lives in the schema being stripped" does not hold.

Confirmed the behavior against a live database with the parent in `other` and the child in the dumped `public`:

```
CREATE TABLE child PARTITION OF other.parent FOR VALUES FROM ('2024-01-01') TO ('2025-01-01');
```

`TrimPrefix` is a no-op there and the parent keeps its schema, which is what the reference needs to resolve. So the code was already correct and only the comment overstated it. Reworded to say the schema is stripped only when the parent is in the one being dumped, and why a foreign parent has to keep it.

Added `TestDump_OmitSchema_PartitionParentInAnotherSchema` rather than a fixture, since the dump harness sets up `public` alone and a second schema would leak into later cases. It uses the existing `setupSchemaDB` helper, which cleans up after itself.

Fixed in d45f309.

---

parser/parser.go
winebarrel: Agreed, the message named the internal variable instead of what a user writes in the schema file. Changed to:

```
pista:renamed-from is ambiguous: ALTER TABLE public.orders adds 2 constraints in one statement
```

Updated `error_multiple_constraints_ambiguous_rename.yml` and the parser test that asserts on the text along with it.

Fixed in d45f309.

---

winebarrel (commented):
## Re-review (563985a, d45f309)

Both review comments are addressed, and I re-ran the full suite on `d45f309` against a local PostgreSQL 16 (`go test -p 1 ./...`) — all green.

**dump.go comment / cross-schema parent** — The reworded comment now states the actual invariant (strip only when the parent is in the dumped schema), and `TestDump_OmitSchema_PartitionParentInAnotherSchema` pins the pass-through behavior. Going with a Go test instead of a fixture is the right call here and matches the repo convention: multi-schema setups needing `setupSchemaDB` belong in Go, per CLAUDE.md.

**Error wording** — The message now says `pista:renamed-from is ambiguous`, with the fixture and the parser test updated consistently.

**Coverage commit (563985a)** — The package-level tests are faithful to the behavior already covered by fixtures and add a few checks the fixtures could not express: `withInheritedColumns` leaving the input tables untouched, the catalog's `PartitionOf` equaling the parent's `FQTN()`, and the INHERITS-child comment being intentionally *not* synthesized (`TestParseCommentStmt_InheritsChildColumnNotSynthesized`), which documents the gap in executable form. The new TODO entry for that gap correctly ties its resolution to materializing inherited columns, same prerequisite as the existing INHERITS entry, and the `OWNED BY NONE` entry now accurately says the parser reads the clause while nothing emits detaching DDL.

Nothing further from me — LGTM.

---

PR #343, opened 2026-07-31
Add -- pista:execute-first directive

winebarrel:
Non-managed SQL marked with -- pista:execute runs after the managed DDL.
That order fails when the managed DDL depends on the object the statement
creates: a CHECK constraint, a GENERATED expression, an index expression,
or a policy that calls a function needs that function to already exist.

-- pista:execute-first runs the statement before the managed DDL instead.
-- pista:execute is unchanged and still runs last, so both directions are
available in the same file. The two names share a prefix, but the execute
pattern only accepts whitespace or end-of-line after the name, so neither
claims the other's comment.

Check SQL is evaluated at the point the statement runs, so an
execute-first check sees the pre-change schema while an execute check
sees the post-change one.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Uxojc2JYxG6zkvMYFZdpCV


apply_test.go
```diff
@@ -638,6 +638,109 @@ CREATE OR REPLACE FUNCTION public.test_func() RETURNS void AS $$ BEGIN END; $$ L
...
+		ConnString: conn.Config().ConnString(),
+		Schemas:    []string{"public"},
+	})
+
+	var buf bytes.Buffer
+	_, err := client.Apply(ctx, &pistachio.ApplyOptions{Files: []string{desiredFile}}, &buf)
+	require.NoError(t, err)
+
+	out := buf.String()
+	assert.Less(t, strings.Index(out, "before_func"), strings.Index(out, "CREATE TABLE public.users"))
```
winebarrel, line 745: `strings.Index` returns -1 when the substring is absent, so this assertion passes for the exact regression it is meant to catch: if the execute-first statement were never emitted, `Index(out, "before_func")` is -1, which is still `Less` than the `CREATE TABLE` index.

Line 741 is safe by accident (a missing `GRANT` gives -1 on the right side and fails), but line 740 is not. Assert presence first:

```go
require.Contains(t, out, "before_func")
require.Contains(t, out, "CREATE TABLE public.users")
require.Contains(t, out, "GRANT")
```

---

README.md
```diff
@@ -513,6 +513,22 @@ END $do$;
...
+
+```sql
+-- pista:execute-first SELECT to_regprocedure('public.lower_v(text)') IS NULL
+CREATE OR REPLACE FUNCTION public.lower_v(t text) RETURNS text AS $$ SELECT lower(t) $$ LANGUAGE sql IMMUTABLE;
+
+CREATE TABLE public.users (
+    id integer NOT NULL,
+    v text,
+    CONSTRAINT users_pkey PRIMARY KEY (id),
+    CONSTRAINT users_v_check CHECK (lower_v(v) <> 'x'::text)
```
winebarrel: The `::text` cast is not needed. I applied this example and re-planned with `CHECK (lower_v(v) <> 'x')`, and the plan is empty -- `normalizeCheckExpr` already strips the text-like cast, so the shorter form does not drift.

Dropping it keeps the example minimal and avoids implying the cast is required to stay drift-free. Same at `directives.md:84` and in `testdata/apply/execute_first_function_used_by_check.yml`.

---

parser/directive.go
```diff
@@ -88,7 +99,15 @@ func extractExecuteDirectives(rawSQL string, stmts []*pg_query.RawStmt) ([]*Exec
 		leadingEnd := findLeadingCommentEnd(region)
 		leading := region[:leadingEnd]
 
-		matches := executeDirectivePattern.FindAllStringSubmatch(leading, -1)
+		// A statement carrying both directives takes the one written closest
+		// to it, matching the "last match wins" rule used within each kind.
+		first := false
+		matches := executeDirectivePattern.FindAllStringSubmatchIndex(leading, -1)
+		firstMatches := executeFirstDirectivePattern.FindAllStringSubmatchIndex(leading, -1)
+		if len(firstMatches) > 0 && (len(matches) == 0 || firstMatches[len(firstMatches)-1][0] > matches[len(matches)-1][0]) {
```
winebarrel: The prefix disambiguation itself is correct: `executeDirectivePattern` requires whitespace or end-of-line after `pista:execute`, so an `execute-first` comment never matches it, and the switch to `FindAllStringSubmatchIndex` with the `lastMatch[2] >= 0` guard preserves the "no check SQL" vs "empty check SQL" distinction.

The part worth reconsidering is the tie-break. Writing both `-- pista:execute` and `-- pista:execute-first` on one statement is a contradiction, not a preference, and silently taking the closest one hides the mistake -- the statement lands on the opposite side of the managed DDL from what half the file says. The repo's precedent for ambiguous directive usage is to error (`-- pista:concurrently does not accept arguments`, and #340 made `renamed-from` on a multi-constraint `ALTER TABLE` an error). Suggest erroring here too, e.g. `-- pista:execute and -- pista:execute-first cannot both apply to one statement`. Repeating the same directive can keep last-match-wins.

---

testdata/apply/execute_first_function_used_by_check.yml
```diff
@@ -0,0 +1,22 @@
...
+      CONSTRAINT users_v_check CHECK (lower_v(v) <> 'x'::text)
+  );
+applied: |
+  -- public.users
+  CREATE TABLE public.users (
+      id integer NOT NULL,
+      v text,
+      CONSTRAINT users_pkey PRIMARY KEY (id),
+      CONSTRAINT users_v_check CHECK (lower_v(v) <> 'x'::text)
+  );
```
winebarrel, line 22: Add `verify_no_drift: true` here. This is the one fixture where the CHECK expression calls a function, which is exactly the shape the new TODO entry says drifts when the call is schema-qualified. Pinning the unqualified form as drift-free turns that entry's "workaround: write the call unqualified" into something the suite enforces, so a later change to `normalizeCheckExpr` cannot quietly break it.

---

winebarrel (commented):
Reviewed at 8b5ef30. Ran `go vet ./...` and the full `TestPlan`/`TestApply` suites plus `./parser/...` against local PostgreSQL 16 -- all green. `golangci-lint` could not run here (the installed binary is built with go1.25, below the module's targeted 1.26), so lint is unverified.

The core design holds up. Plan and apply agree on ordering (execute-first -> managed DDL -> execute), the prefix ordering with pre-SQL is unchanged, and building a fresh `stmts` slice in `plan.go` also removes the old aliasing of `result.Stmts` under `append`. The regex prefix disambiguation is sound.

I reproduced the new TODO entry's claim to confirm it: with `CHECK (public.lower_v(v) <> 'x')` the plan re-emits `DROP CONSTRAINT` + `ADD CONSTRAINT` on every run after a successful apply. The entry describes it accurately.

Inline comments cover a test assertion that cannot fail, the unnecessary `::text` in the examples, the both-directives tie-break, and a fixture that should assert no drift. Beyond those:

**Docs gap.** `getting-started.md:248` still says execute statements "are executed after schema changes during `apply`", and `README.md` points there with "See Getting Started for details" right below the new section. A reader following that link gets the pre-PR picture. It needs the execute-first sentence too.

**Test placement.** Per the repo guidelines, plan/apply tests that are purely SQL-input -> SQL-output belong in YAML. `TestApply_ExecuteFirstRunsBeforeManagedDDL` and `TestApply_ExecuteFirstCheckSQLFalseSkips` both fit: `applied_sql` captures the emitted ordering and the skipped statement, and `applied` captures the resulting schema. `TestApply_ExecuteFirstCheckSQLError` correctly stays in Go, since the apply harness has no `error` field.

**No scenario test.** `test/scenario/execute.test.sh` exists with three steps for `execute`. A step-through of execute-first -- create the function, then add the CHECK that calls it, then verify drift-free -- would exercise the ordering at the CLI level, which is where the failure this directive prevents actually shows up.

**Pre-existing wrinkle the new docs lean on.** `plan` never evaluates check SQL, so an execute-first with `to_regprocedure(...) IS NULL` keeps appearing in plan output and `plan --check` stays at exit 2 forever, even though apply skips it. I confirmed this on the README example. It is not introduced here -- plain `execute` behaves the same -- but the fixture comment ("skipped afterwards") is only true of apply, and the README teaches this pattern as the idempotent one. A sentence noting that plan shows the statement unconditionally would set expectations.

**Minor.** The `ApplyResult.Applied` doc comment (`apply.go:36-38`) still enumerates only `-- pista:execute`; execute-first now sets it too. The `extractConstraintName` / `extractColumnName` unterminated-quote assertions in `parser/directive_test.go` are unrelated coverage top-ups -- they pass, just worth noting they are not part of this feature.

---

testdata/apply/execute_first_function_used_by_check.yml
winebarrel: `verify_no_drift` cannot hold on a fixture that contains an execute directive. `plan` renders every execute statement unconditionally and never evaluates its check SQL (`plan.go`), so `plan.SQL` always carries the `-- pista:execute-first` line and the assertion fails on it, not on the CHECK:

```
Error: Should be empty, but was -- pista:execute-first SELECT to_regprocedure('public.lower_v(text)') IS NULL
       CREATE OR REPLACE FUNCTION public.lower_v(t text) RETURNS text AS $$ SELECT lower(t) $$ LANGUAGE sql IMMUTABLE;
```

The intent is worth pinning, so I did it without the directive instead: `testdata/plan/no_diff_check_calling_function.yml` carries the function in `init` and asserts `plan: ""`. That covers the same guarantee, and covers it more precisely, since nothing else in the fixture can mask a regression in `normalizeCheckExpr`.

Dropping the `::text` cast made that fixture sharper than expected. The desired side now writes `'x'` while `pg_get_constraintdef` returns `'x'::text`, so the fixture pins both halves at once: the unqualified call and the stripped cast have to cancel out for the plan to stay empty.

---

winebarrel: Superseded: `verify_no_drift: true` is now set on this fixture as you asked.

My earlier reply was right about the old behavior but the behavior itself was the problem. `plan` printed every execute statement without evaluating its check, so `plan.SQL` was never empty while a directive was present. Chasing that turned up a worse case: with `-- pista:execute SELECT false`, `plan` printed the SQL while `apply` on the same file reported `-- No changes` and ran nothing.

`plan` now evaluates the check and leaves out what `apply` would skip. On the second run here the function exists, the check is false, the plan is empty, and the assertion holds on its own terms rather than by exception.

`testdata/plan/no_diff_check_calling_function.yml` stays. It pins the same guarantee without a directive in the file, so a regression in `normalizeCheckExpr` cannot hide behind the check evaluation.

The change is marked BREAKING in the changelog. The consequence worth flagging is that `plan` now runs the check on a read-only connection, so a check that writes fails with SQLSTATE 25006 where it used to run only under `apply`.

---

apply_test.go
winebarrel: Fixed in `5f3675e`. Added the three `require.Contains` calls before the position comparisons, with a comment recording why, so the assertion cannot pass on a pair of -1s.

---

README.md
winebarrel: Fixed in `5f3675e`, dropped in all three places.

Worth noting what it exposed in the fixture. The `applied` block is a dump, so it keeps `'x'::text`, which is what `pg_get_constraintdef` renders. The fixture now pins the asymmetry directly: desired writes `'x'`, the database returns `'x'::text`, and they have to compare equal.

`testdata/plan/no_diff_check_calling_function.yml` leans on the same thing, so both halves of the comparison (the stripped cast and the unqualified call) are pinned as drift-free.

---

parser/directive.go
winebarrel: Fixed in `5f3675e`. Carrying both directives is now an error: `-- pista:execute and -- pista:execute-first cannot both apply to one statement`. Repeating one keeps last-match-wins, and `directives.md` states both rules.

The tie-break test became an error test covering each order, so neither can silently resolve.

---

plan.go
```diff
@@ -97,11 +97,35 @@ func (client *Client) Plan(ctx context.Context, options *PlanOptions) (*PlanResu
...
+	// apply would run rather than everything the file holds.
+	appendExecuteStmts := func(stmts []string, first bool) ([]string, error) {
+		for _, es := range result.ExecuteStmts {
+			if es.First != first {
+				continue
+			}
+			shouldExecute := true
+			if es.CheckSQL != "" {
+				if err := conn.QueryRow(ctx, es.CheckSQL).Scan(&shouldExecute); err != nil {
+					return nil, fmt.Errorf("failed to evaluate check SQL: %s: %w", es.CheckSQL, err)
```
winebarrel: Aborting the whole plan when the check SQL errors is the part that needs to change. A check that cannot be evaluated at plan time is not a broken schema file -- it is a question plan cannot answer -- and today it takes the entire diff down with it, including the managed DDL that has nothing to do with the check.

Reproduced on this head:

```sql
-- pista:execute SELECT NOT EXISTS (SELECT 1 FROM public.audit_log)
INSERT INTO public.audit_log (id, note) VALUES (1, 'seed');
CREATE TABLE public.audit_log (
    id integer NOT NULL,
    note text,
    CONSTRAINT audit_log_pkey PRIMARY KEY (id)
);
```

```
$ pista plan ...
pista: error: failed to evaluate check SQL: SELECT NOT EXISTS (SELECT 1 FROM public.audit_log): ERROR: relation "public.audit_log" does not exist (SQLSTATE 42P01)
exit 1

$ pista apply ...   # same file
CREATE TABLE public.audit_log (...);
-- pista:execute SELECT NOT EXISTS (SELECT 1 FROM public.audit_log)
INSERT INTO public.audit_log (id, note) VALUES (1, 'seed');
exit 0
```

`plan` cannot even show the `CREATE TABLE`. On main it printed both statements and `apply` behaved identically.

Suggested fix: treat a failed check as undetermined rather than fatal. Keep the statement in the plan (the pre-PR output) and annotate why, e.g.

```
-- pista:execute SELECT NOT EXISTS (SELECT 1 FROM public.audit_log)
-- check SQL could not be evaluated at plan time: relation "public.audit_log" does not exist; apply will decide
INSERT INTO public.audit_log (id, note) VALUES (1, 'seed');
```

That keeps the change's motivation intact -- a check that evaluates cleanly to false still drops out, so `plan --check` stops reporting a permanent diff -- while a check plan cannot run degrades to the old behavior instead of breaking the command. It also covers the read-only case (SQLSTATE 25006) with the same path, which is worth doing since that failure is an artifact of plan's connection mode rather than anything wrong with the file.

The narrower alternative, if you would rather not add the annotation: evaluate only `execute-first` checks in plan. Those run against the pre-change schema in both commands, so plan and apply agree by construction. Plain `execute` checks are evaluated post-DDL at apply time, which plan fundamentally cannot reproduce.

---

directives.md
```diff
@@ -68,6 +71,28 @@ CREATE OR REPLACE FUNCTION public.my_func() RETURNS void AS $$ ... $$ LANGUAGE p
...
+    id integer NOT NULL,
+    v text,
+    CONSTRAINT users_pkey PRIMARY KEY (id),
+    CONSTRAINT users_v_check CHECK (lower_v(v) <> 'x')
+);
+```
+
+The check SQL is evaluated where the statement runs, so an `execute-first` check sees the schema before the change and an `execute` check sees it after. Put a check that tests for a table or column the same run creates on `execute`. An `execute-first` check answers the same in `plan` and `apply`, since both evaluate it against the pre-change schema.
+
+Statements keep their file order within each group. There is no dependency resolution between them.
```
winebarrel, line 102: This paragraph now contradicts the one added at line 65. Here the reader is told to put a check that tests for a table or column the same run creates on `execute`; there they are told such a check "can answer differently in the two commands". In practice it does not answer differently -- `plan` errors out and exits 1 (repro in my comment on `plan.go`). Following this advice makes `plan` unusable on the file.

If the `plan.go` behavior stays as-is, this sentence has to become a warning rather than a recommendation, and "can answer differently" needs to say "fails". If the check failure degrades to a warning instead, both sentences work as written.

---

winebarrel (commented):
Re-reviewed 8b5ef30..72ae699. `go vet`, the full `go test -p 1 ./...` suite, and `make test-scenario` all pass here against PostgreSQL 16.

Four of the five round-one points are resolved, and two are resolved better than I asked. The `::text` removal came with `testdata/plan/no_diff_check_calling_function.yml`, which pins the cancellation of the two normalizations rather than just dropping the cast. The both-directives case is now an error with a symmetric two-way test, and `directives.md` states the rule. `verify_no_drift: true` is on the CHECK fixture, and the `require.Contains` guards are in with a comment saying why.

The new plan-evaluates-check-SQL commits are where I would hold this. The `execute-first` work itself looks ready.

## plan now fails on the pattern the docs recommend

Details and a full repro are in the `plan.go` comment. In short: a plain `execute` check that references a table the same run creates errors at plan time, and the error aborts the entire plan -- `plan` exits 1 without printing any DDL, while `apply` on the identical file exits 0. `directives.md` recommends exactly this placement two paragraphs below where it warns about it, so the page argues with itself.

The motivation is sound and the `--check` fix is worth having. What needs to change is the failure mode: a check plan cannot evaluate should degrade to the pre-PR behavior (print the statement, annotate why) rather than kill the command. The same path handles the read-only rejection, which is an artifact of plan's connection mode rather than a defect in the file.

Two smaller things follow from the same commit:

- **Scope.** The PR title and body are still only about `execute-first`. A **BREAKING** change to `plan` semantics for the pre-existing `execute` directive affects every current user of that directive and is not mentioned anywhere outside the CHANGELOG. Either update the body, or split it into its own PR -- the `execute-first` half is ready on its own and would merge cleanly without waiting on this.
- **`--no-read-only` has teeth now.** The CHANGELOG offers it as the workaround for a check that writes, but with it, `plan` executes that write. A `nextval` check consumes a sequence value on every plan, and the check then runs a second time at apply. Worth saying outright in the CHANGELOG and `directives.md` that the escape hatch makes `plan` non-read-only, instead of presenting it as a neutral flag.

## Still open from round one

- **`getting-started.md:248`** was not touched, and the new commits make it worse rather than merely incomplete. It says execute statements "are executed after schema changes during `apply`" (no `execute-first`), and line 272 -- "Execute statements appear in `plan` output. During `apply`, the check SQL is evaluated and the statement is skipped if it returns `false`." -- is now wrong on both halves. `README.md` links here from directly under the new section, so it is where a reader ends up. GitHub would not let me comment inline since the file is not in the diff.
- **`TestApply_ExecuteFirstRunsBeforeManagedDDL` and `TestApply_ExecuteFirstCheckSQLFalseSkips`** are still Go tests that `applied_sql` fixtures would cover. Fine to leave if deliberate, but the guidelines point the other way.
- **No scenario test.** `test/scenario/execute.test.sh` has three steps for `execute` and nothing for `execute-first`. It is also the natural place to catch the plan/apply divergence above, since it runs both commands over the same files.

## Minor

- `apply.go:36-38`: the `Applied` doc comment still enumerates only `-- pista:execute`; `execute-first` sets it too.
- `plan_test.go:124`: "Plan never writes" no longer holds -- plan runs user-supplied check SQL, and under `NoReadOnly` that SQL can write.
- The conflict error at `parser/directive.go:107` carries no file or statement context. With several files on the command line, "cannot both apply to one statement" leaves the reader hunting. Worth including the deparsed statement or the file name if that plumbing is available.

---

plan.go
winebarrel: Fixed in `48b8a6d`, taking the annotation option. Reproduced your case first, on that head:

```
$ pista plan ...
pista: error: failed to evaluate check SQL: SELECT NOT EXISTS (SELECT 1 FROM public.audit_log): ERROR: relation "public.audit_log" does not exist (SQLSTATE 42P01)
exit 1
```

and after:

```
CREATE TABLE public.audit_log (
    id integer NOT NULL,
    note text,
    CONSTRAINT audit_log_pkey PRIMARY KEY (id)
);
-- pista:execute SELECT NOT EXISTS (SELECT 1 FROM public.audit_log)
-- check SQL could not be evaluated at plan time: ERROR: relation "public.audit_log" does not exist (SQLSTATE 42P01); apply will decide
INSERT INTO public.audit_log (id, note) VALUES (1, 'seed');
```

Undetermined counts as pending, so `plan --check` still reports a diff. The note is folded to one line, and a test asserts it stays a comment so the plan remains valid SQL.

I did not take the narrower alternative. Restricting evaluation to `execute-first` would have left the original complaint in place for plain `execute`, which is where the `to_regprocedure(...) IS NULL` pattern in the README lives, and that check evaluates fine at plan time.

Two follow-ons. The read-only case now degrades through the same path, so it is no longer a breaking change and the changelog no longer claims one. And `apply` is untouched: the check runs there at its proper moment on a read-write connection, so a failure is real and still stops the run.

The three error fixtures became Go tests, since the output now carries the PostgreSQL error text and an exact match would be fragile across server versions.

---

directives.md
winebarrel: Resolved by the `plan.go` change in `48b8a6d`: the check failure now degrades instead of aborting, so both sentences hold as written.

I still tightened them, since "can answer differently" was vague about what the reader would actually see. This paragraph now says `plan` cannot answer such a check and shows the statement as undetermined, while `apply` decides correctly. The `execute` section gained the note format and an example, and states that a failure during `apply` is still an error.

---

plan.go
```diff
@@ -97,13 +97,42 @@ func (client *Client) Plan(ctx context.Context, options *PlanOptions) (*PlanResu
...
+	appendExecuteStmts := func(stmts []string, first bool) []string {
+		for _, es := range result.ExecuteStmts {
+			if es.First != first {
+				continue
+			}
+			shouldExecute := true
+			note := ""
+			if es.CheckSQL != "" {
+				if err := conn.QueryRow(ctx, es.CheckSQL).Scan(&shouldExecute); err != nil {
+					shouldExecute = true
```
winebarrel, line 120: This resolves it. Verified on this head -- the round-two repro now plans cleanly:

```
CREATE TABLE public.audit_log (...);
-- pista:execute SELECT NOT EXISTS (SELECT 1 FROM public.audit_log)
-- check SQL could not be evaluated at plan time: ERROR: relation "public.audit_log" does not exist (SQLSTATE 42P01); apply will decide
INSERT INTO public.audit_log (id, note) VALUES (1, 'seed');
exit 0
```

Also checked the things that could have gone wrong with a non-fatal error path:

- Several unevaluable checks in one file all degrade independently -- pgx sets read-only as a startup parameter rather than an explicit transaction, so a failed `QueryRow` does not poison the connection for the next one. Confirmed with four execute statements, two of which fail.
- Ordering survives: an undetermined `execute-first` still lands above the managed DDL, an undetermined `execute` below it.
- The note stays one comment line. I piped a plan containing one straight into `psql -v ON_ERROR_STOP=1` and it ran clean, so the "plan output is executable SQL" property holds.

One consequence worth a deliberate decision rather than a change request. `Scan(&shouldExecute)` now folds a genuinely malformed check into the same bucket as an unanswerable one:

```
-- pista:execute SELECT 42
-- check SQL could not be evaluated at plan time: can't scan into dest[0]: cannot scan int4 (OID 23) in binary format into *bool; apply will decide
```

A check with the wrong return type or a typo used to be caught at plan time and now reaches `apply`. The note text is clear and `--check` still reports a diff, so it is visible rather than silent, and I would rather have this than the aborting version. If you want the stricter split later, the natural line is by SQLSTATE: `42P01` / `42703` / `25006` are "plan cannot answer this", and anything else is the file being wrong. Not worth the complexity today unless it bites someone.

---

winebarrel (commented):
Re-reviewed 72ae699..48b8a6d. `go vet`, `go test -p 1 ./...`, and `make test-scenario` all pass against PostgreSQL 16.

The blocking issue is fixed, and fixed well. Details and my verification are in the `plan.go` comment: the round-two repro plans cleanly, several unevaluable checks in one file degrade independently without poisoning the connection, ordering across the managed DDL is preserved, and the plan output is still executable SQL when piped to `psql`. The docs no longer contradict themselves -- `directives.md` shows the undetermined output inline, so the reader sees the shape rather than a description of it, and the CHANGELOG dropped the `--no-read-only` suggestion in favour of stating that a side-effecting check now runs once per `plan`. That is the better call: pointing users at a flag that makes `plan` write was the weaker half of the previous wording.

`TestPlan_ExecuteCheckUnevaluable` asserting the note stays on a `-- ` line is the right thing to pin, and the comment explaining why it is a Go test rather than a fixture (server-version-dependent error text) is the correct read of the guidelines.

This looks good to me. I would have approved if GitHub let me approve your own PR. Everything below is optional and none of it should hold the merge.

## Carried over, still open after three rounds

**`getting-started.md`** is the one I would actually still do. Line 248 says execute statements "are executed after schema changes during `apply`", and line 272 says "Execute statements appear in `plan` output. During `apply`, the check SQL is evaluated and the statement is skipped if it returns `false`." After this PR neither sentence is true: `execute-first` runs before, and `plan` evaluates the check as well. `README.md` links here from directly under the new section, so a reader following "See Getting Started for details" gets the pre-PR model of both directives. Three or four sentences.

**Scope / PR body.** The body still describes only `execute-first`. The **BREAKING** `plan` change stands on its own merits now, but someone reading the merge commit will not find it. Worth a paragraph in the body even if you would rather not split the PR at this point.

## Minor, take or leave

- No scenario test for `execute-first`. `test/scenario/execute.test.sh` has three steps for `execute`, and it is the one place that runs `plan` and `apply` over the same file -- which is exactly the divergence the last two rounds were about.
- `TestApply_ExecuteFirstRunsBeforeManagedDDL` and `TestApply_ExecuteFirstCheckSQLFalseSkips` would fit `applied_sql` fixtures.
- `apply.go:35-38`: the `Applied` doc comment still enumerates only `-- pista:execute`.
- `plan_test.go:124`: "Plan never writes" is no longer accurate -- plan runs user check SQL, and under `NoReadOnly` that SQL can write. Ironically that test's own premise is now the interesting case.
- `parser/directive.go:108`: the both-directives error still carries no file or statement context.

---

plan.go
winebarrel: Keeping it as is, as a deliberate decision.

A malformed check is visible rather than silent: the note carries the scan error verbatim, and `--check` still reports a diff, so it cannot pass unnoticed.

The SQLSTATE split is the right line if this ever bites, but it does not cover this case cleanly. A wrong return type fails in `Scan`, on the pgx side, with no SQLSTATE to branch on, so the split would need two mechanisms, one for server errors and one for scan errors. Not worth it against the current behavior.

Also ran `make test-scenario` on this head: 21 scenarios, 140 steps, all passing, including the existing `execute` scenario.

---

PR #363, opened 2026-08-04
Warn on unsupported statements ignored by the parser

winebarrel:
## Summary

The parser dropped any top-level statement type it did not handle without any output, so a statement like `SET`, `GRANT`, or `CREATE EXTENSION` was silently excluded from the desired schema. This adds a `default` case to the statement switch that writes a warning to stderr:

```
pistachio: ignored unsupported statement: <sql>
```

The snippet is normalized to one line and truncated at 200 characters so the warning stays readable even for large statement bodies. Empty statements (stray semicolons) produce no warning.

## Tests

- Normal path plus the warning message.
- Unsupported statement last with no trailing semicolon, where pg_query reports `StmtLen=0`; also checks the supported statement is still parsed and a multi-line body is normalized to one line.
- Truncation of a statement longer than the limit.
- No warning for a supported statement.

`warnIgnoredStmt` coverage is 90.9%; the only uncovered branch is a defensive guard for an empty snippet, which pg_query never produces because it drops empty statements. Parser package coverage is 86.9%.

## Docs

- README Supported Objects notes that other statements are dropped with a warning.
- CHANGELOG entry under 1.23.1.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,14 +23,19 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - name: Check for non-ASCII characters
         run: |
+          # git grep exit: 0 = found, 1 = none, >1 = real error.
           matches=$(LC_ALL=C git grep -nIP '[^\x00-\x7F]' -- .) && status=0 || status=$?
-          if [ "$status" -eq 0 ]; then
+          if [ "$status" -gt 1 ]; then
+            echo "Error: git grep failed with status $status"
+            exit "$status"
+          fi
+          # A line tagged "ascii-ignore" opts out, for the rare test that must
+          # hold a multibyte character.
+          matches=$(printf '%s\n' "$matches" | grep -ve '^$' -e 'ascii-ignore' || true)
+          if [ -n "$matches" ]; then
             echo "Error: non-ASCII characters found:"
             echo "$matches"
             exit 1
-          elif [ "$status" -ne 1 ]; then
-            echo "Error: git grep failed with status $status"
-            exit "$status"
           fi
 
   windows:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,19 +23,14 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
       - name: Check for non-ASCII characters
         run: |
-          # git grep exit: 0 = found, 1 = none, >1 = real error.
           matches=$(LC_ALL=C git grep -nIP '[^\x00-\x7F]' -- .) && status=0 || status=$?
-          if [ "$status" -gt 1 ]; then
-            echo "Error: git grep failed with status $status"
-            exit "$status"
-          fi
-          # A line tagged "ascii-ignore" opts out, for the rare test that must
-          # hold a multibyte character.
-          matches=$(printf '%s\n' "$matches" | grep -ve '^$' -e 'ascii-ignore' || true)
-          if [ -n "$matches" ]; then
+          if [ "$status" -eq 0 ]; then
             echo "Error: non-ASCII characters found:"
             echo "$matches"
             exit 1
+          elif [ "$status" -ne 1 ]; then
+            echo "Error: git grep failed with status $status"
+            exit "$status"
           fi
 
   windows:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [1.23.1] - 2026-08-04
 
-* Warn when the parser ignores an unsupported statement. A schema file may hold a statement pistachio does not manage, such as `SET`, `GRANT`, or `CREATE EXTENSION`. Before, it was dropped from the desired schema without any notice. pistachio now prints `pistachio: ignored unsupported statement: <sql>` to standard error for each one. The statement is shown as a single canonical line, without the surrounding comments, and truncated when long. Mark a statement with `-- pista:execute` to keep it and silence the warning.
+* Warn when the parser ignores an unsupported statement. A schema file may hold a statement pistachio does not manage, such as `SET`, `GRANT`, or `CREATE EXTENSION`. Before, it was dropped from the desired schema without any notice. pistachio now prints `pistachio: ignored unsupported statement: <sql>` to standard error for each one. The statement is shown as a single canonical line, without the surrounding comments, and truncated when long. Mark a statement with `-- pista:execute` to keep it and silence the warning. A `BEGIN`/`COMMIT` warning points at `--with-tx` and `--try-tx`.
 
 ## [1.23.0] - 2026-08-01
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.23.1] - 2026-08-04
+
+* Warn when the parser ignores an unsupported statement. A schema file may hold a statement pistachio does not manage, such as `SET`, `GRANT`, or `CREATE EXTENSION`. Before, it was dropped from the desired schema without any notice. pistachio now prints `pistachio: ignored unsupported statement: <sql>` to standard error for each one. The snippet is normalized to one line and truncated when long.
+
 ## [1.23.0] - 2026-08-01
 
 * **BREAKING**: `plan` now evaluates `-- pista:execute` check SQL and leaves out the statements `apply` would skip. Before, every marked statement was printed regardless, so a file with a false check produced a plan full of SQL while `apply` on the same file reported `-- No changes`. Three consequences:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [1.23.1] - 2026-08-04
 
-* Warn when the parser ignores an unsupported statement. A schema file may hold a statement pistachio does not manage, such as `SET`, `GRANT`, or `CREATE EXTENSION`. Before, it was dropped from the desired schema without any notice. pistachio now prints `pistachio: ignored unsupported statement: <sql>` to standard error for each one. The snippet is normalized to one line and truncated when long.
+* Warn when the parser ignores an unsupported statement. A schema file may hold a statement pistachio does not manage, such as `SET`, `GRANT`, or `CREATE EXTENSION`. Before, it was dropped from the desired schema without any notice. pistachio now prints `pistachio: ignored unsupported statement: <sql>` to standard error for each one. The statement is shown as a single canonical line, without the surrounding comments, and truncated when long. Mark a statement with `-- pista:execute` to keep it and silence the warning.
 
 ## [1.23.0] - 2026-08-01
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ pista apply ./schema/*.sql         # apply it
 - Array, JSON, UUID, and other built-in types
 - Quoted identifiers
 
+pistachio parses only the statements above. It drops any other statement in a schema file, such as `SET`, `GRANT`, or `CREATE EXTENSION`, and prints a `pistachio: ignored unsupported statement:` warning to standard error for each one.
+
 ## Usage
 
 ```
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ pista apply ./schema/*.sql         # apply it
 - Array, JSON, UUID, and other built-in types
 - Quoted identifiers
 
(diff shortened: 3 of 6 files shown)


winebarrel (commented):
Reviewed at 90b6ec9. The direction is right: silently dropping `GRANT` or `CREATE EXTENSION` from the desired schema is exactly the kind of thing a user should hear about. `go vet` and `go test ./parser/` pass. Two things I verified and liked: a statement marked `-- pista:execute` is skipped before the switch, so it produces no warning, and the `snippet == ""` guard really is unreachable (`;`, `SET a=1;;`, and `SET a=1;\n;\n` all behave as the PR describes).

Two issues in `warnIgnoredStmt` worth fixing, plus some smaller notes.

## 1. The snippet swallows the comments around the statement

`rawStmt.StmtLocation` points at the byte just after the previous semicolon, so every comment and blank line between two statements becomes part of the snippet:

```
CREATE TABLE t (id int);
-- a leading comment
SET foo = 1;
```

```
pistachio: ignored unsupported statement: -- a leading comment SET foo = 1
```

This is not an edge case: in a real schema file most statements have a comment above them, so most warnings will carry one. A `-- pista:ignore` line above an unsupported statement ends up in the warning too, which reads as if the directive were part of the ignored SQL. The `StmtLen == 0` path has the mirror problem, appending whatever trails the last statement:

```
CREATE TABLE t (id int);
GRANT SELECT ON t TO PUBLIC
-- trailing comment
```

```
pistachio: ignored unsupported statement: GRANT SELECT ON t TO PUBLIC -- trailing comment
```

Trimming leading comment lines and whitespace would cover the common case. Deparsing `rawStmt` and falling back to the raw slice when deparse fails would give a clean statement on both ends.

## 2. Truncation cuts bytes, so it can emit invalid UTF-8

`snippet[:maxLen]` splits a multibyte rune whenever the 200th byte lands mid-character:

```go
parser.ParseSQLWithSchema("SELECT 'x"+strings.Repeat("あ", 200)+"';", "public")
// stderr ends with: ...\xe3\x81...
// utf8.ValidString(out) == false
```

The repo is ASCII-only, but the SQL being parsed is user input and string literals or quoted identifiers can hold anything. Truncating on a rune boundary fixes it. Related: with multibyte text the 200-byte cap cuts at roughly 66 characters, so a rune-based limit would also make the truncation point predictable.

## Smaller notes

- `BEGIN;` and `COMMIT;` each produce a warning. Wrapping a schema file in an explicit transaction is common and nothing is lost by ignoring those, so warning on them is noise rather than a notice. Skipping `TransactionStmt` seems worth considering.
- The README paragraph and the CHANGELOG entry describe the warning but not what to do about it. `-- pista:execute` is the way to keep such a statement in the file, and it silences the warning; pointing at it from the new paragraph would make the warning actionable.
- The warning has no file or line. `ParseSQLFilesWithSchema` joins every file before parsing, so it cannot report one today. With a directory of schema files, a bare `GRANT SELECT ON t TO PUBLIC` can be hard to locate. Not necessarily this PR's job, but it limits how useful the message is.
- Version: this adds new output rather than fixing a bug, so 1.24.0 fits the way 1.23.0 and 1.21.1 are used in this changelog better than 1.23.1 does.

---

winebarrel:
Thanks for the review. Pushed 2e3835a.

**1. Comments swallowed.** Fixed. `warnIgnoredStmt` now renders the statement by deparsing it (`ignoredStmtSnippet`), so the snippet is a single canonical line with the surrounding comments and blank lines dropped, on both the normal and the `StmtLen == 0` path. The raw slice is kept only as a fallback for a statement pg_query can parse but not deparse; that branch is defensive and stays uncovered because deparse of an already-parsed statement does not fail in practice.

**2. Rune-boundary truncation.** Fixed. Truncation now cuts on `[]rune`, so the message stays valid UTF-8 for multibyte literals, and the 200-rune cap is predictable. Added a test with a multibyte body that asserts `utf8.ValidString`.

**BEGIN/COMMIT.** Left as warnings per your call, so they go through `default` like any other unsupported statement. Added a test that asserts both warn while the table between them is still parsed.

**Docs.** The README paragraph and the CHANGELOG entry now point at `-- pista:execute` as the way to keep such a statement and silence the warning.

**Version.** Kept at 1.23.1.

**File/line.** Left out of this PR, as you noted; `ParseSQLFilesWithSchema` joins the files before parsing, so it cannot report a location today.

winebarrel (commented):
Re-reviewed at 5dfc5c3. Both issues from the last round are fixed, and I verified each one:

- Deparsing the raw statement drops the surrounding comments cleanly. `CREATE TABLE t (id integer);\n-- a leading comment\nSET foo = 1;` now warns with `SET foo TO 1` and nothing else, and the trailing-comment case is gone too.
- Truncation on a rune boundary produces valid UTF-8; the input that used to end in a split rune now passes `utf8.ValidString`.
- The `-- pista:execute` escape hatch is documented in both the README and the CHANGELOG, and the transaction hint is a better answer than the skip I suggested. `--with-tx` and `--try-tx` both exist and match the wording.

`go test ./parser/` passes at 87.0% coverage. I could not run `make lint` here (the installed golangci-lint is built with go1.25 against a go1.26 module) or the DB-backed suites, so those are unverified from my side.

One new problem came in with the fix, plus a few notes.

## The deparsed snippet is no longer one line

`ignoredStmtSnippet` returns the deparse output as is, so the `strings.Fields` normalization now applies only to the fallback path. Deparse preserves newlines inside a statement body, and the warning spans as many lines as the body does:

```sql
CREATE FUNCTION f() RETURNS int AS $$
BEGIN
  RETURN 1;
END;
$$ LANGUAGE plpgsql;
```

```
pistachio: ignored unsupported statement: CREATE FUNCTION f() RETURNS int AS $$
BEGIN
  RETURN 1;
END;
$$ LANGUAGE plpgsql
```

A `DO $$ ... $$` block gives three lines, and even `SELECT 'a\nb';` gives two. This is not exotic input: a function or a trigger is one of the most likely things to sit unsupported in a schema file, so it is close to the common case. It also contradicts the CHANGELOG and the doc comment, which both promise a single canonical line, and it defeats the rune cap, since a 200-rune body still spans several lines.

Collapsing in `warnIgnoredStmt` rather than in `ignoredStmtSnippet` would cover both paths and put the truncation on the collapsed string:

```go
snippet := strings.Join(strings.Fields(ignoredStmtSnippet(sql, rawStmt)), " ")
```

## The CI opt-out looks avoidable

The workflow change exists to let one test literal hold a multibyte character. Writing it as an escape keeps the source pure ASCII and produces the same three-byte rune at runtime:

```go
body := strings.Repeat("あ", 250)
```

That drops the `ascii-ignore` mechanism and the workflow edit entirely, and the test still asserts what it asserts today. If you would rather keep the opt-out, CLAUDE.md is the place to record it, since it currently states the ASCII rule with no exception and an agent reading only that file will not know the marker exists. The `git grep` status handling in the rewritten step is correct either way.

## Notes

- Deparse rewrites the statement, so the warning no longer quotes the file: `SET foo = 1` prints as `SET foo TO 1`, and `GRANT SELECT ON t TO PUBLIC` prints as `GRANT select ON t TO public` with the privilege lowercased. It stays recognizable and I think the trade is right, but the lowercase keyword reads like a typo, and it may be why `TestParseSQL_WarnsUnsupportedStmt_LastNoSemicolon` had to weaken its assertion to just `GRANT`.
- The hint fires for every `TransactionStmt`, so `ROLLBACK;` and `SAVEPOINT sp;` also get told to use `--with-tx` or `--try-tx`, which does not really answer either one. Restricting the hint to `BEGIN`/`COMMIT` would keep it accurate.
- `--with-tx` and `--try-tx` live on `ApplyOptions`; `PlanOptions` has neither. The warning is emitted during `plan` as well, where the suggested flags are rejected. The wording says "run the apply in a transaction", so it is defensible, but a user who reads it under `plan` may try it there first.
- `TestIgnoredStmtSnippet_DeparseFallback` drives the fallback with a `RawStmt` whose `Stmt` is nil, which pg_query never produces. It is a fine unit test of the helper, but it does not exercise a real parse-but-not-deparse statement, so the fallback stays untested against one. That is reasonable given CLAUDE.md's stance on unreachable defensive code, worth stating rather than acting on.

---

winebarrel:
Correction to the CI opt-out section of my review: the snippet I pasted is what the test already has, which makes the suggestion read as a no-op. What I meant was the Unicode escape, which is ASCII in the source and still yields the same three-byte rune at runtime:

```go
body := strings.Repeat("あ", 250)
```

That is the form that lets the `ascii-ignore` marker and the workflow edit go away. The rest of the point stands unchanged.

---

winebarrel:
My correction above pasted the same literal a second time, because the escape sequence I meant to show does not survive being written out here. Here is the suggestion in a form that cannot be mangled:

```go
body := strings.Repeat(string(rune(0x3042)), 250)
```

That source line is pure ASCII and builds the same three-byte character at runtime (750 bytes, 250 runes, `utf8.ValidString` true), so the truncation test keeps testing exactly what it tests today while `parser_test.go` stops tripping the non-ASCII check. A `"あ"` string literal works the same way if you prefer that spelling. Either one removes the need for the `ascii-ignore` marker and the workflow change.

---

winebarrel:
Thanks, pushed c4e4f66.

**Multi-line snippet.** Fixed. The `strings.Fields` collapse moved into `warnIgnoredStmt`, so both the deparse and fallback paths yield one line, and truncation runs on the collapsed string. `ignoredStmtSnippet` now returns the raw deparse output or raw slice, and its doc comment says so. Added a `CREATE FUNCTION` test that asserts the warning is a single line.

**CI opt-out.** Reverted, both the workflow edit and the marker. The truncation test now uses `strings.Repeat("あ", 250)`, so the source stays ASCII and the escape produces the same 3-byte rune at runtime. The test asserts the same thing as before.

**Transaction hint scope.** Restricted to `TRANS_STMT_BEGIN`, `TRANS_STMT_START`, and `TRANS_STMT_COMMIT` (`END` maps to COMMIT). `ROLLBACK` and `SAVEPOINT` now warn without the hint. I kept `START` alongside `BEGIN` since it opens a transaction the flags cover; a test pins that `ROLLBACK`/`SAVEPOINT` get no hint.

On the notes I did not act on: the lowercased keyword from deparse (`GRANT select`) is why `_LastNoSemicolon` asserts just `GRANT`; the plan-vs-apply wording stays as "run the apply in a transaction"; and the fallback test drives a nil `Stmt` rather than a real parse-but-not-deparse statement, as you noted.

Coverage is 87.0%; `ignoredStmtSnippet` and `txHint` are fully covered, and the only uncovered line in `warnIgnoredStmt` is the empty-snippet guard.

winebarrel (commented):
Re-reviewed at e91a8fa. Everything I raised is resolved, and I verified each one against the branch.

**Multi-line snippets.** Collapsing in `warnIgnoredStmt` covers both paths, and moving it out of `ignoredStmtSnippet` keeps the helper's contract honest. The cases that used to span lines are now one line each:

```
pistachio: ignored unsupported statement: CREATE FUNCTION f() RETURNS int AS $$ BEGIN RETURN 1; END; $$ LANGUAGE plpgsql
pistachio: ignored unsupported statement: DO $$ BEGIN PERFORM 1; END $$
pistachio: ignored unsupported statement: SELECT 'a b'
```

**Transaction hint.** The `TransactionStmtKind` switch draws the line where I would. Confirmed by hand across the whole statement family: `BEGIN`, `START TRANSACTION`, `COMMIT`, and `END` (which deparses to `COMMIT`) get the hint; `ROLLBACK`, `ABORT`, `SAVEPOINT`, `RELEASE`, `ROLLBACK TO SAVEPOINT`, `PREPARE TRANSACTION`, `COMMIT PREPARED`, and `SET TRANSACTION` get none. `BEGIN ISOLATION LEVEL SERIALIZABLE` is still hinted, which is right.

**CI opt-out.** `あ` does the job and `ci.yml` is byte-identical to main again, so the ASCII rule keeps its single meaning and CLAUDE.md needs no exception. Running the workflow's `git grep` over the tree reports no matches.

`go vet ./...` is clean and `go test ./parser/` passes at 87.0% coverage. As before I could not run `make lint` (the installed golangci-lint is built with go1.25 against a go1.26 module) or the DB-backed suites, so those remain unverified from my side.

Two consequences of the deparse-and-collapse approach worth being aware of, neither of which I would change: the warning prints canonical SQL rather than the file's text, so `END;` is reported as `COMMIT` and `SET foo = 1` as `SET foo TO 1`, and the collapse reaches inside string literals and function bodies, so `SELECT 'a\nb'` prints as `SELECT 'a b'`. For a warning whose job is to name the statement that got dropped, both are the right trade.

Nothing blocking from me.

---

PR #364, opened 2026-08-05
Add dump --sort-by-deps to order the dump by dependency

winebarrel:
## Summary

Add `dump --sort-by-deps` to order the dump by object dependency instead of by name (opt-in; the default output is unchanged).

Each object is written after the objects it depends on:
- types before the tables that use them,
- referenced tables before the tables whose foreign keys point at them,
- tables before the views that select from them.

The output then loads from top to bottom with `psql` without forward references. The order is computed with the existing `toposort` package (the same one `plan`/`apply` use). On a dependency cycle the sort fails and dump falls back to the name order. The flag has no effect with `--split`, which writes each object to its own file.

## Changes

- `dump.go`: `DumpOptions.SortByDeps` flag, `DumpResult.SortByDeps`, and a dependency-ordered rendering path in `String()`. Output separators are unchanged, so only the order differs. `tables()`/`views()` are now nil-safe like the other helpers.
- Tests:
  - YAML fixtures `sort_by_deps.yml` (enum + FK + view chain), `sort_by_deps_view_chain.yml` (view depending on a view), `sort_by_deps_cycle.yml` (mutual FK -> name-order fallback).
  - Go tests for order toggle, `--omit-schema` interaction, `--split` unaffected, and nil-map handling.
  - Scenario test `test/scenario/sort_by_deps.test.sh`: dump with `--sort-by-deps`, restore with `psql`, verify no forward references and no drift; and that the default name-ordered dump fails to restore.
- `Makefile`: `test` now depends on `clean-schema` so the suite always runs against a clean database.
- `README.md` and `CHANGELOG.md` (1.24.0) updated.

## Verification

- `make lint` (0 issues), `go vet`, `gofmt` clean.
- Full suite green; new dump code is fully covered.
- End to end: a `--sort-by-deps` dump reloads via `psql` with no error and no drift, while the default order fails on a forward reference.


winebarrel:
## Review

The feature works. I checked out the branch, ran it against a live PostgreSQL 16, and confirmed the dependency ordering is correct beyond what the tests cover: a standalone sequence referenced by a `nextval` default is emitted before the table, a domain before the table that uses it as a column type, a partition parent before its partition, and a materialized view before the view that selects from it. A `--sort-by-deps` dump of that mixed schema restored with `psql -v ON_ERROR_STOP=1` cleanly. I also diffed the two dumps line by line after sorting: the output is byte for byte identical to the default dump apart from the order, so the separator handling is right.

Two things need attention before merge.

### 1. The two new view fixtures fail on PostgreSQL 16 and later (blocking)

`testdata/dump/sort_by_deps.yml` and `testdata/dump/sort_by_deps_view_chain.yml` only define the base `dump:` field, and its content is the PostgreSQL 15 form of `pg_get_viewdef`. From 16 on, the deparse drops the relation qualification, so `SELECT posts.id` becomes `SELECT id`. Both fixtures fail on my PG 16.13:

```
--- FAIL: TestDump (2.83s)
    --- FAIL: TestDump/sort_by_deps (0.03s)
    --- FAIL: TestDump/sort_by_deps_view_chain (0.03s)
```

Everything else in `go test -p 1 ./...` passes, so the failure is specific to these two fixtures. The CI matrix is 15/16/17/18, so this turns three of the four jobs red.

The fix follows the existing convention in `testdata/dump/simple_view.yml`: add a `dump_pg16` block with the unqualified form and alias it to `dump_pg17` / `dump_pg18` with a YAML anchor. For `sort_by_deps.yml` the trailing view becomes

```yaml
  -- public.active_posts
  CREATE OR REPLACE VIEW public.active_posts AS
  SELECT id
     FROM posts;
```

and for `sort_by_deps_view_chain.yml`, `SELECT id FROM t` and `SELECT id FROM z_base`.

### 2. `test: clean-schema` in the Makefile

This is unrelated to the feature and the PR body does not say why it is here. Two problems with it:

- `clean-schema` runs `psql` under the Makefile's exported `PGHOST=localhost` / `PGUSER=postgres`, while the Go tests connect through `TEST_PISTA_CONN_STR`. With a custom connection string, `make test` now wipes a different database than the one it tests against.
- The Windows job runs bare `make`, whose default goal `all` includes `test`. That job invokes `& "$env:PGBIN\pg_isready"` with an explicit path rather than a bare `pg_isready`, which suggests the PostgreSQL bin directory is not on `PATH` there. If so, the bare `psql` in `clean-schema` will not resolve and the job fails.

`testutil.SetupDB` already drops and recreates `public` for each test, so the only leftovers this catches are objects in non-public schemas from `make schema`. That is a reasonable thing to want, but it deserves its own PR and a connection string that matches what the tests use.

### Suggestions

**Warn on the cycle fallback.** When the sort fails, the dump silently comes out in name order. The user asked for dependency order, gets something that looks fine, and only finds out at restore time via a forward-reference error that does not point back at the cause. 1.23.1 established the pattern for this: a `pistachio: ...` line on stderr, along the lines of `pistachio: dependency sort failed, falling back to name order`. `String()` cannot return an error, but `dependencyOrderedSQL` can write to a warn writer the same way `parser.go:81` does.

**`appendDumpItems` zip-by-index.** Correctness rests on the helper methods preserving both order and count relative to the original map, an invariant currently held together by a comment. It does hold today (relation and type names share a namespace in PostgreSQL, so the keys are globally unique, and `--omit-schema` is restricted to a single schema, so stripping cannot collide). But the failure modes are unkind: a key missing from `pos` silently gets position 0 and sorts to the front, and a count mismatch panics on `adjusted[i]` rather than degrading. Using the comma-ok form on `pos[k]` and abandoning the sorted path when a key is missing would make it fail into the existing name-order fallback instead.

**Test coverage.** The paths I exercised by hand above are not pinned by any test: sequence via `nextval` default, domain or composite type as a column type, partition parent, materialized view. `OrderFromSchema` already handles all of them, so one fixture covering a sequence and a domain would be cheap insurance for the dump side.

**Scenario step 04.** Asserting that the default dump *fails* to restore is a useful control, but it pins in a negative. If the default order is ever improved, that step goes red without anything actually being broken. Worth a comment noting that, at least.

---

winebarrel:
Confirmed on 6968623: the fixture fix is correct. `dump_pg16` carries the unqualified deparse form and is aliased to `dump_pg17` / `dump_pg18`, matching the convention in `simple_view.yml`. `go test -p 1 -run TestDump .` is green on my PG 16.13, where it failed before.

Point 2 is still open: `test: clean-schema` is unchanged in the Makefile, so the mismatch with `TEST_PISTA_CONN_STR` and the bare `psql` on the Windows job still stand. The rest of my comment was suggestions rather than blockers.

---

winebarrel:
## Re-review (01467ed)

Every point from the previous round is addressed, and two of them are handled better than what I suggested. Verified on PostgreSQL 16.13.

**Fixture variants.** `dump_pg16` aliased to `dump_pg17` / `dump_pg18`, per the `simple_view.yml` convention. `go test -p 1 ./...` is fully green on PG16.

**Makefile.** `test: clean-schema` reverted.

**Cycle handling.** Erroring in `Client.Dump` is better than the stderr warning I proposed. `--sort-by-deps` is a request for a specific property, and a dump that silently lacks it is worse than no dump. Confirmed at the CLI:

```
$ pista dump --sort-by-deps
pista: error: failed to order dump by dependency: schema dependency sort failed: cycle detected: sorted 0 of 2 nodes
```

Keeping the `ok=false` degradation in `String()` for hand-built `DumpResult`s is the right split, since `String()` has no error channel. The comment saying so is worth having.

**`--split` xor.** Also better than the "no effect" documented before, since the flag combination never made sense.

```
$ pista dump --split /tmp/outdir --sort-by-deps
pista: error: --split and --sort-by-deps can't be used together
```

**`appendDumpItems`.** The comma-ok lookup plus the length check closes both failure modes: a key missing from `pos` no longer silently sorts to the front, and a count mismatch no longer panics. Both now route to the name-order fallback.

**Coverage.** `sort_by_deps_types.yml` pins the sequence-via-`nextval` and domain-as-column-type paths, and step 04 of the scenario now documents that it pins a negative.

### Additional checks

Built a synthetic 300-table / 150-view schema with a 300-deep FK chain and views joining across it. `--sort-by-deps` restored it with `psql -v ON_ERROR_STOP=1` in one pass and `pista plan` reported no changes afterwards. I also timed the duplicate `OrderFromSchema` (once in `Dump` for validation, once in `String`) on that schema: 358-404 ms sorted against 369-481 ms unsorted, so the second pass is lost in the noise even with 150 view definitions going through `pg_query.Parse` twice. Not worth optimizing.

### Remaining nits, neither blocking

The `appenders` slice of six closures in `dependencyOrderedSQL` costs more to read than it saves. Six sequential calls express the same thing more directly:

```go
var items []dumpItem
var ok bool
if items, ok = appendDumpItems(items, r.Enums, r.enums().CollectValues(), pos, model.EnumToSQL); !ok {
    return "", false
}
// ...
```

Same line count, no indirection through a function-typed slice.

`Dump` computes the order purely to validate it and discards it, then `String` recomputes it. Storing the order on `DumpResult` would remove the duplication, but as measured above there is no performance argument for it, so it is only worth doing if it reads better to you.

One thing to watch in CI rather than a finding: `sort_by_deps_types.yml` is the only dump fixture containing a `nextval` default, so it is the first to depend on how `pg_get_expr` renders a standalone sequence reference across the 15/16/17/18 matrix. It passes on 16 here; the other three jobs will confirm.

LGTM once CI is green.

---

winebarrel:
## Re-review (971e843)

The last nit is resolved: the six closures are gone and `dependencyOrderedSQL` now reads as six sequential `appendDumpItems` calls with the `!ok` guard inline. `gofmt`, `go vet`, and `go test -p 1 ./...` are clean on my PG 16.13.

Since only a mechanical refactor landed, I spent this round on the cases I had not exercised yet. All of them behave:

**Cross-schema dependencies.** `-n public -n zsch` with `public.achild` holding an FK to `zsch.base` and a column of type `zsch.col`, plus `public.aview` selecting from `zsch.base`. The order came out `zsch.base`, `public.aview`, `zsch.col`, `public.achild`, and the dump restored into two empty schemas in one `psql -v ON_ERROR_STOP=1` pass. Dependencies are tracked across schema boundaries, not just within one.

**Filters that remove a dependency.** `-E base` drops the FK parent and the enum's table out from under their dependents; `-E col` drops the enum. Neither errors nor mis-sorts. This is the case I was most curious about, since a filtered-out object leaves an edge pointing at nothing, but `OrderFromSchema` only adds edges to names in `defined`, so the graph stays consistent with whatever survived the filter.

Nothing further from me. The remaining `Dump`/`String` double sort is the only thing I would still call optional, and I measured it as free last round, so it is fine to leave.

CI on 971e843 is still running as I write this — the 15/17/18 matrix jobs are the last open question, specifically whether `sort_by_deps_types.yml` renders its `nextval` default identically on those versions. 16 is confirmed locally.

LGTM.

---

PR #369, opened 2026-08-06
Error on a name reused across PostgreSQL catalogs

winebarrel:
## What

Reject a schema file that claims one name twice within a PostgreSQL catalog: for two objects in a schema, or for two members of a single object. `validateNamespaces` (`parser/validate.go`) runs at the end of parsing, next to `validateColumnRefs`.

## Motivation

Each object kind lives in its own map, so `setUnique` only caught a repeat within a kind:

```sql
CREATE TABLE x (id integer);
CREATE VIEW x AS SELECT 1 AS n;
```

`plan` printed both statements and `apply` failed partway through:

```
pista: error: failed to execute SQL: CREATE OR REPLACE VIEW public.x AS
SELECT 1 AS n;: ERROR: "x" is not a view (SQLSTATE 42809)
```

Worse, with the table already in the database, `x` is in the desired tables but not in the desired views, so the run also offered an unrelated `-- skipped: DROP TABLE public.x;`.

The same shape appears wherever a slice or a per-kind map stands in for what a catalog keys as one namespace. All of the following parsed and failed at apply:

```sql
CREATE TABLE t (id integer, CONSTRAINT c CHECK (id > 0), CONSTRAINT c FOREIGN KEY (id) REFERENCES p (id));
CREATE DOMAIN d AS integer CONSTRAINT c CHECK (VALUE > 0) CONSTRAINT c CHECK (VALUE < 10);
CREATE TYPE ct AS (n integer, n integer);
CREATE TYPE et AS ENUM ('a', 'a');
```

## Namespaces

Membership was checked against PostgreSQL rather than assumed, including the catalog indexes: `pg_constraint_conrelid_contypid_conname_index` is unique on `(conrelid, contypid, conname)`, `pg_attribute_relid_attnam_index` on `(attrelid, attname)`, `pg_enum_typid_label_index` on `(enumtypid, enumlabel)`.

| Kind | pg_class | pg_type | scoped to one object |
|---|---|---|---|
| table | yes | yes | - |
| view / materialized view | yes | yes | - |
| composite type | yes | yes | - |
| sequence | yes | no | - |
| index | yes | no | - |
| domain / enum | no | yes | - |
| `PRIMARY KEY` / `UNIQUE` / `EXCLUDE` constraint | yes, under the constraint name | no | pg_constraint, per table |
| `CHECK` / foreign-key constraint | no | no | pg_constraint, per table |
| domain `CHECK` constraint | no | no | pg_constraint, per domain |
| composite type attribute | no | no | pg_attribute, per type |
| enum label | no | no | pg_enum, per type |

pg_class and pg_type are schema-wide; the rest are scoped to one object, so each table, domain and type gets a namespace of its own. An index written without a name is skipped, since PostgreSQL picks that name at create time. Objects marked `-- pista:ignore` take part, since an unmanaged relation still occupies its name.

Errors are aggregated with `errors.Join`, one line per object:

```
duplicate relation name: public.x (table and view)
duplicate type name: public.x (table and domain)
duplicate constraint name: c on public.t (CHECK constraint and FOREIGN KEY constraint)
duplicate attribute: n on public.ct
duplicate enum value: a on public.et
```

Two objects of one kind can also reach the check, since index names are unique per schema while pistachio keys them per table. That case reports `duplicate index: public.x`, the wording `setUnique` already uses for the same clash within one table.

## Behavior changes

- A duplicate index name on two tables in the same schema is now an error. PostgreSQL rejects it; pistachio kept index names unique per table only.
- Unnamed constraints are named after every key column, matching PostgreSQL's `ChooseConstraintName`: `UNIQUE (a, b)` is `t_a_b_key`, not `t_a_key`, and a `CHECK` takes its name from the one column its expression references (`t_check` for none or several) even when written on a column. Adopting a table that already exists no longer plans an `ADD CONSTRAINT` under pistachio's name next to a `DROP CONSTRAINT` for PostgreSQL's on every run. A database built with the old names gets the constraint added under the new name on the next run; dropping the old one needs `--allow-drop constraint`, and until then the table carries both and every plan repeats the skipped drop. Writing the old name as an explicit `CONSTRAINT` clause keeps it instead. An unnamed `ALTER TABLE ... ADD FOREIGN KEY` is named too; it used to reach the plan as `ADD CONSTRAINT  FOREIGN KEY ...`.
- Two unnamed `CHECK` constraints on one domain are now an error. PostgreSQL names them `d_check` and `d_check1`, a suffix pistachio does not predict, so both come out as `d_check`. Unnamed table constraints already behaved this way, and README documents the limitation under "Constraint naming".

## Expression walker

`WalkExprColumnRefs` (`internal/pgast`) skipped five node kinds, so a column referenced only through one of them was invisible to all three of its callers:

| CHECK | PostgreSQL | before |
|---|---|---|
| `a[1] > 0` | `t_a_check` | `t_check` |
| `greatest(a, 1) > 0` | `t_a_check` | `t_check` |
| `ROW(a) IS NOT NULL` | `t_a_check` | `t_check` |
| `a IS TRUE` | `t_a_check` | `t_check` |
| `(a COLLATE "C") > 'x'` | `t_a_check` | `t_check` |
| `f(x => a, y => 1) > 0` | `t_a_check` | `t_check` |
| `xmlelement(name e, a) IS NOT NULL` | `t_a_check` | `t_check` |
| `xmlserialize(content a AS text) > ''` | `t_a_check` | `t_check` |

`A_Indirection`, `MinMaxExpr`, `RowExpr`, `BooleanTest`, `CollateClause`, `NamedArgExpr`, `XmlExpr` and `XmlSerialize` now get walked, which fixes the generated name above, lets `validateColumnRefs` see a column that does not exist behind one of them, and lets the `-- pista:renamed-from` rewriting reach the old name. On `main`, renaming a column referenced through `GREATEST` planned a needless constraint drop and add:

```
ALTER TABLE public.t RENAME COLUMN old_col TO new_col;
ALTER TABLE public.t DROP CONSTRAINT t_chk;
ALTER TABLE public.t ADD CONSTRAINT t_chk CHECK (GREATEST(new_col, 1) > 0);
```

Now it is the `RENAME COLUMN` alone.

The JSON constructors PostgreSQL 16 added are the same shape and are recorded in `TODO.md` rather than fixed here: `JSON_OBJECT` alone needs `JsonObjectConstructor` -> `JsonKeyValue` -> `JsonValueExpr`, five more kinds carry the same gap, pg_query_go has 26 in the family, and the syntax does not parse on PostgreSQL 15. Walking every child node generically is not an option: the visitor mutates what it visits for renames, so it would rewrite a function name, an `A_Indirection` field or a `TypeName` string that happens to match a column.

## Test

- `parser/validate_test.go`: 33 collision cases covering every row in the table above, in both directions where the registration order shows in the message; 12 accepted cases for the kinds and scopes that do not share a namespace (sequence / domain, index / domain, `CHECK`, foreign key, different schemas, names differing only in case, unnamed indexes, one constraint name on two tables or two domains, one attribute name in two types, one label in two enums); plus one-line-per-object and multi-name aggregation.
- `testdata/plan/error_duplicate_relation_name.yml`, `error_duplicate_type_name.yml`, `error_duplicate_constraint_name.yml`: the errors through `plan`.
- `parser/parser_test.go`: 21 cases pinning the generated names against the ones PostgreSQL picks for the same SQL, covering multi-column `UNIQUE` / `EXCLUDE` / foreign keys, the `CHECK` expression rule including the five node kinds above, an `EXCLUDE` on an expression, and the three `ALTER TABLE ADD` forms.
- `internal/pgast/pgast_test.go`: one case per newly walked node kind, following the existing one-per-kind tests, including `xmlattributes` and `xmlforest`, whose elements arrive wrapped in a `ResTarget`. A further case pins the exclusions: an element name, an attribute alias and an argument name are written where a column name could be, and each case has a column of that name on the same table, so picking one up would name the constraint after the wrong column.
- Coverage goes from 87.0% to 90.3% of statements for `parser` and from 88.5% to 91.5% for `internal/pgast`. Everything added is at or near 100%; the new cases also take `parseCompositeTypeStmt` from 0% to 85.7% and `parseSQLWithSchema` from 80.3% to 86.6%. The remaining uncovered spots are nil guards and an unreachable `default` that no SQL input can take.
- `make test`, `make vet`, `make lint`, `make test-scenario` pass.
- `make test-samples`: 33 real-world sample schemas still round-trip through `dump` -> `plan` (`33 passed, 0 failed`), so nothing in them trips the check.


parser/validate.go
```diff
@@ -196,3 +196,105 @@ func walkExprColumnRefs(node *pg_query.Node) []string {
...
+			if err := ns.add(name, kind); err != nil && !reported[name] {
+				reported[name] = true
+				errs = append(errs, err)
+			}
+		}
+	}
+
+	addIndexes := func(indexes *orderedmap.Map[string, *model.Index]) {
+		for _, idx := range indexes.CollectValues() {
+			add(model.Ident(idx.Schema, idx.Name), "index", relations)
```
winebarrel, line 291: An index written without a name reaches here with `idx.Name == ""`. `model.Ident` skips empty parts, so the key collapses to the bare schema name and every unnamed index in a schema registers `public`:

```sql
CREATE TABLE t1 (id integer);
CREATE TABLE t2 (id integer);
CREATE INDEX ON t1 (id);
CREATE INDEX ON t2 (id);
```

```
duplicate index name: public
```

PostgreSQL accepts this and names the indexes `t1_id_idx` and `t2_id_idx`, so the check rejects a schema the server takes, and the message points at the schema rather than an index. Same shape for any name-less relation that reaches `add`.

Guarding `add` on an empty name looks like the smallest fix:

```go
add := func(name, kind string, nss ...*namespace) {
	if name == "" || !strings.Contains(name, ".") {
		return
	}
	...
}
```

or skip the index in `addIndexes` when `idx.Name == ""`, since pistachio cannot know the name PostgreSQL will generate. (`CREATE INDEX ON t (id)` twice on one table already fails in `setUnique` with `duplicate index: `, so unnamed indexes are shaky either way, but this PR extends that across tables.)

---

winebarrel (commented):
Reviewed the change. The direction is right: per-kind maps could never catch a name reused across kinds, and moving the check to the end of parsing, next to `validateColumnRefs`, puts it where the whole `ParseResult` is available.

I checked the catalog table in the description against PostgreSQL 16 rather than taking it as read. `SELECT relname, relkind, reltype FROM pg_class` gives `reltype = 0` for both `relkind = 'S'` and `relkind = 'i'`, so sequences and indexes really do have no `pg_type` row, and the accepted cases `sequence` / `domain` and `index` / `domain` are correct. Splitting the two catalogs instead of using one flat name set is what makes those pass.

The rest of the reading also holds up. `errors.Join` with the `reported` map gives one line per name, and iteration over `orderedmap` keeps the line order stable. Including `-- pista:ignore` objects is right, since an unmanaged relation still holds its name, and `directives.md` says so. Constraint names are filled in by `generateConstraintName` before the check runs, so an inline `PRIMARY KEY` or `UNIQUE` is registered under the name PostgreSQL would pick.

Two things.

**Unnamed indexes produce a false error.** Details inline on `parser/validate.go`. `CREATE INDEX ON t (id)` on two tables in one schema now fails with `duplicate index name: public`, which PostgreSQL accepts.

**The same per-kind-map gap remains for constraint names.** `pg_constraint` requires a constraint name to be unique per relation across all constraint kinds, but `ForeignKeys` and `Constraints` are separate maps, so each `setUnique` only sees its own kind:

```sql
CREATE TABLE p (id integer, CONSTRAINT p_pkey PRIMARY KEY (id));
CREATE TABLE t (id integer, CONSTRAINT c CHECK (id > 0), CONSTRAINT c FOREIGN KEY (id) REFERENCES p (id));
```

The parser accepts it; PostgreSQL answers `ERROR: constraint "c" for relation "t" already exists`. That is the failure mode this PR is about, one catalog over, and it fails partway through an apply the same way. It needs a per-table namespace rather than the two schema-wide ones, so a follow-up is fine, but the CHECK and foreign-key cases in `TestValidateNamespaces_NoCollision` read as if constraint names were fully handled. A note in the comment on `validateNamespaces` saying `pg_constraint` is out of scope would keep that clear.

A nit: `setUnique` reports `duplicate index: i` and the new path reports `duplicate index name: public.x`. Which message a user sees depends only on whether the two indexes sit on the same table, which is not a distinction they care about. Worth settling on one phrasing.

The test set is thorough otherwise; every pair in the table is covered in both directions where the order matters, and the two `plan` fixtures show the error through the CLI path.

---

parser/validate.go
```diff
@@ -196,3 +196,134 @@ func walkExprColumnRefs(node *pg_query.Node) []string {
...
+
+	for name := range r.Sequences.Keys() {
+		add(name, "sequence", relations)
+	}
+
+	for name := range r.CompositeTypes.Keys() {
+		add(name, "composite type", relations, types)
+	}
+
+	for name := range r.Domains.Keys() {
```
winebarrel: pg_constraint's unique index is on `(conrelid, contypid, conname)`, and a domain constraint takes the second column. So the catalog this PR now covers scopes names per domain as well as per table, and a domain reaching this loop registers only its own name:

```sql
CREATE DOMAIN d AS integer CONSTRAINT c CHECK (VALUE > 0) CONSTRAINT c CHECK (VALUE < 10);
```

```
ERROR:  constraint "c" for domain "d" already exists
```

The parser takes it. `Domain.Constraints` is a slice rather than an `orderedmap`, so unlike `Constraints` and `ForeignKeys` this one has no `setUnique` behind it either; the repeat is not caught within a kind, let alone across kinds. Another `newRelationScopedNamespace("constraint", fqdn)` per domain covers it, and `add`'s existing `on <scope>` wording reads the same for a domain as for a table.

Not something this PR introduced, so a follow-up is fine. Worth deciding now, though, since the comment above says pg_constraint keys a constraint name to its relation, which is half the story.

---

winebarrel (commented):
Both points are addressed, and I re-ran the cases rather than reading the diff alone. Looks good to merge from my side.

The unnamed-index skip is right. `CREATE INDEX ON t1 (id)` next to `CREATE INDEX ON t2 (id)` now parses clean, and skipping at the point where the name is read keeps the empty string out of `model.Ident` entirely, so nothing registers a bare schema name. The comment gives the reason, which is the part worth keeping: PostgreSQL picks the name at create time, so pistachio has nothing to compare.

The pg_constraint namespace matches the catalog. A constraint name repeated on two tables is accepted, and `CONSTRAINT c CHECK (...)` next to `CONSTRAINT c FOREIGN KEY ...` on one table is rejected, which is what PostgreSQL does in both directions. The `PRIMARY KEY` name on two tables still fails through pg_class, since that one carries an index; on PostgreSQL 16 the same SQL answers `relation "x" already exists`, so the split between the two namespaces lands where the server draws it. Names are generated before the check runs for the unnamed `CHECK` and foreign-key forms too, so those go in as `t_check` and `t_id_fkey` rather than as empty strings.

Dropping the `reported` map in favor of returning after the first rejection is a good trade. Reporting per object instead of per name means a name that clashes in both catalogs now gives a line for each object involved, rather than hiding the second behind the first:

```sql
CREATE TABLE x (id integer);
CREATE VIEW x AS SELECT 1 AS n;
CREATE DOMAIN x AS integer;
```

```
duplicate relation name: public.x (table and view)
duplicate type name: public.x (table and domain)
```

A table and a composite type still report once, since the object is not registered anywhere after its first rejection. The wording change lines the same-kind case up with `setUnique`, so an index name reused across tables and one reused within a table now read alike.

One thing left, inline on `parser/validate.go`: domain constraints are the other half of pg_constraint and go unchecked. Fine as a follow-up.

Two more of the same shape, further out and clearly separate work, both from a slice standing in for a keyed map: `CREATE TYPE ct AS (n integer, n integer)` and `CREATE TYPE et AS ENUM ('a', 'a')` both parse, and PostgreSQL answers `column "n" specified more than once` and a `pg_enum_typid_label_index` violation.

`go build`, `go vet` and the `parser` package tests pass here.

---

winebarrel (commented):
The domain constraints are covered now, and the two cases I had listed as further out came along with them. Looks good to merge from my side.

Generalizing `scope` from a relation to any owning object is the right shape: pg_constraint keys on `(conrelid, contypid, conname)`, so a table and a domain want the same construct with a different owner, and pg_attribute and pg_enum are the same idea one level down. `newScopedNamespace` reads better than the relation-specific name it replaces. The note that a single-kind namespace never prints its catalog label is worth having, since `"attribute"` and `"enum value"` are otherwise dead-looking arguments.

Checked the scoping against PostgreSQL 16 in both directions. A constraint name on two domains, a constraint name shared by a domain and a table, an attribute name in two composite types, and a label in two enums are all accepted, matching the new cases. `"N"` next to `n` in one composite type and `'A'` next to `'a'` in one enum are accepted too: attribute names are folded identifiers and labels are literals, and comparing the raw strings gets both right without a special case.

The unnamed-domain-CHECK behavior change is a good trade, and better than the description claims. On `main`, that schema does not merely fail at apply, `plan` prints DDL that cannot work:

```
$ pista plan d.sql
CREATE DOMAIN public.d AS integer
    CONSTRAINT d_check CHECK (value > 0)
    CONSTRAINT d_check CHECK (value < 10);
```

So the parse error replaces a plan that was wrong on its face. Documenting it in the CHANGELOG, in the test comment, and against the README limitation is the right way to land it.

Two small things, neither blocking.

The Test section still closes with "so the check rejects nothing PostgreSQL accepts". That was true two commits ago; the domain `CHECK` entry under Behavior changes is now exactly such a case. What the samples run shows is that nothing in 33 real schemas trips it, which is the more useful claim anyway.

The README paragraph this leans on has the suffix form wrong, and it is not in the diff so it will not get fixed by accident:

```sql
CREATE TABLE tt (id integer, CHECK (id > 0), CHECK (id < 10));
SELECT conname FROM pg_constraint WHERE conrelid = 'tt'::regclass;
```

```
 tt_id_check
 tt_id_check1
```

No separator before the number, same as `d_check1`. README line 822 says "a numeric suffix (e.g. `_1`)". The CHANGELOG and the PR description both spell it correctly, so only the README is off. One more variant lands on the same path, if you want it in the test table: an explicit `CONSTRAINT d_check` next to an unnamed `CHECK` on one domain, where PostgreSQL also picks `d_check1`.

Ran the full suite against PostgreSQL 16 here, not just the parser package: `go vet ./...` clean and `go test -p 1 ./...` green across all packages.

---

winebarrel (commented):
Both nits are closed. Nothing left blocking from my side.

The README line now matches what the server does, and adding the consequence is the better half of the change: a reader hitting `duplicate constraint: users_id_check` can trace it back from the limitation. The Test section reads "so nothing in them trips the check", which is what the samples run actually shows.

The new case is the right expectation. PostgreSQL creates `d_check` and `d_check1` for an explicit `CONSTRAINT d_check` next to an unnamed `CHECK`, so the collision is pistachio's naming rather than the input, and the comment says so.

`go vet ./...` clean and `go test -p 1 ./...` green across all packages against PostgreSQL 16, including the root integration package.

One thing I noticed while checking that README paragraph, out of scope for this PR but in the list it edits. `ChooseConstraintName` joins every column of the key, and `autoNameConstraint` takes only the first, so an unnamed multi-column `UNIQUE` gets a different name from PostgreSQL's without any duplicate involved:

```sql
CREATE TABLE public.tm (a integer, b integer, UNIQUE (a, b));
```

PostgreSQL names that constraint `tm_a_b_key`; pistachio calls it `tm_a_key`. Applying from an empty database hides it, since pistachio writes its own name into the DDL, but adopting a table that already exists does not:

```
$ pista plan mc.sql
ALTER TABLE public.tm ADD CONSTRAINT tm_a_key UNIQUE (a, b);
-- skipped: ALTER TABLE public.tm DROP CONSTRAINT tm_a_b_key;
```

That is churn on every run against an existing database, and it is a third limitation next to the suffix and the 63-byte truncation. `PRIMARY KEY` is unaffected, since its name never carries a column.

A smaller one in the same paragraph: the suffix bullet says "when multiple constraints would generate the same name", but PostgreSQL also disambiguates when a generated name meets an explicit one, e.g. `CREATE TABLE tk (a integer UNIQUE, CONSTRAINT tk_a_key CHECK (a > 0))` gives `tk_a_key` and `tk_a_key1`. pistachio rejects that file too, and only one of the two names is generated.

---

parser/parser.go
```diff
@@ -647,6 +658,81 @@ func autoNameConstraint(tableName, colName string, contype pg_query.ConstrType)
...
+
+// checkExprCols returns the single column a CHECK expression references, or nil
+// when it references none or several. PostgreSQL names such a constraint after
+// the column only in the single-column case, and it reads the expression even
+// for a constraint written on a column, so `a integer CHECK (a > b)` becomes
+// {table}_check rather than {table}_a_check.
+//
+// walkExprColumnRefs does not descend into every expression node, so an exotic
+// CHECK can come back empty and take the {table}_check form where PostgreSQL
+// would name the column.
```
winebarrel, line 711: This caveat is real, and it is worth naming the node types, because the ones that fall through are not that exotic. Diffed pistachio's name against PostgreSQL 16 for a batch of `CHECK` forms:

| CHECK | PostgreSQL | pistachio |
|---|---|---|
| `a[1] > 0` | `t_a_check` | `t_check` |
| `greatest(a, 1) > 0` | `t_a_check` | `t_check` |
| `least(a, 1) > 0` | `t_a_check` | `t_check` |
| `ROW(a) IS NOT NULL` | `t_a_check` | `t_check` |
| `a IS TRUE` | `t_a_check` | `t_check` |
| `(a COLLATE "C") > 'x'` | `t_a_check` | `t_check` |

Five node kinds: `A_Indirection` (subscripts and field selection), `MinMaxExpr` (`GREATEST` / `LEAST`), `RowExpr`, `BooleanTest`, `CollateClause`. Everything else I threw at it matches, including `coalesce`, `substring`, casts, `CASE`, `ANY (ARRAY[...])`, `BETWEEN`, `IS DISTINCT FROM`, `EXTRACT`, `~`, `LIKE`, and the jsonb operators.

Each of these plans an `ADD CONSTRAINT` under pistachio's name plus a drop of PostgreSQL's, on every run against a database that already holds the table, which is the case this commit set out to close.

`WalkExprColumnRefs` backs two other things, so those five cases go wrong in the same way there: `validateColumnRefs` misses a column that does not exist when it is only referenced through one of these nodes, and `rewriteColumnRefsInExpr` (`diff/rename.go`) leaves the old name in place when a renamed column is referenced that way. Adding the five cases to the walker fixes all three at once, and the walker is the only place that needs to change.

---

winebarrel (commented):
Bigger than what I raised, and the naming now matches the server. I checked it rather than reading it: 29 unnamed-constraint forms created in PostgreSQL 16, `conname` read back from `pg_constraint`, and the same SQL through the parser. All 29 agree, including the ones not in the test table: `EXCLUDE (a WITH =, (b + 1) WITH =)` giving `t_a_expr_excl`, column-level `UNIQUE` and `REFERENCES`, the `ALTER TABLE ADD UNIQUE` / `ADD CHECK` / `ADD EXCLUDE` forms, a `CHECK` naming its column after a repeated reference is deduped, and `(a, b) IS NOT NULL` falling to `t_check` on two columns.

Splitting the column list out into `constraintKeyCols` / `fkAttrCols` / `checkExprCols` is what makes that readable, and each one lands on the right catalog path: `Keys` for `PRIMARY KEY` and `UNIQUE`, `Exclusions` with `expr` standing in for a non-column element, `FkAttrs` for the local side of a foreign key. `checkExprCols` matching `AddRelationNewConstraints`, which pulls vars from the expression and names the column only when exactly one survives deduplication, is the subtle one, and reading the expression even for a column-written `CHECK` is the part that is easy to get wrong. The `a integer CHECK (a > b)` case pins it.

The `ALTER TABLE ... ADD FOREIGN KEY` fix is a real bug caught in passing. That path never auto-named, so an unnamed one planned `ADD CONSTRAINT  FOREIGN KEY ...` with an empty name, which the server rejects outright.

Two things.

**The migration is not self-completing under default flags.** The CHANGELOG says a database built with the old names "sees one drop and add on the next run", but a constraint drop needs `--allow-drop`, so by default only the add lands:

```
$ pista apply mig.sql
ALTER TABLE public.tm ADD CONSTRAINT tm_a_b_key UNIQUE (a, b);
-- skipped: ALTER TABLE public.tm DROP CONSTRAINT tm_a_key;

$ psql -c "SELECT conname FROM pg_constraint WHERE conrelid = 'tm'::regclass"
 tm_a_b_key
 tm_a_key

$ pista plan mig.sql
-- skipped: ALTER TABLE public.tm DROP CONSTRAINT tm_a_key;
-- No changes
```

The table is left with two identical unique constraints, so two indexes on `(a, b)`, and the skipped drop is printed on every run from then on. Worth saying in the CHANGELOG that finishing it takes `--allow-drop constraint`, or an explicit `CONSTRAINT` name in the schema file to keep the old one.

**The walker gap has five concrete instances.** Inline on `parser/parser.go`, with the PostgreSQL names next to pistachio's. Not blocking, and the comment is honest that the gap exists, but the same five also affect `validateColumnRefs` and the column-rename rewriting in `diff/rename.go`, and one change to `WalkExprColumnRefs` covers all three.

`go vet ./...` clean and `go test -p 1 ./...` green across every package against PostgreSQL 16, including the two fixture renames.

---

winebarrel (commented):
All five node kinds are covered, and I confirmed each of the three consumers the CHANGELOG names.

Names first: the six forms I reported now agree with PostgreSQL 16, and so do the harder combinations around them. `(c).f` on a composite column gives `t_c_check`, `a[b]` falls to `t_check` because the subscript is a second column, `a[1:2]` and `greatest(a[1], 1)` resolve through two layers, and `((c).f + (c).g)` dedupes to the one column. Re-ran the original 29-case matrix as a regression check; still all matching.

Walking `AIndices.Lidx` / `Uidx` but not the field-selection strings in `Indirection` is the right split. `a[b]` needs the subscript to count as a reference, and `(c).f` must not rename `f` when a column named `f` exists elsewhere. Easy to get backwards.

The other two consumers behave as claimed. `CHECK (greatest(zz, 1) > 0)` on a table with no `zz` now reports the missing column instead of passing, and a rename through one of these nodes rewrites cleanly:

```
$ pista plan rn.sql
ALTER TABLE public.t RENAME COLUMN a TO z;
```

with no `DROP CONSTRAINT` / `ADD CONSTRAINT` beside it, where the constraint is `CHECK (greatest(a, 1) > 0)`. The CHANGELOG entry covering all three in one bullet matches what the code does.

The migration note now says what actually happens, including that the table carries both constraints until the drop is allowed, and that an explicit `CONSTRAINT` clause keeps the old name.

There is a tail left, smaller than the one just closed and not worth holding this:

| CHECK | PostgreSQL | pistachio |
|---|---|---|
| `f(x => a, y => 1) > 0` | `t_a_check` | `t_check` |
| `xmlelement(name e, a) IS NOT NULL` | `t_a_check` | `t_check` |
| `JSON_OBJECT('k': a) IS NOT NULL` | `t_a_check` | `t_check` |

`NamedArgExpr`, `XmlExpr`, and the PostgreSQL 16 JSON constructors. Named-argument notation is the plausible one of the three; the other two need an XML or JSON constructor inside a `CHECK`.

Worth saying that the explicit switch is the right design even so. A generic walk over every child node would reach `FuncCall.Funcname`, the field names in `A_Indirection` and `TypeName` strings, and the visitor mutates what it visits for renames, so it would rewrite a function or field that happens to share a column's name. Enumerating the kinds is what keeps that from happening, and the cost is exactly this kind of tail.

`go vet ./...` clean and `go test -p 1 ./...` green across every package against PostgreSQL 16.

---

winebarrel (commented):
Named arguments and the XML constructors check out, and the shadowing cases are the part I wanted to see. Keeping `XmlExpr.ArgNames` out and unwrapping `ResTarget` to its `Val` means an alias cannot be mistaken for a column, which matters because these names sit in the same namespace as columns in the source text:

| CHECK | PostgreSQL | pistachio |
|---|---|---|
| `xmlelement(name a, b)` on `(a text, b text)` | `t_b_check` | `t_b_check` |
| `xmlelement(name e, xmlattributes(b AS a))` | `t_b_check` | `t_b_check` |
| `xmlforest(b AS a)` | `t_b_check` | `t_b_check` |
| `f(x => a, y => 1)` on `(x integer, a integer)` | `t_a_check` | `t_a_check` |

Each of those would have come out `t_check` if the name had leaked in as a second reference, so they pin the exclusions rather than just the additions. `xmlpi`, `xmlparse` and `xmlroot` ride along on `XmlExpr` and match too.

The TODO entry is accurate. I reproduced the `JSON_OBJECT` case it describes, and the node-kind count is the right reason to stop: `JSON_OBJECT` alone is three kinds deep and the family does not end there. The paragraph on why a generic walk is not the answer belongs in the file, since that is the option a later reader would reach for first.

One straggler in the same family, and it is the only thing I found:

```sql
CREATE TABLE t (a xml, b integer, CHECK (xmlserialize(content a AS text) > ''));
```

PostgreSQL names it `t_a_check`, pistachio `t_check`. `xmlserialize` parses to `XmlSerialize`, a separate node kind from `XmlExpr`, so the new case does not reach it. One line covers it:

```go
case *pg_query.Node_XmlSerialize:
	WalkExprColumnRefs(n.XmlSerialize.Expr, visit)
```

Otherwise it is worth a line in the TODO entry, since the CHANGELOG now says the XML constructors are handled and this one is not.

`go vet ./...` clean and `go test -p 1 ./...` green across every package against PostgreSQL 16.

---

winebarrel (commented):
`xmlserialize` matches now, and turning the four shadowing cases into `TestWalkExprColumnRefs_NamesAreNotColumns` is the better half of the commit. Those are the ones a later addition to the walker can silently break, and a test that asserts what must *not* come back is the only thing that catches it.

Ran a closing sweep over everything raised across this review, 18 forms created in PostgreSQL 16 with `conname` read back and compared against the parser. Seventeen match, including `xmlserialize`, `a IS DOCUMENT` riding on `XmlExpr`, the argument-name and attribute-alias cases, `(c).f`, `a[b]` falling to `t_check` on its second column, `EXCLUDE (a WITH =, (b + 1) WITH =)` giving `t_a_expr_excl`, and the multi-column `UNIQUE`.

The one that does not is `CHECK (a IS JSON)`, `t_a_check` against `t_check`, which is `JsonIsPredicate` and is named in the TODO entry already. Nothing new.

`go vet ./...` clean and `go test -p 1 ./...` green across every package against PostgreSQL 16.

That closes everything I raised. Reading back over the series, the shape held up: the namespace check found a naming bug, the naming fix found a walker gap, and the walker gap turned out to sit under the rename rewriting and the column validation as well. Each round the fix went to the one place that fed all the consumers, which is why the diff stayed small while the blast radius kept growing. Nothing outstanding from me.

---

PR #371, opened 2026-08-07
Walk the SQL/JSON expressions, and strip a view's qualification wherever it sits

winebarrel:
Closes the TODO entry `JSON constructors hide a column from the expression walker`, and the drift found while verifying it.

## Expression walker

`WalkExprColumnRefs` (`internal/pgast`) enumerates the node kinds it walks, and the SQL/JSON nodes PostgreSQL 16 and 17 added were not among them. Three things read columns out of an expression through it, and all three were wrong for those kinds: an unnamed `CHECK` reaching its only column that way lost the column from its generated name, giving `t_check` where PostgreSQL says `t_a_check`; the desired-schema validation missed a column that does not exist; and a column renamed with `-- pista:renamed-from` kept its old name inside the expression.

`JSON_OBJECT`, `JSON_ARRAY`, `JSON`, `JSON_SCALAR`, `JSON_SERIALIZE`, `IS JSON`, `JSON_VALUE`, `JSON_QUERY`, `JSON_EXISTS`, `JSON_OBJECTAGG` and `JSON_ARRAYAGG` are walked now. A `JSON_OBJECT` key is an expression as well, so `JSON_OBJECT(a: b)` reaches two columns and takes `{table}_check`, which is what the server does. The names written where a column could go stay out: a `PASSING` argument's name, and a `JSON_TABLE` output column name.

## Definition comparison

Fixing the name alone left the constraint drifting. `pg_get_expr` spells out the `RETURNING json` that `JSON_OBJECT`, `JSON_ARRAY` and the two aggregates return when the clause is left off, and drops the `FORMAT` of a `RETURNING` clause it does print, while the desired side keeps what was written. Neither is read as a difference now; a `RETURNING` naming another type still is. The normalizations a `CHECK` already gets, stripping a text cast and reading `= ANY (ARRAY[...])` as `IN (...)`, reach inside the constructors as well, and so does the asymmetric cast alignment.

## View qualification

Verifying the above turned up a wider drift, present before this branch. `pg_get_viewdef` returns a column reference unqualified where the desired file writes `table.column`, and `stripQualifications` (`diff/views.go`) only reached the reference through a handful of node kinds. A view whose body wrote `COALESCE(t.a, 'x')`, `CASE WHEN t.a ...`, `GREATEST`, a subscript, an array or row constructor, `IS TRUE`, `COLLATE`, an XML or SQL/JSON constructor, an aggregate `ORDER BY` or `FILTER`, an `OVER` clause, `DISTINCT ON`, `GROUP BY ROLLUP` or a sub-query nested under any of them kept the prefix, so `CREATE OR REPLACE VIEW` was re-emitted on every plan and `plan --check` stayed at exit code 2 forever.

`stripQualifications` delegates to the expression walker instead of carrying its own list of node kinds, so the two cannot drift apart again. The walker gained an optional sub-link callback, since the contained SELECT is out of scope for the other two callers but not for this one, and it visits qualified references through a new entry point. `normalizeSelectExprs` and `alignSelectCasts` follow it into `DISTINCT ON` and a named `WINDOW`, which had qualification stripping but no normalization.

## Tests

- Walker unit tests per node kind, including the PostgreSQL 17 forms, which pg_query parses on any server.
- `equalConstraintDef` and `equalViewDef` unit tests for the RETURNING normalization, the cast alignment inside the new nodes, and the differences that must still surface.
- Constraint naming table entries, each name verified against PostgreSQL 16.
- Plan and apply fixtures: no drift for the constructors, their options, a generated column, a view, an index predicate, `DISTINCT ON` and a named `WINDOW`; a `RETURNING` type change; a rename through a JSON expression; a validation error; and a view holding qualified references in every node kind above.
- The plan and apply harnesses take a `min_pg` field, which skips a fixture on a server that predates the syntax it needs.

Verified against a local PostgreSQL 16 server: `make test`, `make test-scenario`, `make lint`, `go vet`. Statement coverage goes from 94.257% at the branch point to 94.354%.

`testdata/plan/no_diff_json_query_functions.yml` is the one fixture nobody has run: it needs PostgreSQL 17, so only the 17 and 18 CI jobs exercise it. It asks the server what it prints for a query function written without a `RETURNING` clause, which is the evidence the matching TODO entry lacks.

## Left in TODO.md

`JSON(a)` is stored as a cast, so a desired file written that way still re-emits the constraint; write `a::json`. The `RETURNING` of the PostgreSQL 17 query functions is not normalized. `JSON_ARRAY` over a subquery is stored rewritten as an aggregate over a derived table, and `JSON_TABLE` is a FROM item.

Two view drifts found while testing are recorded and not fixed here: PostgreSQL rewrites `BETWEEN`, and it decides the output column alias on its own.

https://claude.ai/code/session_019h9DBfk27GEeTUjEc6ocnP


winebarrel (commented):
## Review

Checked out the branch and verified it locally: `go vet ./...`, `go test ./diff/... ./internal/pgast/... ./parser/...`, and the full `TestPlan` / `TestApply` suites against a PostgreSQL 16 server. All green. I also confirmed the naming claims directly on a 16 server (`CHECK (JSON_OBJECT('k': a) IS NOT NULL)` gets `nm_a_check`, `CHECK (JSON_OBJECT(a: b) IS NOT NULL)` gets `nm_check`) and that `pg_get_constraintdef` spells out `RETURNING json` where the written form leaves it off. No PostgreSQL 17 was available here either, so the 17-only forms are unverified on my side as well.

The shape is right. Splitting `WalkExprColumnRefNodes` out and making it the single enumeration that `stripQualifications`, the constraint namer, the validator and the renamer all share is what keeps the three from drifting apart again, and the `visitSubLink` callback is a clean way to let the view path descend into a sub-query while the CHECK path does not. The `min_pg` field is the right minimal addition for fixtures that need syntax a 15 server cannot parse.

Three things below, one of them worth acting on before merge.

### 1. `normalizeSelectExprs` was not widened alongside `stripQualifications`

`stripQualifications` now reaches `SelectStmt.DistinctClause` and `SelectStmt.WindowClause`, but `normalizeSelectExprs` and `alignSelectCasts` (`diff/views.go`) still skip both. The pair has to move together, so an expression in either clause gets its qualification stripped and no normalization at all. Confirmed against a 16 server:

```sql
CREATE TABLE public.dw (a text, n integer);
CREATE VIEW public.dwv AS
    SELECT DISTINCT ON ((dw.a IN ('x','y'))) dw.n AS c FROM public.dw ORDER BY (dw.a IN ('x','y'));
CREATE VIEW public.dwv2 AS
    SELECT sum(dw.n) OVER w AS c FROM public.dw WINDOW w AS (PARTITION BY (dw.a IN ('x','y')));
```

`pg_get_viewdef` returns `= ANY (ARRAY['x'::text, 'y'::text])` in both places, and `pista plan` against that database re-emits both views on every run:

```
CREATE OR REPLACE VIEW public.dwv AS
SELECT DISTINCT ON (dw.a IN ('x', 'y')) dw.n AS c FROM public.dw ORDER BY dw.a IN ('x', 'y');
CREATE OR REPLACE VIEW public.dwv2 AS
SELECT sum(dw.n) OVER w AS c FROM public.dw WINDOW w AS (PARTITION BY dw.a IN ('x', 'y'));
```

The same holds for a text cast the catalog adds inside `DISTINCT ON`. The inline `OVER (PARTITION BY ...)` case does work, because `normalizeAggClauses` reaches it through `FuncCall.Over`; only the named `WINDOW` form and `DISTINCT ON` are left out. This is pre-existing rather than introduced here, but the fix is a few lines in the two functions this PR is already editing, and it is the same class of bug the PR set out to close. A fixture alongside `no_diff_view_qualified_refs.yml` would pin it.

### 2. The TODO.md note on the leftover RETURNING gap reads understated

The entry says the PostgreSQL 17 constructs left out of `normalizeJsonOutput` mean "a written `RETURNING ... FORMAT JSON` on one of them may drift", and that "only the FORMAT part carries over". As far as I can tell from ruleutils, `get_json_returning` appends `RETURNING <type>` unconditionally once a return type is set, so on 17 a `CHECK (JSON_VALUE(a, '$.x') > '')` should come back as `JSON_VALUE(a, '$.x' RETURNING text)` and drift on the whole clause, not just on the FORMAT. Same for `JSON_SERIALIZE` and `JSON_QUERY`. `JSON_SCALAR` also belongs in the default-is-json group the entry excludes it from, since it returns json.

If that is right, `normalizeJsonOutput` applied to `JsonFuncExpr.Output`, `JsonParseExpr.Output`, `JsonScalarExpr.Output` and `JsonSerializeExpr.Output` would cover `JSON_SCALAR` outright, and the rest would need the default type per node kind rather than a fixed json. Since neither of us has a 17 server, the actionable part is just a `min_pg: 17` fixture for one of these forms - it either passes and the TODO entry is wrong, or it fails and the entry gets the sharper wording. Worth doing before the note ships, because a reader will take it as the record of what is left.

### 3. Two nits

`isJSONTypeName` (`diff/tables.go`) looks only at the name parts, so `RETURNING json[]` is read as `RETURNING json` and compares equal to no clause at all. The server rejects that spelling (`cannot cast type json to json[]`), so nothing reachable depends on it, but a `len(tn.GetArrayBounds()) == 0` guard makes the function say what it means.

The new `Node_SubLink` case in `WalkExprColumnRefNodes` skips `Testexpr` entirely when `visitSubLink` is nil. `Testexpr` holds the LHS of `x IN (SELECT ...)`, which is in the local scope, so walking it unconditionally and gating only `Subselect` behind the callback would match how `stripSubLinkQualifications` already treats it. Nothing reaches it today (a CHECK and an index predicate both reject a sub-query, and the validator only reads table expressions), so this is about the walker's contract rather than a live bug.

---

winebarrel:
All three addressed in b96dfc5.

### 1. `normalizeSelectExprs` / `alignSelectCasts` widened

Reproduced your case on 16 before touching anything, and the plan output matched yours. Both functions now handle `DistinctClause` and `WindowClause`, so the three normalizations move together with `stripQualifications` across all of them. `normalizeAggClauses` and the new `normalizeWindowDefExprs` share one body, so the inline `OVER` and the named `WINDOW` cannot diverge either.

`testdata/plan/no_diff_view_distinct_on_and_window.yml` pins it with three views: `DISTINCT ON` over an `IN` list, a named `WINDOW` whose `PARTITION BY` holds one and whose `ORDER BY` holds a timestamp literal, and `DISTINCT ON` over that literal. It carries no `min_pg`, so it runs from 15 up. `equalViewDef` unit tests cover the same four shapes plus a real difference in either clause.

### 2. RETURNING on the 17 query functions

Took the sharper wording. The entry now says the whole clause is expected to drift rather than the FORMAT alone, puts `JSON_SCALAR` and `JSON()` with the constructs that return json, and names the fix as a default type per node kind instead of a fixed json.

`testdata/plan/no_diff_json_query_functions.yml` is the probe you suggested, `min_pg: 17`, a view over `JSON_VALUE`, `JSON_QUERY`, `JSON_EXISTS`, `JSON_SCALAR` and `JSON_SERIALIZE` with no `RETURNING` written. A view rather than a `CHECK` so nothing depends on those functions being immutable. It skips here on 16; the 17 and 18 jobs decide it. If it fails, the failure carries the catalog form and the entry gets rewritten around that rather than around ruleutils.

### 3. Nits

`isJSONTypeName` takes the `len(tn.GetArrayBounds()) != 0` guard.

`WalkExprColumnRefNodes` walks `Testexpr` unconditionally now, and `visitSubLink` covers the contained SELECT alone. `stripQualifications` lost its own `SubLink` branch as a result, so a sub-query predicate has one path through the walker instead of two, and `stripSubLinkQualifications` is down to the `Subselect`. A walker test pins that `a IN (SELECT b FROM other)` yields `a` with or without a callback.

### Verification

`make lint`, `go vet`, `make test` and `make test-scenario` against 16, all green. Statement coverage is 94.354% against 94.257% at the branch point. The uncovered lines this branch adds are five nil guards on values the grammar always sets.

---

winebarrel (commented):
## Re-review of b96dfc5

All three points from the last round are addressed, and I re-verified each one against a local PostgreSQL 16 server rather than taking the tests at their word.

**DISTINCT ON and named WINDOW.** `normalizeSelectExprs` and `alignSelectCasts` now follow `stripQualifications` into both clauses, and pulling the `WindowDef` loop out into `normalizeWindowDefExprs` keeps the inline and named forms on one path. Rebuilt the exact reproduction from the last round - the two views plus a third with a catalog-spelled literal - and `pista plan` reports `-- No changes` where it re-emitted both before. `SELECT DISTINCT` with no `ON` also still works, which was the case I was least sure of: the raw parse leaves a nil element in `DistinctClause` and the new loop passes it through untouched. Real differences still surface: `DISTINCT` against `DISTINCT ON`, a frame bound change, and a window-count mismatch all compare unequal.

**Sub-link LHS.** Moving `Testexpr` into the walker and dropping the `SubLink` branch from `stripQualifications` is the better shape - the walker now owns the whole expression side and the callback owns only the contained SELECT, which is what the two names say. Checked that the LHS is still stripped through `IN (SELECT ...)`, `= ANY (SELECT ...)`, and a sub-link nested under a `COALESCE`, both as a top-level `WHERE` and through the walker's fallthrough.

**json[].** `RETURNING json[]` now compares unequal to a bare `RETURNING json`, and the comment says why.

`go vet` is clean and the full `go test -p 1 ./...` passes against 16.

### One thing blocks the merge

`test (17)` and `test (18)` are red on this head, from `testdata/plan/no_diff_json_query_functions.yml`. That is the fixture doing what it was built to do, and I think adding it was the right call - it is the first real evidence about the 17 query functions and it settles the TODO entry either way. But it cannot land red. The GitHub API truncates the job log before the assertion diff, so I could not read the catalog form out of it, and no 17 server is reachable from here either (`apt.postgresql.org` is blocked by the proxy), so I only have the fact of the failure, not what the server printed.

Two ways out, and I would take the first:

**Fix it.** `normalizeJsonOutput` already does the work; what it lacks is the default type per node kind rather than a fixed json. Pass the default in at the call site: json for `JSON_OBJECT`, `JSON_ARRAY`, the two aggregates, `JSON()` and `JSON_SCALAR`; text for `JSON_SERIALIZE` and `JSON_VALUE`; jsonb for `JSON_QUERY`. Then call it for `JsonParseExpr.Output`, `JsonScalarExpr.Output`, `JsonSerializeExpr.Output` and `JsonFuncExpr.Output` alongside the four that have it today. Two things to watch while scoping that: ruleutils skips the clause for `JSON_EXISTS` when it returns boolean, so that one likely round-trips already and needs nothing; and `JSON_QUERY` carries `WRAPPER` and `QUOTES` behaviour that the catalog prints as well, which is a separate question from the return type. The fixture then turns from a probe into the regression test for the fix.

**Or record it.** If the fix is out of scope for this branch, change the fixture to assert the plan the server actually produces instead of `plan: ""`. CI goes green, the catalog form lands in the repository where the TODO entry can point at it, and flipping it back to `plan: ""` is the one-line change that closes the entry later. Either way, please do not merge with 17 and 18 failing - a red matrix leg trains everyone to stop reading it.

Small note while you are in there: `min_pg` is not in the fixture-field list in `CLAUDE.md`. The list already defers to the `planTestCase` / `applyTestCase` structs as authoritative, so this is optional, but the two existing entries are close enough that adding it costs nothing.

Everything else looks ready.

---

winebarrel:
Took the first way out, in a787617. The fixture is now the regression test rather than a probe.

### The rule, without the table of defaults

Rather than pass the default type in per node kind, the clause is dropped from the current side where the desired side wrote none. That is what `alignCurrentCasts` already does with the casts the catalog spells out, and it needs no knowledge of what the server resolved, which is the part neither of us can check from here. A clause the desired side did write is compared as written, so `RETURNING text` against `RETURNING varchar` still surfaces.

It also sidesteps the two things you flagged while scoping. If ruleutils skips the clause for `JSON_EXISTS`, the rule is a no-op there and nothing had to know that in advance. And `WRAPPER` / `QUOTES` are resolved and printed the same way, so they fall under the same rule rather than being a second question: `dropCurrentJsonFuncClauses` covers the RETURNING type, the wrapper, the quote behaviour and the `ON EMPTY` / `ON ERROR` defaults together.

`JSON()` and `JSON_SCALAR` turned out to need nothing. The grammar rejects a RETURNING clause on both, which the parser confirms, so their `Output` is always nil and the entry that grouped them with the constructors was wrong about the shape rather than the default type.

The FORMAT reset now applies to the query functions too, since PostgreSQL never prints it there either.

### Tests

Unit tests in `diff/tables_test.go` cover each of these against the catalog form: the RETURNING drop for `JSON_VALUE`, `JSON_QUERY` and `JSON_SERIALIZE`, the FORMAT drop, the wrapper, quotes and `ON EMPTY` / `ON ERROR` drops, and four cases where a written clause must still differ. pg_query parses PostgreSQL 17 syntax on any server, so these run everywhere; the 17 and 18 legs are what confirm the catalog side.

The TODO entry is down to what is actually left: `JSON(a)` stored as a cast, `JSON_ARRAY` over a subquery, and `JSON_TABLE` as a FROM item.

### Verification

`make lint`, `go vet`, `make test` and `make test-scenario` against 16, all green. Coverage 94.390% against 94.257% at the branch point.

I cannot run 17 or 18 here either, so that leg is again the check. If it comes back red, the failure now names the single view that drifted, which narrows it to one function.

`min_pg` is in the `CLAUDE.md` fixture-field list now, along with the apply suite's `applied_pg16` / `applied_pg17` / `applied_pg18`, which were missing from the same list.

---

winebarrel (commented):
## Re-review of a787617

I got PostgreSQL 17.4 and 18.0 running locally this time (zonky's embedded-postgres binaries off Maven Central, since apt.postgresql.org is blocked here), so the 17-only paths are observed now rather than reasoned about. Two findings, one of which is the answer to why CI is still red.

### The 17 and 18 failure is the jsonpath literal, not the RETURNING clause

`no_diff_json_query_functions.yml` still fails on both, and the reason has nothing to do with what this commit changed. PostgreSQL parses the path argument into a `jsonpath` value and prints its canonical form back:

```
 v_exists    |  SELECT JSON_EXISTS(a, '$."z"') AS c FROM t;
 v_query     |  SELECT JSON_QUERY(a, '$."y"' RETURNING jsonb WITHOUT WRAPPER KEEP QUOTES) AS c FROM t;
 v_scalar    |  SELECT JSON_SCALAR(1) AS c FROM t;
 v_serialize |  SELECT JSON_SERIALIZE(a RETURNING text) AS c FROM t;
 v_value     |  SELECT JSON_VALUE(a, '$."x"' RETURNING text) AS c FROM t;
```

The fixture writes `'$.x'` and the catalog holds `'$."x"'`, so the three views with a path drift on the string literal. `v_scalar` and `v_serialize` have no path and already pass, which is the evidence that the RETURNING work in this commit is correct: `RETURNING text` on `JSON_SERIALIZE` is being dropped exactly as intended, and `JSON_QUERY`'s `WITHOUT WRAPPER KEEP QUOTES` is too.

Rewriting the three paths in the fixture to the spelling the server prints turns it green on both 17 and 18 - I ran the full suite against each with that one edit and got `ok github.com/winebarrel/pistachio`. Nothing else in the tree fails on 17 or 18. The canonical spelling round-trips exactly, including the compound cases:

```
JSON_VALUE(a, '$."x"')       ->  '$."x"'
JSON_VALUE(a, '$.x.y[0]')    ->  '$."x"."y"[0]'
JSON_EXISTS(a, 'strict $."z"') -> 'strict $."z"'
```

So: change the fixture to the canonical paths and it documents the fix instead of a bug, and add a TODO entry for the rewrite itself. Closing that one properly needs a jsonpath parser to canonicalise the desired side, which pg_query does not give you - the string is just an `A_Const` there - so a TODO plus the workaround (write the path the way `pista dump` emits it) looks like the right resting place. The entry is worth having: `'$.x'` is the spelling everyone writes.

### `dropCurrentJsonFuncClauses` drops too much, and hides real drift

This is the part I would not merge as written. The rule is "no clause written on the desired side means accept whatever the current side holds", but the clauses are not all-or-nothing - the server prints the resolved value whether or not it is the default, so accepting anything makes pistachio blind to a genuine difference. Built on 17: a database holding these four views, against a desired file that writes each of them with no options at all.

```
fnv1 | JSON_QUERY(a, '$."x"' RETURNING jsonb WITH UNCONDITIONAL WRAPPER KEEP QUOTES)
fnv2 | JSON_VALUE(a, '$."x"' RETURNING text ERROR ON ERROR)
fnv3 | JSON_VALUE(a, '$."x"' RETURNING character varying(10))
fnv4 | JSON_QUERY(a, '$."x"' RETURNING jsonb WITHOUT WRAPPER OMIT QUOTES)
```

```
$ pista plan -I 'fn*' fn.sql
-- Plan for schema public (1 table, 4 views, ...)
-- No changes
```

All four differ in what they return at runtime - a wrapped array instead of a bare value, an error instead of a NULL, a truncating varchar, unquoted output - and pistachio reports the schema as settled. That is the failure mode the tool exists to catch, and it is a regression against `main`, where all four re-emit the view.

The fix keeps the shape and narrows the test: drop the current side's clause only when it holds the value the server resolves for an omitted one, rather than whenever the desired side wrote nothing.

- `Wrapper`: only `JSW_NONE`.
- `Quotes`: only `JS_QUOTES_KEEP`.
- `OnEmpty` / `OnError`: only the NULL behaviour with no expression. PostgreSQL prints nothing at all when these are default, so in practice the current side is already nil there and this case costs you nothing - which is exactly why dropping unconditionally only ever fires on the non-default value.
- `Output`: only the function's own default - text for `JSON_VALUE`, jsonb for `JSON_QUERY`, text for `JSON_SERIALIZE`. This is the per-node-kind default the TODO entry described before it was deleted; it is three constants, and it is what separates `RETURNING text` from `RETURNING character varying(10)`.

The fixture still passes under the narrower rule, since the catalog prints exactly those defaults for a function written without options. Worth adding the four cases above as an `equalViewDef` or `equalConstraintDef` test asserting `false` - they are cheap and they pin the difference between "the server spelled out the default" and "the server spelled out something else".

### Small

`git diff` on the deparsed output shows `JSON_EXISTS(t.a , '$.z')` - a stray space before the comma from libpg_query's deparser for `JsonFuncExpr`. It is valid SQL and it only shows up in emitted DDL, so it is cosmetic, but it will appear in a plan any time a view with a query function is re-emitted.

The `AGENTS.md` fixture-field list and the `applied_pg16`/`17`/`18` addition are right.

Everything from the first two rounds still holds up: `go vet` clean, and the full suite passes against 15, 16, 17 and 18 with the one fixture edit.

---

PR #372, opened 2026-08-07
Walk the parse tree generically, and read the SQL/JSON expressions the way the catalog prints them

winebarrel:
Replaces #371, which is closed. That branch extended each walker's node-kind list one kind at a time and stalled; this one starts from `main` and removes the lists instead. Its PostgreSQL 17 conclusions were reasoned from `ruleutils.c` with no 17 server to hand, and two of them were wrong. Everything here was read off a running 15, 16, 17 and 18.

## Walk the tree instead of a list of node kinds

`WalkExprColumnRefs` (`internal/pgast`), `stripQualifications` (`diff/views.go`) and the expression normalizations in `diff/tables.go` each carried a `switch` over the kinds they knew, about twenty of pg_query's 268 between them. Anything outside was skipped in silence, and `normalizeSelectExprs` / `alignSelectCasts` existed only to drive the others across a statement, drifting from `stripQualifications` over which clauses they reached.

pg_query nodes are protobuf messages, so a node's children are its message-typed fields. `Walk` and `WalkPair` enumerate them through protoreflect and name no kind. The four walkers become a policy function each, and the two driver functions are gone: the walk already reaches a CTE, a set operation, a join qualifier, a sub-select, `DISTINCT ON`, a named `WINDOW` and `GROUPING SETS`. About 570 lines of switch out, 240 in.

Two rules keep it honest. The visitor is handed whole nodes and never the `String` nodes inside one, so the column renamer cannot reach a function name in `FuncCall.Funcname`, a field name in an `A_Indirection` or a type name in a `TypeName` -- the cases `TODO.md` cited against a generic walk. That entry was wrong about a walk visiting `ColumnRef` alone; the tests pinning those names pass unchanged. And the walk stops at a sub-link's contained `SELECT` for the three callers that work within one table.

This alone fixes a view whose body writes `table.column` inside `COALESCE`, `CASE`, `GREATEST`, a subscript, an array or row constructor, `IS TRUE`, `COLLATE`, an XML expression, an aggregate `ORDER BY` or `FILTER`, an `OVER` clause, `DISTINCT ON`, `GROUP BY ROLLUP` or a sub-query nested under any of them. Each kept the prefix the catalog strips, so `CREATE OR REPLACE VIEW` was re-emitted on every plan and `plan --check` never reached exit 0.

## SQL/JSON

The walk reaches the SQL/JSON nodes now, which fixes the three things that read columns out of an expression: an unnamed `CHECK (JSON_OBJECT('k': a) IS NOT NULL)` takes `t_a_check` rather than `t_check`, the desired-schema validation sees a column that does not exist, and `-- pista:renamed-from` rewrites the reference inside the expression.

Naming it correctly left it drifting, because the catalog writes these back with the clauses the server resolved. Observed on 17 and 18:

```
JSON_VALUE(a, '$.x')   ->  JSON_VALUE(a, '$."x"' RETURNING text)
JSON_QUERY(a, '$.x')   ->  JSON_QUERY(a, '$."x"' RETURNING jsonb WITHOUT WRAPPER KEEP QUOTES)
JSON_SERIALIZE(a)      ->  JSON_SERIALIZE(a RETURNING text)
JSON_OBJECT('k': a)    ->  JSON_OBJECT('k' : a RETURNING jsonb)
JSON_EXISTS(a, '$.x')  ->  JSON_EXISTS(a, '$."x"')
```

The query functions return a fixed type, so writing it and leaving it off mean the same thing and both sides drop it. `JSON_QUERY` prints its wrapper and quote behaviour whether or not they hold the default, so `WITHOUT WRAPPER` and `KEEP QUOTES` are dropped and nothing else is. `WITH UNCONDITIONAL WRAPPER`, `OMIT QUOTES`, `ERROR ON ERROR` and `RETURNING character varying(10)` each change what the expression returns at runtime and still surface, which is the failure mode #371's rule had: it dropped whatever the desired side had not written, and reported a settled schema for all four.

The constructors are the one asymmetric case. `JSON_OBJECT` over a jsonb value returns jsonb and over anything else json, so the resolved default is not knowable without type inference, which pg_query does not do. Their clause is dropped from the catalog side only where the desired file wrote none. The case that leaves uncovered is in `TODO.md`.

## jsonpath

PostgreSQL parses the path into a `jsonpath` value and stores its canonical form, so a written `'$.x'` came back as `'$."x"'` and never compared equal. This was the actual cause of #371's red 17 and 18 legs. `CanonicalJSONPath` covers the accessor grammar and refuses a filter, arithmetic or a comparison rather than guessing, since a canonicaliser that lost information would report two different paths as equal. Outside the query functions a path is an ordinary argument (`a @? '$.x'`) and only the catalog spells out the `::jsonpath` cast that identifies it, so pairing the two sides is what finds it.

## Tests

- `Walk` / `WalkPair` unit tests: replacement in a singular and a repeated field, the root, nested non-`Node` containers, `SkipSubqueries`, the `ResTarget` rule, and the kind and length guards on the pair walk.
- `CanonicalJSONPath` against the exact output of `'<path>'::jsonpath::text` on a server, plus the forms it refuses.
- `equalViewDef` / `equalConstraintDef` against the catalog strings above, including the cases that must still differ.
- Plan and apply fixtures: no drift for the constructors, the query functions, the path spellings, a rename through a JSON expression, a view holding qualified references in every node kind, an auto-named constraint with `verify_no_drift`, and a validation error. `alter_json_query_clauses.yml` pins the four differences that must surface.
- The plan and apply harnesses take `min_pg`, which skips a fixture on a server older than the syntax it needs.

## Verification

`make test` and `make test-scenario` green against 15, 16, 17 and 18 locally, `go vet`, `make lint` and `make deadcode` clean. Statement coverage 93.7%, the same as `main`. `make schema` then `pista plan` against the sample schemas reports no changes on all of them, and the gitlab schema's output is byte-identical to `main`'s.

`equalViewDef` costs about 0.37ms against 0.27ms on `main` in a microbenchmark, the price of reflection. It does not show end to end: `pista plan` over the 25k-line gitlab schema runs in 2.1s against 1.9-2.5s on `main`.

https://claude.ai/code/session_01GMx8Mt4gfR6zYEfuNdCini


winebarrel (commented):
Reviewed by reading the diff against `e2df00b` and running `go build`, `go vet`, and the `diff` / `internal/pgast` unit tests locally (no PostgreSQL in this environment, so the plan/apply fixtures were read rather than run).

The direction is right. Replacing five hand-maintained node-kind switches with one protoreflect traversal removes a whole class of silent-skip bugs, and it is a net deletion in the places that matter. `WalkPair`'s two guards (same oneof member, equal repeated length) are the right conservatism: a structural difference stops the walk instead of pairing unrelated positions. The asymmetric/symmetric split in `diff/json.go` is well argued, and `alter_json_query_clauses.yml` pinning the four clauses that must still surface is the part that makes the rest trustworthy. `min_pg` plus the existing 15/16/17/18 CI matrix means the new fixtures actually run rather than silently skipping.

Three things below. Only the first is a defect.

## 1. `alignJsonPathCast` misses the case where the desired side also writes the cast

`diff/json.go`:

```go
func alignJsonPathCast(desired, current *pg_query.Node) {
	ct := current.GetTypeCast()
	if ct == nil || !isPlainJsonTypeName(ct.TypeName, "jsonpath") {
		return
	}
	canonicalizeJsonPathLiteral(desired)
}
```

`canonicalizeJsonPathLiteral` bails unless the node it is handed is an `A_Const`. When the desired side is itself `'$.x'::jsonpath`, `desired` at this position is a `TypeCast`, so nothing is canonicalised; and because both sides carry a cast, the strip loop in `alignCastNode` returns immediately and the walk descends to the `TypeCast.arg` pair, where `current` is a bare `A_Const` and `alignJsonPathCast` returns at the first check. The literal is never canonicalised on either path.

Confirmed against this branch:

```go
// both fail
equalViewDef(`SELECT (t.a @? '$."x"'::jsonpath) AS c FROM public.t`,
             `SELECT a @? '$.x'::jsonpath AS c FROM t`)
equalViewDef(`SELECT jsonb_path_query_first(t.a, '$."x"'::jsonpath) AS c FROM public.t`,
             `SELECT jsonb_path_query_first(a, '$.x'::jsonpath) AS c FROM t`)

// passes, as the existing test covers
equalViewDef(`SELECT (t.a @? '$."x"'::jsonpath) AS c FROM public.t`,
             `SELECT a @? '$.x' AS c FROM t`)
```

So a view or constraint whose file spells the cast out re-emits `CREATE OR REPLACE VIEW` on every plan. Writing the cast is a natural thing to do here, since it is what the catalog form looks like, and `TestEqualViewDef_jsonPathOutsideQueryFunctions` only exercises the uncast desired side.

Unwrapping a matching cast on the desired side closes it:

```go
	if dt := desired.GetTypeCast(); dt != nil && isPlainJsonTypeName(dt.TypeName, "jsonpath") {
		canonicalizeJsonPathLiteral(dt.Arg)
		return
	}
	canonicalizeJsonPathLiteral(desired)
```

I applied that locally: the two cases above pass, `'$.x'::jsonpath` against a catalog `'$."y"'::jsonpath` still reports a difference, and `make test` for `diff` and `internal/pgast` stays green. Worth adding both spellings to `TestEqualViewDef_jsonPathOutsideQueryFunctions`.

## 2. The "never the String nodes underneath one" claim is not what the walk does

The top comment in `internal/pgast/walk.go`, the `WalkExprColumnRefs` comment, and the CHANGELOG entry all state that the traversal does not hand the visitor the `String` nodes inside a node, and that this is what keeps a rename off a function name, an `A_Indirection` field or a type name. The walk does reach them: `FuncCall.funcname` and `TypeName.names` are `repeated Node`, each element wrapping a `String`, and the list branch visits every one. Walking `SELECT lower(a::text) FROM t` hands the visitor `lower`, `a` and `text`.

The behaviour is still correct, but for a different reason: `WalkExprColumnRefs` acts only on `ColumnRef`, and those names are `String` nodes sitting outside any `ColumnRef`. That is worth saying as written, because this is the exact objection the old `TODO.md` entry raised against a generic walk, and a future visitor written against the current comment (one that matches `*pg_query.String` directly, say) would be unsafe in a way the comment promises it cannot be.

## 3. `SkipSubqueries` guards only `SubLink.subselect`

`isSubLinkSubselect` names one field, but `JsonArrayQueryConstructor.query` holds a `SelectStmt` directly rather than through a `SubLink`, so the walk descends into it:

```
SELECT 1 FROM t WHERE b IN (SELECT c FROM other)                     -> [b]
SELECT 1 FROM t WHERE JSON_ARRAY(SELECT c FROM other) IS NOT NULL    -> [c]
```

No user-visible bug today: all three `SkipSubqueries` callers work on CHECK expressions, index expressions and index predicates, and PostgreSQL rejects a sub-query in all of them, so the only effect is that a file writing one gets pistachio's "column does not exist" instead of the server's "cannot use subquery in check constraint". But it makes the invariant weaker than the comment states, and the ironic part is that the generic walk is what newly exposes it. Keying the skip on the field's message type (any field holding a `SelectStmt`) rather than on `SubLink` would cover both.

## Smaller notes

- `TODO.md`, `CHANGELOG.md` and the `AGENTS.md` fixture-field list are all updated, including the `JSON_TABLE` and constructor-`RETURNING` cases this deliberately does not close. That is the right level of honesty about what is left.
- The `equalViewDef` cost (0.37ms against 0.27ms) is fine given it does not show end to end, and the trade is clearly the right one.
- `internal/pgast/walk_test.go` covers the guards that matter (kind mismatch, list-length mismatch, root replacement, replacement in both singular and repeated fields), and `jsonpath_test.go` asserting against real `::jsonpath::text` output including the refusal set is the correct shape for a canonicaliser whose failure mode is reporting two different paths as equal.

Fix 1 and this looks good to merge.

---

winebarrel:
All three addressed in f32eef5. I reproduced each one before touching anything; all three were real.

### 1. `alignJsonPathCast` with the cast on both sides

Your reading of why is exactly right, and the fix is the one you proposed. Reproduced first:

```
both cast (op): false
both cast (fn): false
desired uncast : true
```

and after:

```
both cast (op): true
both cast (fn): true
desired uncast : true
```

`TestEqualViewDef_jsonPathOutsideQueryFunctions` now carries both spellings for the operator and the function form, plus `'$.x'::jsonpath` against a catalog `'$."y"'::jsonpath` still differing.

### 2. The String claim

You are right, and the wording was worse than wrong: it promised an invariant the walk does not hold. `SELECT lower(a::text) FROM t` hands the visitor `lower`, `a` and `text`, as you said.

The comment in `walk.go` now says the traversal decides nothing about which nodes matter, that those names do arrive, that what keeps the renamer off them is acting on `ColumnRef` alone, and that a visitor matching `String` directly would not be safe. `WalkExprColumnRefs` and the CHANGELOG entry say the same. This mattered more than the other two, since the old `TODO.md` entry raised this exact objection and the reason it does not apply has to be recorded accurately.

### 3. `SkipSubqueries`

Reproduced, matching your output:

```
SELECT 1 FROM t WHERE b IN (SELECT c FROM other)                   -> [b]
SELECT 1 FROM t WHERE JSON_ARRAY(SELECT c FROM other) IS NOT NULL  -> [c]
```

Keying on the field could not cover it, since `JsonArrayQueryConstructor.query` is a `Node` like any other, so the skip keys on the node being a `SelectStmt`. That covers `SubLink.subselect`, `JsonArrayQueryConstructor.query`, `CommonTableExpr.ctequery` and `RangeSubselect.subquery` alike, and the root of a walk is never checked so passing a whole statement still works. `TestWalk_SkipSubqueries` covers the sub-link, the JSON array and a CTE.

### Verification

`make test` and `make test-scenario` green against 15, 16, 17 and 18, `go vet`, `make lint` and `make deadcode` clean. Statement coverage 4976/5308 = 93.745%, against 4957/5292 = 93.670% on `main`.

Also added since your review: a `no_diff_json_generated_column.yml` fixture, because a generated column is compared through `equalSelectExpr` and nothing reached the SQL/JSON rules by that path; an assertion that a `JSON_SERIALIZE` returning a type the server never resolves to still differs; and one that a constant which is not a string is not read as a path.

---

winebarrel (commented):
Re-reviewed `5ed5f74` and `f32eef5`. All three points are addressed, and two of the fixes are better than what I suggested. Ran `go vet` and the `internal/pgast` / `diff` / `parser` / `model` suites locally, plus my own reproductions from the first round. (`golangci-lint` will not run in this container: the binary is built with go1.25 against a go1.26 `go.mod`. Still no PostgreSQL here, so the fixtures were read, not run.)

**1. jsonpath cast on the written side.** Fixed as suggested, with a `return` on the outer branch so a cast that is not `jsonpath` does not fall through. Both failing cases now pass and `'$.y'::jsonpath` still reports a difference. The four new assertions in `TestEqualViewDef_jsonPathOutsideQueryFunctions` cover exactly the matrix that was missing.

While checking for over-reach I tried `a @? '$.x'::text` against the catalog form and it compares equal, which surprised me until I traced it: `normalizeCheckExpr` strips the text-like cast before `alignJsonPathCast` ever sees the position, so it reduces to the uncast case. That is the pre-existing symmetric rule doing its job, not a new hole. `'$.x'::jsonb` still differs.

**2. `SkipSubqueries`.** Keying on the node being a `SelectStmt` rather than on `SubLink.subselect` is the better fix. I checked the two ways it could over-reach and neither bites: a set operation's arms hang off `SelectStmt.larg` / `rarg`, which are `*SelectStmt` rather than `Node`, so `SELECT a FROM t UNION SELECT b FROM u` still yields `[a b]`; and the root of a walk is never tested, so passing a whole statement in still works. `RangeSubselect.subquery` is now skipped where it was not before, which is the stricter and correct reading. Confirmed:

```
SELECT a FROM t WHERE JSON_ARRAY(SELECT c FROM other) IS NOT NULL -> [a]
SELECT a FROM (SELECT c FROM other) s                             -> [a]
SELECT a FROM t UNION SELECT b FROM u                             -> [a b]
```

The three-shape test in `TestWalk_SkipSubqueries` pins it.

**3. The String-node comment.** Now says what the walk does and why the behaviour is safe anyway, and the "do not write one" warning is the part that was actually missing. `walk.go`, `pgast.go` and the CHANGELOG all agree. `SELECT lower(a::text) FROM t` still hands over `lower`, `a` and `text`, as the new wording says.

**On the two gaps you found yourselves:** `equalSelectExpr` (`diff/policies.go`) being a third entry point into these rules is a good catch, and `no_diff_json_generated_column.yml` is the right way to pin it. Worth noting that it walks from `curTarget.Val` rather than the statement, so `Ctx` is zero at the root and the `IsSelectTarget` guard does not apply there. That is correct for a generated column, where `pg_get_expr` does add a cast that should be stripped, but it is a real asymmetry with `equalViewDef` and might deserve a line somewhere if anyone later wonders why the same expression is treated differently in the two places. Not blocking.

`nodeOneof` dropping the empty-oneofs guard is safe: both call sites are inside a `*pg_query.Node` type assertion, and the nil result for an empty Node is still handled by each caller.

Nothing left from my side. LGTM once CI is green on all four server versions.

---

PR #373, opened 2026-08-07
Strip the schema from a function call before comparing expressions

winebarrel:
## What

A CHECK constraint calling a function schema-qualified, written as `CHECK (public.lower_v(v) <> 'x')`, was re-emitted as `DROP CONSTRAINT` + `ADD CONSTRAINT` on every plan. Applying it changed nothing, so `plan --check` stayed at exit code 2 forever. `pg_get_constraintdef` prints the call unqualified when the function's schema is on the search_path, while the desired side keeps what the file says.

TODO.md recorded this as a CHECK-constraint problem. `normalizeCheckExpr` is shared by every expression comparison, so one branch in `normalizeExprNode` closes all of these, each verified against a live server:

| path | what was re-emitted |
|---|---|
| table CHECK, domain CHECK | `DROP` + `ADD CONSTRAINT` |
| index expression, partial-index predicate | `DROP` + `CREATE INDEX` |
| column DEFAULT, domain DEFAULT | `SET DEFAULT` |
| RLS `USING` / `WITH CHECK` | `ALTER POLICY` |
| view body | `CREATE OR REPLACE VIEW` |
| materialized view body and its index | `DROP` + `CREATE MATERIALIZED VIEW` |

A stored generated column did not drift, it failed. `diffColumns` treats an expression inequality as an error, so a column written `GENERATED ALWAYS AS (public.f(v)) STORED` stopped the run with `cannot change GENERATED expression; DROP COLUMN + ADD COLUMN is required` and produced no plan at all.

## Second fix

Writing the tests turned up a related gap. An `EXCLUDE` constraint keeps its elements in `Exclusions` and its predicate in `WhereClause`, both outside the `RawExpr` that `equalConstraintDef` normalized, so no expression normalization reached it: not the new schema strip, not the pre-existing text-like cast strip, not the `= ANY(ARRAY[...])` canonicalisation. `normalizeConstraintIndexExprs` closes that.

## pg_catalog is exempt

The deparser reads a `pg_catalog` prefix off the funcname to print `SUBSTRING(x FROM y FOR z)`, `POSITION`, `OVERLAY`, `EXTRACT`, `OVERLAPS`, `TRIM`, `AT TIME ZONE`, `NORMALIZE`, `IS NORMALIZED`, `COLLATION FOR`, `xmlexists` and `SYSTEM_USER` in keyword form, and `SIMILAR TO` carries a `pg_catalog.similar_to_escape` call matched the same way. Stripping the prefix rewrites what those deparse to.

An equality assertion cannot guard this. The strip is symmetric, so without the guard both sides degrade together and still compare equal. `TestNormalizeCheckExpr_keywordSyntaxLeftAlone` pins the deparsed output for all 13 forms instead.

## Tradeoff

Two same-named functions in different schemas now compare equal. The diff does not thread the search_path, so it cannot say which schema the unqualified form resolves to. `stripQualifications` already takes the same bargain for a table and a column reference. Recorded in TODO.md along with what still drifts: a `pg_catalog` prefix, a three-part name, and the asymmetric cast alignment inside an `EXCLUDE`.

`OPERATOR(public.=)` is the same bug class and is left out. `A_Expr.Name` means different things per `A_Expr_Kind`, so the strip needs an `AEXPR_OP` narrowing with its own round-trip tests, and a qualified operator appears nowhere in `testdata/` or in any sample schema. TODO entry added.

## Tests

4 fixtures and 20 Go tests. Each path was confirmed to drift with the implementation stashed out.

- `no_diff_qualified_function_call.yml` covers 12 paths, one object each, so a failure names the one that drifted
- `no_diff_generated_column_qualified_function.yml` is separate because it errored rather than planned
- `change_qualified_function_call.yml` and three Go tests are the negative cases
- `qualified_function_check_idempotent.yml` proves the round trip with `verify_no_drift`

`diff` coverage is 96.1%, unchanged. `stripFuncCallSchema`, `normalizeConstraintIndexExprs` and `normalizeExprNode` are at 100%.

Verified with `make test` on PostgreSQL 15, `make lint`, `make vet`, `make deadcode`, `make test-scenario`, and `make test-samples` (33 real-world schemas, all PASS).

https://claude.ai/code/session_011X3tP56QxVfoL27aQzXM1q


diff/tables.go
```diff
@@ -665,6 +704,33 @@ func equalConstraintDef(current, desired string) bool {
...
+	// so the IndexElem lookup skips the second item on its own.
+	for _, ex := range con.Exclusions {
+		for _, item := range ex.GetList().GetItems() {
+			if ie := item.GetIndexElem(); ie != nil && ie.Expr != nil {
+				ie.Expr = normalizeCheckExpr(ie.Expr)
+			}
+		}
+	}
+	if con.WhereClause != nil {
+		con.WhereClause = normalizeCheckExpr(con.WhereClause)
```
winebarrel, line 730: Reaching `con.WhereClause` here is right, and it makes the gap in the two sibling walkers over a constraint definition visible. Both stop at `Exclusions` and never look at `WhereClause`, while their index-side counterparts do walk `is.WhereClause`:

- `rewriteColumnsInConstraintDef` (`diff/rename.go:571`) rewrites `RawExpr` and each `Exclusions` `IndexElem`, but not `con.WhereClause`. `rewriteColumnsInIndexDef` handles `is.WhereClause` at `diff/rename.go:559`.
- `collectColumnRefsInConstraintDef` (`parser/validate.go:138`) collects from `Keys`, `Including`, `RawExpr` and `Exclusions`, but not `con.WhereClause`. `collectColumnRefsInIndexDef` appends `walkExprColumnRefs(is.WhereClause)` at `parser/validate.go:129`.

Checked both against this branch:

```
rewriteColumnsInConstraintDef(
  "EXCLUDE USING gist (during WITH &&) WHERE (v IS NOT NULL)",
  map[string]string{"during": "dur", "v": "val"})
-> "EXCLUDE USING gist (dur WITH &&) WHERE (v IS NOT NULL)"

collectColumnRefsInConstraintDef(
  "EXCLUDE USING gist (during WITH &&) WHERE (nosuch IS NOT NULL)")
-> [during]
```

The element renamed, the predicate did not, so a column renamed with `-- pista:renamed-from` produces an `ADD CONSTRAINT ... WHERE (v ...)` naming a column that no longer exists. And a predicate referencing a column that is not on the table passes desired-schema validation.

Same blind spot, same two lines to close. Worth folding in here, or recording in TODO.md next to the `alignCurrentCasts` note.


---

diff/tables.go
```diff
@@ -665,6 +704,33 @@ func equalConstraintDef(current, desired string) bool {
...
+// `EXCLUDE USING gist ((public.lower_v(v)) WITH =)` was compared verbatim
+// against the `lower_v(v)` the catalog prints and the constraint was recreated
+// on every plan. The same holds for the text-like cast strip and the
+// `= ANY(ARRAY[...])` canonicalisation, which never reached an EXCLUDE either.
+//
+// Only the symmetric normalizations run here. alignCurrentCasts is asymmetric
+// and walks a pair of trees, which the exclusion list would have to be paired
+// up element by element to support; a non-text cast the catalog spelled out
+// inside an EXCLUDE expression is still read as a difference.
+func normalizeConstraintIndexExprs(con *pg_query.Constraint) {
```
winebarrel, line 719: Naming nit: nothing here touches an index. The function normalizes an EXCLUDE's exclusion elements and its predicate, so `normalizeConstraintExclusionExprs` (or just `normalizeExcludeExprs`) says what it does. `normalizeIndexStmt` right below is the one that works on an `IndexStmt`, and the two names read as a pair today when they are not.

Also, the two calls at lines 697-698 sit after `alignCurrentCasts`. The result is the same because `alignCurrentCasts` only walks `RawExpr`, but grouping the symmetric normalizations together before the asymmetric one matches the order the doc comment describes.


---

winebarrel (commented):
Reviewed the branch at `a95bcf6`. The fix is well placed: `normalizeExprNode` is the one point every expression comparison already funnels through, so a single `FuncCall` branch covers the CHECK, domain CHECK, index expression, index predicate, DEFAULT, RLS `USING` / `WITH CHECK`, view and materialized-view paths at once, and the generated-column path that failed rather than drifted. The walk is post-order, so a nested call is reached, and the early return skips only `normalizeJsonClauses`, which matches on `Json*` node kinds a `FuncCall` can never be. All 14 checks are green and `go test ./diff/` passes locally.

`TestNormalizeCheckExpr_keywordSyntaxLeftAlone` is the right shape for the `pg_catalog` guard. An equality assertion really cannot catch a symmetric degradation, and pinning the deparsed output for all 13 forms does. I confirmed the guard is load-bearing rather than defensive: `substring(v FROM 1 FOR 2)` parses to `pg_catalog.substring` while `substring(v, 1, 2)` parses to a bare `substring`, so dropping the exemption would rewrite the first into the second.

Three things.

**1. The `WhereClause` blind spot is wider than the comparison.** Left as an inline comment on `normalizeConstraintIndexExprs`. In short: `rewriteColumnsInConstraintDef` (`diff/rename.go`) and `collectColumnRefsInConstraintDef` (`parser/validate.go`) both walk `RawExpr` and `Exclusions` and stop, while their index-side counterparts do walk `is.WhereClause`. I confirmed on this branch that a rename leaves an EXCLUDE predicate holding the old column name, and that a predicate naming a column the table does not have passes validation. Same class as what this PR found, two lines each to close.

**2. The strip is wider than the search_path needs to be.** `stripFuncCallSchema` strips any two-part name, so `a.f(v)` and `b.f(v)` compare equal. TODO.md records that and points at the missing schema list, but the diff may already have enough to narrow it: `apply` sets `search_path` to the target schemas plus `public` (README), and `stripTypeSchema` already takes the narrower shape of stripping only the container's own schema. Stripping only when the qualifier is the owning object's schema or `public` would still fix the reported bug, which is `public.lower_v` on a table in `public`, while keeping a move between two non-public schemas visible. `equalConstraintDef` has no schema parameter today, though `equalFKDef` and `equalTypeName` do, so the plumbing is partly there. Was the narrower rule considered and rejected for a reason worth adding to the TODO entry?

**3. Nothing else regressed that I could find.** I checked whether the newly reaching text-like cast strip could hide a real EXCLUDE difference. It does not: `EXCLUDE (((v)::text) WITH =)` still compares unequal to `EXCLUDE (v WITH =)`, because one side lands in `IndexElem.Expr` and the other in `IndexElem.Name`, so the strip never gets the chance to make them equal. Worth a line in the doc comment, maybe, since the comment currently reads as if the cast strip now applies to an EXCLUDE unconditionally.

Optional, low priority: the `pg_catalog` exemption is blanket, so `pg_catalog.length(v)` keeps drifting. Narrowing it to the specific keyword-form names would close that, and `keywordSyntaxExprs` is already the table that would guard it. The blanket rule is the safer default while pg_query owns that list, so this only matters if someone reports it.

The fixture layout is good. One object per path in `no_diff_qualified_function_call.yml` so a failure names the path, the generated column split out because it errored rather than planned, and `verify_no_drift` on the round trip.


---

PR #377, opened 2026-08-08
Compare SQL/JSON expressions as written

winebarrel:
Removes `diff/json.go` and `internal/pgast/jsonpath.go`.

Both were solving a problem the round trip does not have. `pista dump` emits what the catalog holds, so a file written that way already compares equal. Everything the two of them covered was drift that appears only when the file is written some other way, which the design policy calls low priority.

One of the rules was worse than the drift it removed. `JSON_OBJECT`, `JSON_ARRAY`, `JSON_OBJECTAGG` and `JSON_ARRAYAGG` resolve their `RETURNING` type from their argument types, and pg_query does no type inference, so the rule dropped the catalog's clause wherever the file wrote none. A database holding `JSON_OBJECT('k': 1 RETURNING jsonb)` against a file writing `JSON_OBJECT('k': 1)` reported no difference where the two return different types. It failed silently instead of re-planning.

## Verification

Built a schema using the constructors, the aggregates, the PostgreSQL 17 query functions, a generated column, an unnamed `CHECK`, the jsonpath operators and `JSON_TABLE` on 16, 17 and 18, dumped it, and planned the dump back. `No changes` on all three, with the normalization gone.

Full test suite passes on 15, 16, 17 and 18. `make lint` is clean and the scenario tests pass.

## Fixtures

The `no_diff_json_*` fixtures paired a catalog form against a shorter written one, which is the case that no longer compares equal. Both sides are now written the way `pista dump` emits them, which is what has to feed back clean. `alter_json_query_clauses` stays as it was, since a clause the file writes differently still has to surface; only its header comment changed.

Three TODO.md entries turn out not to hit the round trip, and are now fixtures:

- a jsonpath holding a filter (`'$."x"?(@ > 1)'`) or arithmetic (`'($."x" + 1)'`), in `no_diff_json_path_spelling`
- `JSON_TABLE`, which was recorded as unchecked, in the new `no_diff_json_table`. The server adds a `LATERAL` and names the row pattern `json_table_path_0`, and `dump` emits both, so the dump feeds back clean.

The two CHANGELOG entries this reverses were unreleased, so they are removed rather than superseded.


winebarrel:
## Review

The direction is right, and the correctness argument for it is the strongest part. I reproduced the silent failure on `main` to confirm it: a database holding `JSON_OBJECT('k': t.a RETURNING jsonb)` against a file writing `JSON_OBJECT('k': t.a)` planned `-- No changes`, and with this branch the same pair emits `CREATE OR REPLACE VIEW`. Trading 900 lines of inference for a difference that surfaces is a good trade, and it lines up with the design policy the last few PRs settled on.

I could only run PostgreSQL 16 here, so the 17/18 claims in the description are unverified on my side. On 16: `go vet` clean, full suite green, and a schema using the constructors, both aggregates, a generated column, an unnamed `CHECK`, `IS JSON`, `@?` and `jsonb_path_query_array` dumped and planned back to `No changes`. I also confirmed the two removed CHANGELOG entries were unreleased - the `[1.25.0]` section carries no SQL/JSON entry - so nothing released is being reverted without a note.

Three things I would fix before merging.

### 1. Nothing pins the behaviour this PR exists to fix

The headline justification is that a catalog `RETURNING` clause against a file writing none must now surface. `alter_json_query_clauses.yml` covers the PG17 query functions, not the constructors, and the constructor case is the one that used to fail silently. Every constructor fixture in the branch has both sides writing the clause, so the suite would still pass if the old asymmetric rule came back.

This is testable on 16. I added the fixture below to the branch and it passes:

```yaml
# testdata/plan/alter_json_constructor_returning.yml
#
# JSON_OBJECT and JSON_ARRAY resolve RETURNING json or jsonb from their
# argument types, and pg_get_viewdef prints the one the server picked. A file
# leaving the clause off returns whatever pistachio would resolve, which it
# cannot know, so the catalog's clause has to surface rather than be dropped.
min_pg: 16
init: |
  CREATE TABLE public.t (
      a jsonb,
      b text
  );
  CREATE VIEW public.v_obj AS SELECT JSON_OBJECT('k': t.a RETURNING jsonb) AS c FROM public.t;
  CREATE VIEW public.v_arr AS SELECT JSON_ARRAY(t.b RETURNING jsonb) AS c FROM public.t;
desired: |
  CREATE TABLE public.t (
      a jsonb,
      b text
  );
  CREATE VIEW public.v_obj AS SELECT JSON_OBJECT('k': t.a) AS c FROM public.t;
  CREATE VIEW public.v_arr AS SELECT JSON_ARRAY(t.b) AS c FROM public.t;
plan: |
  CREATE OR REPLACE VIEW public.v_arr AS
  SELECT JSON_ARRAY(t.b) AS c FROM public.t;
  CREATE OR REPLACE VIEW public.v_obj AS
  SELECT JSON_OBJECT('k': t.a) AS c FROM public.t;
```

### 2. `alter_json_query_clauses.yml` no longer isolates what its comment claims

The desired side still writes `'$.x'` while the catalog holds `'$."x"'`. Now that the path is compared as written, that spelling alone makes all four views differ, so the fixture would pass even if clause comparison were removed entirely. Its header says "Every one of them has to surface", but nothing in it depends on the clauses any more.

I confirmed the path-alone difference on 16 through the operator position: a view stored as `t.a @? '$."x"'::jsonpath` against a file writing `'$.x'` re-emits on this branch.

Writing the desired paths canonically (`'$."x"'`) restores the fixture to testing only the clause. The `plan` block needs the same spelling, since it echoes the desired side.

### 3. Nothing pins the `dump` half of the round trip

The plan fixtures hand-write the catalog form and assert it compares equal, which is the second half. The first half - that `dump` actually emits the `RETURNING` clause and the canonical path - is asserted nowhere. `testdata/dump/json_types.yml` only covers `json`/`jsonb` column types, and `add_json_check_constraint.yml` is the only JSON fixture with `verify_no_drift`. Since "a dump feeds back clean" is now the whole argument for calling the remaining drift low priority, one `dump` fixture over a view using the constructors, or an apply fixture with `verify_no_drift: true`, would pin it.

### Smaller notes

- `no_diff_json_path_spelling.yml` no longer tests spelling variance - both sides are identical now. It still earns its place as a check that each canonical form parses and deparses stably, but the name points at the old purpose. `no_diff_json_path_canonical` would match what it does.
- TODO.md says "Observed on PostgreSQL 16, 17 and 18" over a list that includes `JSON_QUERY` and `JSON_TABLE`, neither of which exists before 17. The path and constructor points hold on 16; the rest are 17+.
- In `normalizeExprNode`, the `return node` was dropped from the `A_Expr` branch but kept in the `TypeCast` branch, and both are now equally redundant. Worth making the two branches agree.

---

winebarrel:
All three fixed, plus the smaller notes. Pushed as b1b9968.

### 1. The constructor case is now pinned

Added `alter_json_constructor_returning.yml`, close to what you wrote. One change: your `v_obj` stored `RETURNING jsonb` over a `jsonb` argument, which is the type the server would have resolved anyway, so the fixture caught the regression without the two sides actually returning different types. Both views now store the type that is **not** the default for their argument (`RETURNING json` over `jsonb`, `RETURNING jsonb` over `text`), so a rule that dropped the clause would be reporting a settled schema for a view that returns something else.

Verified it does catch the regression: with `diff/json.go` and the two call sites restored, the fixture gets `-- No changes` where it expects two `CREATE OR REPLACE VIEW`.

### 2. `alter_json_query_clauses` isolates the clause again

You were right that the path spelling alone accounted for all four. Rather than only canonicalising the path, the desired side now spells out every clause but the one under test, so each view differs in exactly one thing:

| view | catalog | desired |
| --- | --- | --- |
| `v_wrapper` | `RETURNING jsonb WITH UNCONDITIONAL WRAPPER KEEP QUOTES` | `RETURNING jsonb WITHOUT WRAPPER KEEP QUOTES` |
| `v_quotes` | `RETURNING jsonb WITHOUT WRAPPER OMIT QUOTES` | `RETURNING jsonb WITHOUT WRAPPER KEEP QUOTES` |
| `v_on_error` | `RETURNING text ERROR ON ERROR` | `RETURNING text` |
| `v_returning` | `RETURNING character varying(10)` | `RETURNING text` |

The path-alone difference you confirmed through the operator position is now its own fixture, `alter_json_path_spelling.yml`. It needs no `min_pg`, since `@?` and `jsonb_path_query_array` work on 15.

### 3. The `dump` half is pinned

Added `testdata/dump/json_expressions.yml`: an init schema that writes neither the `RETURNING` clause nor the canonical path, and an expected dump that carries both.

That needed `min_pg` on the dump harness, which only had `dump_pg16`/`dump_pg17`/`dump_pg18`. Added as one optional field, per the "extend the struct rather than keep it in Go" rule in AGENTS.md; the field list there is updated too.

### Smaller notes

- Renamed to `no_diff_json_path_canonical.yml`, and the header now says what it tests: that each canonical form parses and deparses stably, with a pointer to `alter_json_path_spelling` for the other direction.
- TODO.md now carries the servers per form: the path on every supported server (I confirmed `'$.x'` -> `'$."x"'` on 15), the constructors on 16+, the query functions and `JSON_TABLE` on 17+.
- `normalizeExprNode`: the `TypeCast` branch's `return node` is gone, so both branches fall through.

### Verification

Full suite green on 15, 16, 17 and 18, `make lint` clean, scenario tests pass. The two new `alter_*` fixtures were each confirmed to fail with the old rules restored.

winebarrel:
## Re-review of b1b9968

All three are addressed, and the two changes you made beyond what I asked for are the better call. LGTM.

### What I verified on PostgreSQL 16

`go vet` clean, full suite green.

**The constructor fixture now bites.** Your point about the argument type is right, and I confirmed the resolution on 16:

```
d_obj | SELECT JSON_OBJECT('k' : a RETURNING jsonb) AS c FROM t;   -- jsonb arg -> jsonb
d_arr | SELECT JSON_ARRAY(b RETURNING json) AS c FROM t;           -- text arg  -> json
```

So `v_obj` storing `RETURNING json` and `v_arr` storing `RETURNING jsonb` are each the type the server would not have picked. My version had `v_obj` storing the resolved default, which made the fixture pass for the weaker reason - it caught the dropped clause without the two sides actually returning different types. Yours pins the silent failure itself.

**Both new `alter_*` fixtures fail with the old rules restored.** I put `diff/json.go` and `internal/pgast/jsonpath.go` back, re-added the three call sites in `diff/tables.go`, and re-ran:

```
--- FAIL: TestPlan/alter_json_constructor_returning
--- FAIL: TestPlan/alter_json_path_spelling
```

`TestDump/json_expressions` passes either way, which is right - it pins the catalog side, and the diff rules never reached it. That is the half that was missing, and it is now the thing the "low priority" argument rests on.

**`alter_json_query_clauses` one-clause-at-a-time.** Spelling out every clause but the one under test is a stronger fix than the canonicalisation I suggested, since it rules out the `RETURNING` type as well as the path. I could not run it - no 17 or 18 server here - so the table in your comment and everything else behind `min_pg: 17` is unverified on my side.

### Two small leftovers

**`no_diff_json_path_canonical.yml` keeps `min_pg: 17`, but `v_op` and `v_func` do not need it.** Those two are the ones you added precisely because the operator and function positions work on every supported server, and `alter_json_path_spelling` uses them ungated for the drift direction. Gated at 17 they leave 15 and 16 with no fixture asserting a canonical path in that position is settled - the only jsonpath plan fixture that runs there is the one asserting a difference. Splitting the two ungated views into their own fixture would make the pair symmetric.

**The stray-space note points at a TODO entry that does not cover it.** The fixture now says the spaces "before the comma and around the RETURNING type" are recorded in TODO.md, but "Deparsed SQL/JSON query functions carry a stray space" only describes the comma. The expected plan carries two more shapes - `RETURNING text )` and the double space in `RETURNING jsonb  WITHOUT WRAPPER`. Widening the TODO entry, or narrowing the fixture comment to the comma, would keep the pointer honest.

Neither blocks merging.

---

winebarrel:
Both leftovers fixed in 561c88f.

**`no_diff_json_path_canonical` no longer gates the two positions that work everywhere.** `v_op` and `v_func` moved to their own fixture, `no_diff_json_path_argument.yml`, with no `min_pg`. The eight query-function views stay behind `min_pg: 17`. On 15 the pair now runs symmetric:

```
--- PASS: TestPlan/no_diff_json_path_argument
--- PASS: TestPlan/alter_json_path_spelling
--- SKIP: TestPlan/no_diff_json_path_canonical
```

**The stray-space TODO entry covers all three shapes.** It described only the space before the comma; the deparser also puts one before the closing parenthesis of a bare `RETURNING` clause and two after a `RETURNING` type:

```
JSON_VALUE(t.a , '$."x"')
JSON_VALUE(t.a , '$."x"' RETURNING text )
JSON_QUERY(t.a , '$."x"' RETURNING jsonb  WITHOUT WRAPPER KEEP QUOTES)
```

All three are listed, with a pointer to `alter_json_query_clauses.yml` as the fixture that pins them. The fixture's own comment names the three positions rather than saying "around the RETURNING type".

Full suite green on 15, 16, 17 and 18; `make lint` clean.

winebarrel:
## Re-review of 561c88f

Both leftovers are closed. Confirmed on 16, where the pair runs the same way you reported on 15:

```
--- PASS: TestPlan/no_diff_json_path_argument
--- PASS: TestPlan/alter_json_path_spelling
--- SKIP: TestPlan/no_diff_json_path_canonical
```

Full suite green, `go vet` clean. The split is the right shape - the eight query-function views keep the `min_pg: 17` they need, and the two positions that work everywhere are now pinned in both directions on every supported server. Widening the TODO entry to all three deparser shapes rather than trimming the fixture comment is the better of the two options I offered, since the fixture is what pins them.

### One dangling pointer left by the split

Two comments still name `no_diff_json_path_canonical` where they now mean `no_diff_json_path_argument`:

- `testdata/dump/json_expressions.yml:7` - "The second half ... is testdata/plan/no_diff_json_constructors and no_diff_json_path_canonical". The dump fixture is `min_pg: 16` and its jsonpath is `a @? '$."x"'::jsonpath`, the ordinary-argument position. On 16 that sends the reader to a fixture that skips and no longer covers that position at all.
- `testdata/plan/alter_json_path_spelling.yml:1` - "The other side of no_diff_json_path_canonical". Its two views are `v_op` and `v_func`, so the settled counterpart is now `no_diff_json_path_argument`. `no_diff_json_path_argument` already points back correctly, so only this direction is stale.

Comment-only, and nothing else references the pre-split file. Fix those two and this is done from my side.

---

winebarrel:
Fixed in e279002. Both now point at `no_diff_json_path_argument`, which is the fixture holding the operator and function positions each of them is about. Swept the rest of the cross-references while there; nothing else named the pre-split file.

winebarrel:
## Re-review of e279002

Both pointers now name the fixture that actually covers the position. Nothing else changed, and nothing outside the two comment lines was touched.

I checked every cross-reference in `testdata/` and TODO.md rather than just the two you fixed - all eight resolve to a file that exists, and the two directions of each pair agree:

| from | to |
| --- | --- |
| `dump/json_expressions` | `no_diff_json_constructors`, `no_diff_json_path_argument` |
| `plan/alter_json_path_spelling` | `no_diff_json_path_argument` |
| `plan/no_diff_json_path_argument` | `alter_json_path_spelling` |
| `plan/no_diff_json_path_canonical` | `alter_json_path_spelling`, `no_diff_json_path_argument` |
| `plan/alter_json_query_clauses` | `no_diff_json_query_functions` |
| `plan/no_diff_json_query_functions` | `alter_json_query_clauses` |
| `TODO.md` | `alter_json_query_clauses` |

Full suite green on 16, `go vet` clean.

No further comments - this is good to merge from my side. Everything I could not run stays behind `min_pg: 17`, so the 17 and 18 results rest on your runs and CI.

---

PR #378, opened 2026-08-08
Read the catalog under a search_path we set

winebarrel:
The catalog reads the schema through `pg_get_viewdef`, `pg_get_constraintdef`, `pg_get_expr` and `format_type`, and each of them drops the schema from an object the session can reach unqualified. The connection never set `search_path`, so a server-side `ALTER ROLE ... SET search_path`, or its database-level form, decided what `dump` wrote and what the diff compared.

Measured on PG 15.18 with `myschema` on the path, against the same database read without it:

| Catalog output | search_path unset | `search_path=myschema` |
|---|---|---|
| `pg_get_viewdef` | `FROM myschema.users` | `FROM users` |
| `format_type` | `myschema.mood` | `mood` |
| `pg_get_expr` | `nextval('myschema.counter'::regclass)` | `nextval('counter'::regclass)` |
| `pg_get_constraintdef` | `CHECK (myschema.non_empty(name))` | `CHECK (non_empty(name))` |

Two things follow. A dump of that schema comes out unqualified and reloads into whatever schema the next session resolves the bare names to. And a desired schema that qualifies its own objects plans on every run, since neither `equalDefault` nor `normalizeCheckExpr` touches a schema qualification:

```
ALTER TABLE myschema.users ALTER COLUMN n SET DEFAULT nextval('myschema.counter'::regclass);
ALTER TABLE myschema.users DROP CONSTRAINT users_name_check;
ALTER TABLE myschema.users ADD CONSTRAINT users_name_check CHECK (myschema.non_empty(name));
```

`pg_get_indexdef` always qualifies, and the serial detection in `catalog/columns.go` renders both sides the same way, so neither was affected.

## Change

Every connection now sets `search_path`, to PostgreSQL's own default `"$user", public` rather than the target schemas. That keeps the output form the diff's normalizations and the dump's reloadable SQL are already written against: an object outside `public` stays qualified. `--search-path` (env `$PISTA_SEARCH_PATH`) chooses another. `--search-path=myschema` gives the unqualified output above, and `--search-path=` qualifies everything, which is the one value whose dump reloads the same way wherever it runs.

It goes out as a startup parameter, so it costs no extra round-trip. `apply` issues its own `SET` later, so the DDL it runs still resolves an unqualified reference against the target schemas. Pre-SQL runs before that `SET`, and so under this setting.

## What this does not fix

The default keeps `"$user"`, so a schema named after the connecting role stays on the path and is still read without its schema. A dump taken as that role is only reloadable by a role of that name, and the plan drift above reappears for it. `--search-path=public` drops the entry and settles all of it. The default keeps PostgreSQL's own value because that is what pre-SQL and anything else sharing the connection expect; whether `public` should be the default outright is recorded in TODO.md. `TestDump_SchemaNamedAfterRole` pins the behaviour as it stands, both halves.

## Tests

Added, each failing on `main` and passing here:

- `TestDump_ServerSideSearchPath`, one assertion per catalog function above.
- `TestPlan_ServerSideSearchPath`, the drift shown above.
- `TestDump_SearchPathOption`, `TestDump_SearchPathEmpty`, `TestConnect_SearchPathOption`, the option and its empty value.
- `TestDump_SchemaNamedAfterRole`, the `"$user"` case and the way out of it.
- `TestConnect_SearchPath`, `TestConnect_SearchPathIgnoresServerSide`, `TestOptions_SearchPathDefault`, `TestOptions_SearchPathEmpty`, `TestOptions_SearchPathFromConfig`.

`PGOPTIONS` stands in for `ALTER ROLE ... SET search_path`; it reaches the backend the same way without altering a role the whole test run shares.

Coverage stays at 97.5%, with `connect` at 100%. `make lint` reports 0 issues and `make test-scenario` passes.

https://claude.ai/code/session_013MFBBShsEJ8WrbRB4SQyvd


options.go
```diff
@@ -6,12 +6,26 @@ import (
...
+// reads the schema through pg_get_viewdef, pg_get_constraintdef, pg_get_expr
+// and format_type, each of which drops the schema from an object the session
+// can reach unqualified, so this value decides what dump writes and what the
+// diff compares. Left to a server-side "ALTER ROLE ... SET search_path", one
+// database would read differently depending on who connected.
+//
+// kong defaults the Options field to this, and connect falls back to it, so a
+// library caller that builds Options directly connects the same way the CLI
+// does.
+const DefaultSearchPath = `"$user", public`
```
winebarrel, line 26: The `"$user"` entry keeps the role-dependence this PR is written to remove. The description lists it as part of the problem ("A schema named after the connecting role did the same through the default `"$user"` entry, so one database read differently depending on who connected"), but the fix pins the path to a value that still contains it, so the case survives -- it is now hard-coded rather than inherited.

Reproduced on this branch (`6d112cf`, PG 16.13), with role `myschema` owning schema `myschema` and no server-side setting anywhere:

```
$ ./pista dump -c postgres://postgres@localhost/postgres -n myschema
    m myschema.mood,
    n integer DEFAULT nextval('myschema.counter'::regclass),
    CONSTRAINT users_name_check CHECK (myschema.non_empty(name))
   FROM myschema.users;

$ ./pista dump -c postgres://myschema@localhost/postgres -n myschema
    m mood,
    n integer DEFAULT nextval('counter'::regclass),
    CONSTRAINT users_name_check CHECK (non_empty(name))
   FROM users;
```

The plan half reproduces too, against the desired schema from `search_path_test.go`:

```
$ ./pista plan -c postgres://myschema@localhost/postgres -n myschema desired.sql
ALTER TABLE myschema.users ALTER COLUMN n SET DEFAULT nextval('myschema.counter'::regclass);
ALTER TABLE myschema.users DROP CONSTRAINT users_name_check;
ALTER TABLE myschema.users ADD CONSTRAINT users_name_check CHECK (myschema.non_empty(name));
```

And the dump taken as `myschema` is not reloadable by another role, which is the outcome the PR is trying to prevent:

```
$ psql -d reload_test -v ON_ERROR_STOP=1 -f dump-as-myschema.sql
ERROR:  type "mood" does not exist
LINE 3:     m mood,
```

`--search-path=public` makes all three cases behave, so `public` alone looks like the value the rationale actually asks for: `public` is there to keep an object in `public` unqualified, which the dump and the diff normalizations are written against, while `"$user"` buys nothing for a catalog read and is the only piece that varies by connecting role.

The cost of dropping it is that pre-SQL referring to an unqualified object in the connecting role's own schema would stop resolving. That seems much narrower than the case above, and pre-SQL already moves under this setting either way.

Whichever value you land on, the role-named-schema case is worth a test -- the current fixtures all drive the server-side path through `PGOPTIONS`, which is why this one is not caught.

---

README.md
```diff
@@ -383,6 +383,8 @@ echo $?  # 0: no changes, 2: changes, 1: error
 
 `plan` and `dump` open a read-only connection, so they cannot write to the database. Pass `--no-read-only` (env `$PISTA_NO_READ_ONLY`) to use a read-write connection.
 
+Every connection sets `search_path` to PostgreSQL's default, `"$user", public`. The catalog reports an object reachable through `search_path` without its schema, so a server-side `ALTER ROLE ... SET search_path` would otherwise decide how `dump` writes it. Pass `--search-path` (env `$PISTA_SEARCH_PATH`) to change it: `--search-path=myschema` makes `dump -n myschema` write `FROM users` instead of `FROM myschema.users`.
```
winebarrel: `--search-path` is on `Options`, so it reaches `plan` and `apply` as well, but the paragraph only describes what it does to `dump`. Pointing a target schema at it there produces exactly the drift the PR is fixing, since the catalog then reads unqualified while the desired schema qualifies:

```
$ ./pista plan -n myschema --search-path=myschema desired.sql
ALTER TABLE myschema.users ALTER COLUMN n SET DEFAULT nextval('myschema.counter'::regclass);
ALTER TABLE myschema.users DROP CONSTRAINT users_name_check;
ALTER TABLE myschema.users ADD CONSTRAINT users_name_check CHECK (myschema.non_empty(name));
```

A sentence saying the option is for shaping `dump` output, and that putting a target schema on the path makes a qualified desired schema plan on every run, would keep someone from finding this the slow way.

Separately, an empty `search_path` cannot be expressed: `connect` reads `""` as "unset" and falls back to `DefaultSearchPath`, so `--search-path=` gives the default rather than a path that qualifies everything. That is the one value that makes the catalog output fully self-describing, so it may be worth reaching with a `*string` or a sentinel.

---

winebarrel (commented):
Reviewed the change and reproduced its premise on a local PG 16.13 built from this branch (`6d112cf`).

The diagnosis holds and the mechanism is the right one. Sending `search_path` as a startup parameter puts it at `PGC_S_CLIENT`, and the backend applies `port->cmdline_options` before `port->guc_options`, so the value wins over both `ALTER ROLE ... SET` and `PGOPTIONS` regardless of how pgx orders the map -- the `PGOPTIONS` fixtures are testing a real ordering guarantee, not an accident. It is a GUC value rather than a `SET` statement, so there is no injection surface, and the comment's claim about invalid values checks out:

```
$ ./pista dump -n myschema --search-path='"unclosed'
pista: error: pistachio: failed to connect database: ... FATAL: invalid value for parameter "search_path": ""unclosed"
```

Reading the catalog before `apply`'s own `SET` is also correct, so the diff still compares under this setting while the DDL resolves against the target schemas.

One substantive thing: the chosen default still carries `"$user"`, so the role-named-schema case the description lists as part of the problem survives the fix -- same database, different dump depending on who connects, and a dump that another role cannot reload. Details and repro inline on `options.go`. Two smaller notes on the README paragraph.

---

options.go
winebarrel: Reproduced it here as well, with a role owning a same-named schema and no server-side setting: as `postgres` the dump comes out qualified, as `myschema` it comes out bare, and `--search-path=public` settles both.

Decision is to keep `"$user", public` as the default, so this stays a known limitation rather than something the PR fixes. The reasoning is the one you named as the cost: PostgreSQL's own default is what pre-SQL and anything else sharing the connection expect, and pre-SQL is the surface that would silently change. `--search-path=public` is the way out for anyone in that setup.

Followed the rest of the note:

- `TestDump_SchemaNamedAfterRole` covers it, both halves. It builds the schema from `current_user`, so it needs no extra role and reaches the `"$user"` entry in any environment, which is what the `PGOPTIONS` fixtures could not do.
- TODO.md records the open question of making `public` the default outright, with the repro and the cost.
- The PR description no longer counts this among the problems solved, and it has a section saying what it leaves.

54fbf7a

README.md
winebarrel: Both taken.

The paragraph now says the setting decides how `dump` writes an object and how `plan` compares it, and that putting a target schema on the path makes a desired schema that qualifies its own objects plan on every run.

`Options.SearchPath` is a `*string`, so `--search-path=` reaches an empty path instead of falling back; nil is what means the default now. Confirmed it qualifies everything, including `public`, which is the property you named:

```
$ pista dump -n public --search-path=
    m public.m1,
   FROM public.t1;
```

`TestOptions_SearchPathEmpty` pins that the flag arrives set-but-empty, and `TestDump_SearchPathEmpty` pins the output.

54fbf7a

README.md
```diff
@@ -383,6 +383,10 @@ echo $?  # 0: no changes, 2: changes, 1: error
 
 `plan` and `dump` open a read-only connection, so they cannot write to the database. Pass `--no-read-only` (env `$PISTA_NO_READ_ONLY`) to use a read-write connection.
 
+Every connection sets `search_path` to PostgreSQL's default, `"$user", public`. The catalog reports an object reachable through `search_path` without its schema, so a server-side `ALTER ROLE ... SET search_path` would otherwise decide how `dump` writes that object and how `plan` compares it.
+
+Pass `--search-path` (env `$PISTA_SEARCH_PATH`) to change it. `--search-path=` qualifies everything, so the dump reloads the same way wherever it runs. `--search-path=myschema` drops `myschema` from the output, and makes a desired schema that qualifies its own objects plan on every run. The default keeps `"$user"`, so a schema named after the connecting role is reported without its schema; `--search-path=public` covers that case.
```
winebarrel: "the dump reloads the same way wherever it runs" is true of `psql`, but the usual thing to do with a `dump` is commit it as the desired schema, and there the setting has to match on `plan` too. Taken with `--search-path=` and planned without it, the fully qualified dump drifts forever.

On `54fbf7a` against PG 16.13, a `public` schema with a `nextval` default, a CHECK calling a function, and a view calling one:

```
$ ./pista dump -c $C --search-path= > desired.sql
$ ./pista plan -c $C desired.sql
ALTER TABLE public.users ALTER COLUMN n SET DEFAULT nextval('public.counter'::regclass);
ALTER TABLE public.users DROP CONSTRAINT users_name_check;
ALTER TABLE public.users ADD CONSTRAINT users_name_check CHECK (public.non_empty(name));
CREATE OR REPLACE VIEW public.active_users AS
SELECT id, name FROM public.users WHERE public.non_empty(name);
```

Both settings are self-consistent -- dump and plan under the same value are clean either way, and the reverse mismatch drifts symmetrically -- so this is the option being used inconsistently rather than a break in the round trip. But `public` is the default target, so it is the easy mistake to make, and nothing in the output says which side is out of step.

The three statements are the qualification-normalization gap already filed in TODO.md: the CHECK and the view body are "Perpetual drift on a schema-qualified function call in a CHECK constraint" (`normalizeCheckExpr` never touches a `FuncCall` name, and `stripQualifications` clears `RangeVar.Schemaname` but not the call). The `SET DEFAULT` looks like a fourth instance, a qualified sequence in a default expression, which neither that entry nor the `serial` one covers -- worth a line in the TODO entry if you keep the flag as is.

A sentence here saying the value has to be the same for `dump` and for the `plan`/`apply` that reads the result would cover it.

---

winebarrel (commented):
Re-reviewed at `54fbf7a` on a local PG 16.13. Both points are answered, and the `"$user"` one is answered better than a fix would have been -- the behaviour is pinned by `TestDump_SchemaNamedAfterRole`, the way out is in the README, and TODO.md states the open question rather than burying it. Keeping the default at PostgreSQL's own value is a defensible call; I have no objection to it standing.

The pointer change reaches an empty path through all three input channels, which I checked separately since kong, env, and the YAML resolver each decide "set but empty" on their own:

```
--- flag --search-path= ---                    m public.mood,  FROM public.users;
--- env PISTA_SEARCH_PATH= (set but empty) --- m public.mood,  FROM public.users;
--- config: search-path: "" ---                m public.mood,  FROM public.users;
--- config: search-path: (null) ---            m mood,         FROM users;
```

The null case still falling through to the default is the right split. `go test -run 'SearchPath|SchemaNamedAfterRole' ./...` passes, and `apply --search-path=` runs and leaves no drift under either setting.

One thing left, inline on the README: a dump taken with `--search-path=` plans clean only when `plan` gets the same value, and `public` -- the default target -- is where the mismatch bites.

On the wording in TODO.md and the changelog, "what pre-SQL and anything else sharing the connection expect": `apply` overrides the path before its DDL and the catalog wants the opposite, so pre-SQL is the whole of "anything else". And for anyone who has a server-side `ALTER ROLE ... SET search_path` today, this PR already moves pre-SQL off it. The reasoning is narrower than it reads, though it does not change where you landed.

---

README.md
winebarrel: Reproduced it, including the fourth statement your paste shows for the view. Under the same value both directions are clean, so the round trip holds per setting and what breaks is only a mismatch between the two runs.

We are not adding anything for it. The rule is to pass the same `--search-path` to `dump` and to the `plan` / `apply` that reads the result, and that is the whole of it. A directive that carried the value in the dump was on the table and is dropped: it would put a knob in the schema file that changes how the database is read, which is a different kind of thing from every directive we have, and the mismatch it prevents is one the option's own value already decides.

Two notes on the analysis, which do not change the outcome.

The `SET DEFAULT` is a fourth instance of the missing normalization, but not one the entry's proposed fix reaches. `nextval('public.counter'::regclass)` carries the schema inside a string literal argument, not in the `FuncCall` name, so stripping the name symmetrically leaves it. Same for the view body, where `stripQualifications` clears `RangeVar.Schemaname` and never looks at the call. Worth writing into the entry as separate sites rather than as one more case.

On "what pre-SQL and anything else sharing the connection expect": you are right that pre-SQL is the whole of it. `apply` sets its own path before the DDL, and `plan` and `dump` run nothing else, so the reasoning covers exactly one surface. Fixing the wording in TODO.md, the changelog, and the `DefaultSearchPath` comment.

README.md
```diff
@@ -383,6 +383,13 @@ echo $?  # 0: no changes, 2: changes, 1: error
 
 `plan` and `dump` open a read-only connection, so they cannot write to the database. Pass `--no-read-only` (env `$PISTA_NO_READ_ONLY`) to use a read-write connection.
 
+Every connection sets `search_path` to PostgreSQL's default, `"$user", public`. The catalog reports an object reachable through `search_path` without its schema, so a server-side `ALTER ROLE ... SET search_path` would otherwise decide how `dump` writes that object and how `plan` compares it.
+
+Pass `--search-path` (env `$PISTA_SEARCH_PATH`) to change it. `--search-path=` qualifies everything, so the dump reloads the same way wherever it runs. `--search-path=myschema` drops `myschema` from the output. The default keeps `"$user"`, so a schema named after the connecting role is reported without its schema; `--search-path=public` covers that case.
```
winebarrel: Small thing the rewrite dropped. The previous text read "`--search-path=myschema` drops `myschema` from the output, and makes a desired schema that qualifies its own objects plan on every run"; the second clause is gone, and the new NOTE does not reach it, since it is scoped to a dump and the plan reading it.

A hand-written desired schema never goes through `dump`, and drifts on its own:

```
$ cat hand.sql   # written by hand, qualified throughout
CREATE TABLE myschema.users (
    n integer DEFAULT nextval('myschema.counter'::regclass),
    CONSTRAINT users_name_check CHECK (myschema.non_empty(name)) ...

$ ./pista plan -n myschema --search-path=myschema hand.sql
ALTER TABLE myschema.users ALTER COLUMN n SET DEFAULT nextval('myschema.counter'::regclass);
ALTER TABLE myschema.users DROP CONSTRAINT users_name_check;
ALTER TABLE myschema.users ADD CONSTRAINT users_name_check CHECK (myschema.non_empty(name));

$ ./pista plan -n myschema hand.sql      # default path, same file
(clean)
```

Putting a target schema on the path is the one setting that makes a qualified desired schema disagree with the catalog no matter where the file came from, so it is worth keeping the clause, or widening the NOTE past "the output of a dump" to any desired schema.

---

winebarrel (commented):
Re-reviewed at `16bb534`, rebased onto `6af3279`. All three points from the last round are answered, and the TODO entry is now more accurate than what I reported: the `SET DEFAULT` case is filed as a distinct node rather than a fourth `FuncCall`, since the schema sits inside the `'public.counter'::regclass` string literal and stripping a call name symmetrically would never reach it. The view-body half is folded into the CHECK entry, and the pre-SQL justification now says what it covers and what it does not.

Verified on the rebased branch against PG 16.13:

- `go build`, `go vet`, and `go test -p 1 ./...` all green.
- The NOTE's guidance holds: `dump --search-path=` paired with `plan --search-path=` is clean, and the same pairing is unchanged by the rebase.

One leftover inline: the rewrite dropped the clause about `--search-path=myschema` making a qualified desired schema plan on every run, and the new NOTE does not cover it, because that case needs no dump.

Nothing blocking from my side.

---

README.md
```diff
@@ -383,6 +383,10 @@ echo $?  # 0: no changes, 2: changes, 1: error
 
 `plan` and `dump` open a read-only connection, so they cannot write to the database. Pass `--no-read-only` (env `$PISTA_NO_READ_ONLY`) to use a read-write connection.
 
+Every connection sets `search_path` to PostgreSQL's default, `"$user", public`, so a server-side `ALTER ROLE ... SET search_path` does not reach it. `--search-path` (env `$PISTA_SEARCH_PATH`) sets another value.
+
+The catalog reports an object reachable through `search_path` without its schema. `dump` writes the object as the catalog reports it, and `plan` compares that form against the desired schema, so a desired schema that qualifies an object the catalog reports bare differs on every run. Under `--search-path=` nothing is reachable and every object keeps its schema. Under `--search-path=myschema` the objects in `myschema` lose theirs, and under the default so do those in `public` and in a schema named after the connecting role.
```
winebarrel: "Under `--search-path=` nothing is reachable and every object keeps its schema" promises a little more than it delivers: `pg_catalog` is searched implicitly and an empty `search_path` does not remove it, so a built-in keeps coming out bare.

The dump of a table using only built-ins, taken with `--search-path=`:

```
CREATE TABLE public.t (
    id integer NOT NULL,
    s text,
    created timestamp with time zone DEFAULT now(),
    CONSTRAINT t_s_check CHECK (lower(s) <> 'x'::text)
);
CREATE INDEX t_lower_s_idx ON public.t USING btree (lower(s));
```

No `pg_catalog.` anywhere in the output. Nothing goes wrong from it -- pistachio manages nothing in `pg_catalog`, and both sides render a built-in the same way, so it never becomes drift. It is only that a reader taking the sentence at face value would expect `pg_catalog.text` and `pg_catalog.now()`. "every object outside `pg_catalog` keeps its schema" would say it exactly.

The rest of the rewrite covers what I raised: "a desired schema that qualifies an object the catalog reports bare differs on every run", followed by which schemas lose their qualification under each value, reaches the hand-written case without needing a rule about dumps.

---

winebarrel (commented):
Re-reviewed at `01f5291`. Describing the mechanism instead of prescribing a rule is the better shape here -- the earlier NOTE only covered a desired schema that came from `dump`, and stating which schemas lose their qualification under each value covers a hand-written one as well, without the reader having to be told twice.

The commit is README-only, so the code is unchanged since `16bb534`, where `go build`, `go vet` and `go test -p 1 ./...` were green on PG 16.13.

One wording nit inline on `--search-path=` and `pg_catalog`. Nothing blocking.

---

PR #381, opened 2026-08-08
Default search_path to public alone

winebarrel:
The connection set `search_path` to PostgreSQL's own default, `"$user", public`, so a schema whose name matches the connecting role sat on the path. The catalog then reported its objects bare, and a `dump -n <role>` taken that way wrote `m mood` and `FROM users` where another role got `m myschema.mood` and `FROM myschema.users`. The role that runs migrations is often not the role the application connects as, so what the dump wrote and what the diff compared depended on who ran it.

The default is now `public` alone. `--search-path` restores `"$user"` for a pre-SQL that wants it, and `--search-path=` still qualifies everything.

`TestDump_SchemaNamedAfterRole` is inverted to pin the new behaviour: the default keeps the schema, and an explicit `--search-path='"$user", public'` is what reads it bare. The TODO entry that held this as an open question is removed, and the unreleased CHANGELOG entry for the `search_path` work is rewritten rather than added to.

## Test

`make vet`, `make test`, `make lint` (0 issues) and `make test-scenario` (11 passed) all pass.


winebarrel:
Reviewed at bba4bd9: checked out the branch and read `options.go`, `client.go`, `apply.go`, the tests and the docs. `go build` and `go vet` pass. This container has no PostgreSQL, so I could not run `make test` or `make test-scenario` and am relying on the report in the description for those.

The direction is right. Pinning `search_path` on the connection exists to keep the catalog read independent of who connects, and the `"$user"` entry left that half done. `public` alone lines the value up with the reason it is set, and the dump round trip now holds for a schema named after the connecting role too. Removing the TODO entry and inverting the test are both the right follow-through.

## The pre-SQL escape hatch points the wrong way

`options.go:22` offers `--search-path` as the way back to `"$user"` for a pre-SQL that wants it. That value is also what the catalog reads under, so taking it reopens exactly the drift this change closes: someone targeting a role-named schema who passes `--search-path='"$user", public'` for their pre-SQL gets the bare names back in the dump and in the diff.

The pre-SQL does not need the flag. `apply` reads the catalog at `apply.go:59` before running pre-SQL at `apply.go:135`, and `plan` only prepends pre-SQL to the output without executing it. So a pre-SQL that wants `"$user"` can open with `SET search_path TO "$user", public;`. It cannot affect a catalog read that already happened, and the `SET search_path` at `apply.go:174` overrides it before the DDL. Pointing the comment there instead would keep the next reader off the route that reopens the problem.

## The README help blocks are missing --search-path

`pista dump --help` prints `--search-path=public`, but the help output pasted into the README never listed the flag at all. That predates this PR (#378 added it), though the default it displays changes here, so this is a convenient moment to regenerate those blocks.

## Wording in README:386

"it would read a schema named after the connecting role without its schema" reads as a schema being read without a schema. What loses the qualification is the objects: "it would read the objects of a schema named after the connecting role without their schema", or similar.

## Two things that check out

`TestConnect_SearchPath` is stronger than it was. While the expected value matched PostgreSQL's own default, the assertion did not separate "we set the parameter" from "the server would have said that anyway". `public` differs from the default, so it now carries its weight.

Dropping the kong tag-parsing sentence from `config_test.go` costs no coverage. `TestOptions_SearchPathFromConfig` still carries a comma through the config file, and the second half of `TestDump_SchemaNamedAfterRole` still pushes `"$user", public` through pgx's RuntimeParams.

None of this blocks the merge. The first point is the one I would fix before merging.

---

winebarrel:
Re-reviewed at 14dd397. All three points are addressed. LGTM.

**The pre-SQL escape hatch.** Dropping the pointer and leaving the fact ("Pre-SQL runs under this setting.") is the right landing. I suggested that such a pre-SQL can set the path itself, but that is one option among several -- qualifying the reference, or splitting the pre-SQL, work too -- so the comment does not need to pick one. What mattered was not sending the reader back to the value this default moves away from, and that is gone.

**The README help blocks.** I diffed all four blocks (top level, `plan`, `apply`, `dump`) against the `--help` output of a build of this branch. They match byte for byte, including the narrower column in the top-level block. Taken from the real output rather than hand-written, as far as I can tell.

**README:386.** "the objects of a schema named after the connecting role without their schema" reads correctly now.

Verification on my side: `go build` and `go vet` pass. I could not run `golangci-lint` here (the installed binary is built with go1.25 and cannot load a config targeting 1.26), and this container has no PostgreSQL, so `make test` and `make test-scenario` are still unrun. Since 14dd397 touches only comments and the README, nothing that passed at bba4bd9 can have moved.

---

PR #383, opened 2026-08-08
Keep PARTITION BY on a sub-partitioned partition

winebarrel:
`model.Table.SQL` finished the `CREATE TABLE` right after the `PARTITION OF ... FOR VALUES` clause, so a partition that is itself partitioned lost its own partition key. `dump` wrote the middle level of a multi-level partitioned table as a leaf, and reloading that dump failed at the level below it, which still said `PARTITION OF` that table. That breaks the round trip the design policy holds as a bug.

The same code builds the `CREATE TABLE` a plan emits, so it is not only a dump problem: `apply` created such a partition unpartitioned, and the partition under it failed with `is not partitioned`.

`PartitionDef` is read from the catalog and parsed from the desired schema either way; it was only left out of the output. `PARTITION BY` goes before `TABLESPACE`, matching the `CREATE TABLE` grammar.

The TODO entry that recorded this is removed.

## Test

Two fixtures, both confirmed to fail without the change:

- `testdata/dump/partitioned_subpartition.yml` for the dump output of a three-level table.
- `testdata/apply/create_subpartitioned_table.yml` for the create path, with `verify_no_drift`.

Checked live on PostgreSQL 15 as well: a `dump --sort-by-deps` of a three-level table reloads into another schema without error, `pg_get_partkeydef` on the middle level still reads `LIST (c)`, and a plan against that dump says `-- No changes`.

`make vet`, `make test`, `make lint` (0 issues) and `make test-scenario` (11 passed) all pass.

https://claude.ai/code/session_01VBgVZETDfsX3hCfQNd1upa


winebarrel:
## Review

The fix reads correctly and is as small as the bug allows. `PARTITION BY` before `TABLESPACE` matches the `CREATE TABLE ... PARTITION OF` grammar, and the guard mirrors the one on the plain branch. Both fields it tests are set together on either side -- the parser derives them from `cs.Partspec`, and the catalog reads `relkind = 'p'` alongside `pg_get_partkeydef` -- so the condition cannot half-fire and emit a `PARTITION BY` with nothing after it.

I could not run the database-backed suites in this environment, so I exercised the emitter through the parser instead. Every partition shape produces valid SQL: a range bound, a `DEFAULT` bound, and a bound plus tablespace, the last of which comes out as `FOR VALUES ... / PARTITION BY LIST (kind) / TABLESPACE fast_ssd`. `go build ./...`, `go vet ./model/...` and `go test ./model/...` pass. The two fixtures are picked up by the globs in `dump_test.go` and `apply_test.go`, and multi-level partitioning predates the CI matrix, so leaving `min_pg` off is right.

Two notes, neither blocking:

The tablespace combination is only covered indirectly. A unit test beside `TestTable_SQL_partitionOf_withBound_tablespace` would pin the clause order without a database, which the two YAML fixtures cannot do on their own.

The `INHERITS` branch below needs no equivalent change. PostgreSQL rejects a partitioned table as an inheritance child, so a table on that branch never carries a partition key.

One thing the change leaves standing, outside its scope: `PartitionDef` and `PartitionBound` are still never compared in the diff, so altering the key or the bound of a table that already exists plans nothing at any level. That was true before this PR and needs a table rewrite rather than an `ALTER`, but since the TODO entry recording the neighbouring problem goes away here, the gap may be worth an entry of its own.

---

winebarrel:
Both notes are addressed in 27493cc.

The tablespace combination is now covered directly. `TestTable_SQL_partitionOf_subpartitioned` and `TestTable_SQL_partitionOf_subpartitioned_tablespace` assert the whole emitted string, so `PARTITION BY` before `TABLESPACE` is pinned without a database. That also fixes the `codecov/patch` failure: coverage is collected per package with no `-coverpkg`, so the YAML fixtures reaching `model.Table.SQL` from the root package counted for nothing and the patch scored 0%. `Table.SQL` is at 100% within the `model` package now.

Agreed on the `INHERITS` branch, no change there.

The partition key and bound gap is now a TODO entry of its own. Confirmed on 15.18: turning `PARTITION BY RANGE (r)` into `LIST (r)` and moving a partition's `FOR VALUES` bound in the same desired file plans `-- No changes`. `PartitionBound` is only tested against nil to pick a branch, and `PartitionDef` never reaches the diff at all. The entry notes that neither is an `ALTER` away: a partition key cannot be changed, and a bound needs `DETACH PARTITION` plus `ATTACH PARTITION ... FOR VALUES`, which fails when a row sits outside the new bound. Erroring at plan time may fit better than emitting either.

winebarrel:
## Re-review of 27493cc

Both notes are addressed, and the follow-up is better than what I asked for. The two model tests assert the whole emitted string rather than a substring, which is stronger than the `Contains` calls next to them, and the tablespace case pins the clause order rather than leaving it implied. Both pass here, as does `go vet ./...`. The coverage reasoning in the commit message holds: the fixtures reach `model.Table.SQL` from the root package, so with per-package collection and no `-coverpkg` the new lines were never going to count.

The TODO entry is accurate on the part that matters most, which is that neither drift is an `ALTER` away. A partition key has no `ALTER` at all, and a bound only moves through `DETACH` and `ATTACH`, which fails on a row outside the new bound. Erroring at plan time does look like the better end state.

Two small things on that entry, neither worth holding the PR for:

The same branch never compares `PartitionOf` either, so re-parenting a partition, or a table gaining or losing `PARTITION OF` while keeping its name, is silent in exactly the same way and belongs to the same entry.

The branch tests `PartitionOf` and `PartitionBound` against nil, not `PartitionBound` alone, so the first sentence undersells slightly what it looks at.

CI is green across the matrix, samples, keywords, ascii and CodeQL, with only the windows job still running when I checked. Looks good to merge.

---

winebarrel:
Both taken in b7dd2fd. The entry is now "Silent drift on the partition shape".

`PartitionOf` checks out. Verified on 15.18, all three plan `-- No changes`: re-parenting a partition from one parent to another, a plain table gaining `PARTITION OF`, and a partition losing it. They are listed alongside the key and the bound.

The wording is fixed too. The branch tests `PartitionOf` and `PartitionBound` against nil, not the bound alone, and the entry says so.

One thing the wider scope changes: the parent cases do not sit with the key on the fix side. A partition key has no `ALTER` at all, so it needs the table recreated and the data moved, while the parent and the bound both run through `DETACH PARTITION` and `ATTACH PARTITION ... FOR VALUES`. Detaching is cheap; attaching scans the table and fails on a row the bound does not cover. The entry splits them that way rather than saying none of it is reachable.

Documentation only, no code touched.

winebarrel:
## Re-review of b7dd2fd

Documentation only, and it closes both notes. The entry now names all four shapes, and splitting them by how they would be fixed is the right cut: the key has no `ALTER` and needs a rewrite, while the parent and the bound both go through `DETACH` and `ATTACH`, where the attach is the part that scans and can fail on a row the bound does not cover. The opening sentence matches what the branch actually tests.

I checked the two cases the entry adds beyond re-parenting against the code, and they behave as described. A plain table gaining `PARTITION OF` takes the partition branch, which skips the column diff and emits no `ATTACH`. A partition losing it takes the regular branch, where the columns the desired table spells out match the ones the partition already carries, so nothing comes out there either.

Nothing left from my side. The code is unchanged since 27493cc, which was green across the matrix; the run on this commit was still in flight when I looked.

---

PR #384, opened 2026-08-08
Diff the logged <-> unlogged transition

winebarrel:
`model.Table.Unlogged` was read from the catalog and parsed from the desired schema, and used to write `CREATE UNLOGGED TABLE`, but the diff never compared it. An `UNLOGGED` table with a plain `CREATE TABLE` in the desired schema planned `-- No changes`, and so did the other direction. It now emits `ALTER TABLE ... SET LOGGED` / `SET UNLOGGED`.

The statement goes before the rest of the table's, which puts it after the foreign-key drops and before the adds. A logged table may not reference an unlogged one, so a table turning unlogged needs the key pointing at it gone first. `SET LOGGED` / `SET UNLOGGED` is not on `isMergeableAction`'s whitelist, so `--bulk-alter` leaves it alone.

A partitioned table is skipped. PostgreSQL changes persistence by rewriting the table, and a partitioned one has nothing to rewrite, so `ALTER TABLE ... SET LOGGED` on it reports success and changes neither the parent nor the partitions under it. Emitting it there would replan forever. A partition holds its own storage and takes the transition normally. The TODO entry is rewritten to record what is left, which is that flipping a partitioned table's own persistence stays silently ignored; the value only decides the default for partitions created later.

## Test

Four plan fixtures and one apply fixture, all confirmed to fail without the change except the partitioned one, which pins the skip:

- `alter_table_set_logged.yml`, `alter_table_set_unlogged.yml`
- `alter_partition_child_set_logged.yml`
- `no_diff_unlogged_partitioned_table.yml`
- `alter_table_set_unlogged.yml` (apply, with `verify_no_drift`)

The existing `no_diff_unlogged_table.yml` covers the unchanged case.

Checked the partitioned no-op directly on 15.18 and 17.10: `ALTER TABLE` returns success and `relpersistence` stays `u`, on the parent and on a partition under it. Checked the foreign-key ordering live as well: a plan that drops a key and turns the referenced table unlogged applies in that order and replans clean.

`make vet`, `make test`, `make lint` (0 issues) and `make test-scenario` (11 passed) all pass on PG15. The plan, apply and dump suites also pass against PG17.


diff/policies.go
```diff
@@ -11,6 +11,23 @@ import (
 
 // diffRLS emits ALTER TABLE ... ENABLE/DISABLE/FORCE/NO FORCE ROW LEVEL
 // SECURITY statements for changes to the table-level RLS flags.
+// diffPersistence emits the logged <-> unlogged transition for a table that
```
winebarrel: The new function was inserted between `diffRLS`'s doc comment and `diffRLS` itself. As it stands, `diffPersistence` carries a leading sentence about RLS flags, and `diffRLS` is left with no doc comment at all.

Also, `diff/policies.go` holds the RLS/policy diffing. Persistence is neither, and `diffTable` in `diff/tables.go` is its only caller, so that file looks like the natural home. Moving it there fixes the comment collision at the same time.

---

diff/policies.go
```diff
@@ -11,6 +11,23 @@ import (
...
+// holds storage. A partitioned table is skipped: PostgreSQL changes
+// persistence by rewriting the table, and a partitioned one has nothing to
+// rewrite, so ALTER TABLE ... SET LOGGED reports success and leaves
+// relpersistence alone, on the parent and on the partitions under it alike.
+// Emitting it there would replan forever. A partition holds its own storage
+// and takes the transition normally.
+func diffPersistence(fqtn string, current, desired *model.Table) []string {
+	if desired.Partitioned || current.Unlogged == desired.Unlogged {
```
winebarrel: The guard reads `desired.Partitioned`, but the reason for the skip is a property of the relation that is already in the database: it has no storage to rewrite. The desired side only agrees with that when the partitioning matches.

`Partitioned` is never diffed (its only other use is `model.Table.SQL`), so a desired schema that drops `PARTITION BY` from an existing partitioned table is accepted in silence. If that table also flips persistence, this emits `ALTER TABLE ... SET LOGGED` against a relation that is still partitioned in the database, which is the exact no-op-and-replan-forever case the skip exists to avoid.

`current.Partitioned || desired.Partitioned` covers both sides and costs nothing.

---

diff/tables.go
```diff
@@ -129,6 +129,11 @@ func diffTable(current, desired *model.Table, dc DropChecker) (*tableDiffResult,
 	result := &tableDiffResult{}
 	fqtn := desired.FQTN()
 
+	// The persistence change goes first, so it lands after the FK drops and
+	// before the FK adds. A logged table may not reference an unlogged one,
+	// so a table turning unlogged needs the FK pointing at it gone first.
+	result.Stmts = append(result.Stmts, diffPersistence(fqtn, current, desired)...)
```
winebarrel: The ordering rationale covers the case where the key is dropped in the same plan, but not the case where the key stays and both tables flip together. That one looks like it lands in the wrong order every time, not by chance.

PostgreSQL's check in `ATPrepChangePersistence` is symmetric: it scans `pg_constraint` by `conrelid` and by `confrelid`, so the referenced side is checked too. Turning a table unlogged is rejected while any logged table still references it.

Statements from `TableDiffResult.Stmts` are sorted by `createPosMap`, and `toposort/schema.go:79` adds an FK edge from the referencing table to the referenced one, so the referenced table is emitted first. For

```sql
-- init
CREATE TABLE public.parents (id integer NOT NULL, CONSTRAINT parents_pkey PRIMARY KEY (id));
CREATE TABLE public.children (
    id integer NOT NULL,
    parent_id integer NOT NULL,
    CONSTRAINT children_pkey PRIMARY KEY (id),
    CONSTRAINT children_parent_id_fkey FOREIGN KEY (parent_id) REFERENCES public.parents(id)
);
-- desired: both UNLOGGED, FK unchanged
```

the plan should come out as

```sql
ALTER TABLE public.parents SET UNLOGGED;
ALTER TABLE public.children SET UNLOGGED;
```

and the first statement fails, because `children` is still logged at that point. The other direction is fine under the same order: `parents` goes logged first, then `children` follows a table that is already logged. So `SET LOGGED` wants create order and `SET UNLOGGED` wants the reverse.

I could not run a database in this session, so please confirm against a live server before deciding. If it reproduces, splitting the unlogged direction into a bucket ordered like the drops would fix it; recording it in `TODO.md` next to the partitioned-table entry would at least stop it from being silent.

---

testdata/plan/alter_partition_child_set_logged.yml
winebarrel, line 1: A plan fixture pins the text of the statement, not that PostgreSQL accepts it. Of the four new cases this is the one whose DDL is least obviously valid, since it turns a partition logged underneath a parent that stays unlogged. The apply fixture went to the plain-table case, where the statement is not in doubt. An apply fixture with `verify_no_drift` here would buy more than that one does, and it would pin the claim across the whole CI matrix rather than the 15 and 17 checked by hand.

Two more gaps worth a fixture, both cheap:

- The FK-drop ordering is the stated reason the statement is emitted first, and nothing in the suite holds it in place. A plan fixture that drops the key and turns the referenced table unlogged would pin the two lines in order.
- `--bulk-alter` leaving the statement alone is asserted in the description but not tested. `SET LOGGED` next to an `ADD COLUMN` on the same table would cover it.

---

winebarrel (commented):
Reviewed the change. The core of it is right: `Unlogged` was parsed, read from the catalog and written on create but never compared, and comparing it is a two-line function. Emitting nothing was the wrong answer.

I read the ordering claims against the code and they hold for what they cover. `FKDropStmts` are written before `Stmts` in `diff_all.go`, so the statement does land after the drops and before the adds, and `isMergeableAction` really does have no `SET LOGGED` branch, so `--bulk-alter` passes it through. The partitioned-table skip matches how `ATRewriteTables` treats a relation with no storage, and pinning it with a fixture rather than trusting the reasoning is the right call.

Four things, in descending order of how much they matter:

1. The unlogged direction across two FK-linked tables that both flip while keeping the key looks like it comes out in the order PostgreSQL rejects, deterministically. Details inline on `diff/tables.go`. This is the one I would want checked on a live server before merging; I could not run one here.
2. The new function was inserted between `diffRLS`'s doc comment and `diffRLS`, so one function has a comment about the other and one has none.
3. The partitioned guard tests the desired side, when the property that matters belongs to the relation already in the database.
4. Test gaps: the partition-child case is the one that most needs an apply fixture and did not get one, and neither the FK ordering nor the `--bulk-alter` pass-through is pinned.

The `TODO.md` rewrite is a good trade. It closes the entry that this PR actually resolves and leaves behind the narrower thing that is still true, with the verification recorded.

---

winebarrel:
All four are addressed in d14248e.

**1. FK ordering.** Confirmed on 15.18 and 17.10, and it was worse than the review guessed: the plan followed neither order, it followed the desired file. Written the natural way, with the referenced table first, `pista apply` died on its first statement with `could not change table "parent" to unlogged because it references logged table "child"`.

`DiffTables` now collects the transitions and sorts each direction: turning unlogged waits for the tables that reference it, turning logged waits for the tables it references. Building that on top of `Stmts` does not work, which is worth recording. `orderStatements` tags every statement in `Stmts` with the object's topological position and sorts, so the logged direction came out right by accident and the unlogged direction was silently reordered back to the failing order. The transitions travel in their own `TableDiffResult.PersistenceStmts` and are appended after the creates, which keeps them after any rename, and before the FK adds.

The edges come from the desired foreign keys, which are the ones in force at that point: the drops have already run and the adds have not. Mutual keys have no valid order at all, so those tables keep the order they came in.

**2. Doc comment.** `diffPersistence` moved to `diff/persistence.go`, which is where the ordering belongs anyway, and `diffRLS` has its comment back.

**3. Partitioned guard.** Now reads `current.Partitioned`. Agreed on the reasoning: what decides whether the statement does anything is the relation already in the database.

**4. Test gaps.** Apply fixtures for the partition child and for the unlogged FK ordering, both with `verify_no_drift`; plan fixtures for both FK directions, each written in the order that fails without the sort; unit tests in `diff/persistence_test.go` for the mutual-FK cycle, an unqualified reference and both directions in one plan; and `SET LOGGED` / `SET UNLOGGED` added to the bulk-alter negatives.

**CI.** `test (18)` was failing because PostgreSQL 18 rejects an unlogged partitioned table outright, so `no_diff_unlogged_partitioned_table` could not even load its `init`. The plan harness takes a `max_pg` field now, the mirror of `min_pg`, and the fixture is capped at 17. The TODO entry records that the shape cannot arise on 18.

Full suite run against 15, 17 and 18, plus `make lint` (0 issues) and `make test-scenario` (11 passed).

diff/persistence.go
```diff
@@ -0,0 +1,125 @@
...
+	var visit func(int)
+	visit = func(i int) {
+		if state[i] != unvisited {
+			return
+		}
+		state[i] = open
+		for _, j := range refs[i] {
+			visit(j)
+		}
+		state[i] = done
```
winebarrel: `open` and `done` are never told apart. `visit` returns on `state[i] != unvisited`, which both satisfy, so the write on line 114 is never read and the three-state enum carries no information a `visited []bool` would not.

The usual reason to separate them is cycle detection: hitting a node still `open` is a back edge. Here a cycle is deliberately not an error, so there is nothing to detect, and the plain form says the same thing with fewer moving parts.

---

diff/persistence.go
```diff
@@ -0,0 +1,125 @@
...
+// reference it, and a table turning logged waits for the tables it references.
+// The two directions therefore run in opposite orders, and the tables turning
+// logged go first; a cross-direction pair either has no constraint between it
+// or describes a desired schema PostgreSQL rejects whichever order this picks.
+//
+// Only the tables that change are ordered. One that keeps its persistence is
+// already on the right side of the rule, or the desired schema breaks it
+// regardless. The desired foreign keys are what the order is built from, since
+// those are the keys in force here: the drops have already run and the adds
+// come later.
```
winebarrel: The desired keys are not quite the keys in force. The drops have run, so what exists at this point is `current` intersected with `desired`; a key that appears only in `desired` is added later and is not there yet. So the desired set is a superset, and every extra edge only tightens an order that was already valid, which is why this is safe rather than wrong.

The one place the superset bites is a cycle: a key added in the same plan can close a loop that the in-force keys do not have, and the tables in a cycle keep their input order. That is narrow enough to leave alone, but the sentence claims exactness the set does not have. Saying it is a superset, and that the extra edges only constrain further, would describe what the code does.

---

diff/persistence.go
```diff
@@ -0,0 +1,125 @@
...
+	for i, c := range changes {
+		for _, fk := range c.table.ForeignKeys.CollectValues() {
+			if fk.RefTable == nil {
+				continue
+			}
+			schema := c.table.Schema
+			if fk.RefSchema != nil {
+				schema = *fk.RefSchema
+			}
+			if j, ok := index[model.Ident(schema, *fk.RefTable)]; ok && j != i {
```
winebarrel: `toposort/schema.go` orders tables by their foreign keys too, and it resolves an unqualified reference differently: `resolveUnqualified` tries the owning schema and then falls back to `public`. This tries the owning schema alone, so `CREATE TABLE app.child (... REFERENCES parent(id))` against a `public.parent` looks up `app.parent`, misses, and the edge is dropped. If both tables flip in the same plan, the order goes unconstrained.

`normalizeFKSchema` is a reasonable thing to follow, but it answers a different question: it is comparing two FK definitions for equality, where filling both sides the same way is all that matters. Here the schema has to name the table that actually gets altered, and since #381 put `public` on the path by default, that is where the bare name lands.

Narrow enough to just record. Either match `resolveUnqualified` (the fallback is five lines, though it is unexported in `toposort`), or keep the comment but say the fallback is missing rather than citing `normalizeFKSchema` as if the two questions were the same.

---

testdata/plan/alter_table_set_logged_fk_order.yml
winebarrel, line 1: The unlogged direction got a matching apply fixture and this one did not, so the order here is pinned as text but never handed to a server. That is the asymmetry I raised last round, one level down.

It matters a little more than the usual "add the apply too", because the two directions are checked by different legs of `ATPrepChangePersistence`, and only the unlogged leg is exercised end to end right now. A copy of `testdata/apply/alter_table_set_unlogged_fk_order.yml` with the persistence flipped would close it.

---

winebarrel (commented):
Re-reviewed at d14248e. All four points from the last round are addressed, and the ordering one is addressed properly rather than papered over.

Notes on the fixes:

- The persistence changes moving into their own `TableDiffResult` field is the right shape. Leaving them in `Stmts` and trying to influence the sort could not have worked, since one direction wants create order and the other wants its reverse, and `createPosMap` has only one of those.
- Placing the phase between the creates and the FK adds holds up on the other paths too. Foreign keys on dropped tables already go to `FKDropStmts`, so they are gone by the time the phase runs, and `mergeAlterTable` only ever sees `Stmts`, so `--bulk-alter` cannot reach these statements at all now. The negatives added to `bulk_alter_test.go` are belt and braces rather than the actual guard, which is fine.
- `current.Partitioned` is better than the `current || desired` I suggested. Adding `PARTITION BY` to an existing plain table leaves a relation that still has storage, so the statement does work there and skipping it would have been wrong.
- The apply fixture for the unlogged direction settles a question I could not answer by reading: `ALTER TABLE public.child SET UNLOGGED` runs while `public.parent` is still logged and is accepted, so the check really is one-directional and only looks at the tables referencing the target. That is what makes "a table that keeps its persistence is already on the right side of the rule" true, so the fixture is load-bearing for the doc comment as well as for the order.
- `max_pg` is a fair way to handle 18, and `AGENTS.md` records that it is plan-only.

What is left is small and none of it is a correctness problem:

1. The `open` / `done` states in `sortByRefs` are never distinguished.
2. The doc comment calls the desired foreign keys the keys in force; they are a superset of them.
3. `sortByRefs` resolves an unqualified reference to the owning schema only, where `toposort` falls back to `public`.
4. The logged direction is pinned by a plan fixture but not an apply one.

`go vet ./diff/...` and `go test ./diff/...` pass here; I have no server in this session, so the fixture suites are on CI.

---

winebarrel:
All four in 475b128.

**1. `open` / `done`.** Correct, they were never told apart, and the reason to separate them does not apply here since a cycle is deliberately not an error. It is a `visited []bool` now.

**2. Superset.** Reworded. The comment says the desired keys are a superset of the ones in force, that the drops have run so what exists is the current keys intersected with the desired, that every extra edge tightens an order that was valid without it, and that the one place it bites is a key added in the same plan closing a cycle the in-force keys do not have.

**3. Unqualified reference.** Implemented rather than documented. `resolveRef` now tries the owning schema and falls back to `public`, matching `toposort.resolveUnqualified`. The point about `normalizeFKSchema` answering a different question is right and the comment no longer cites it: filling both sides of an equality check the same way is all that matters there, while here the schema has to name the table that actually gets altered.

**4. Logged apply fixture.** Added as `testdata/apply/alter_table_set_logged_fk_order.yml`, the mirror of the unlogged one, so both legs of `ATPrepChangePersistence` are exercised end to end.

**Coverage.** Checking the patch turned up the same per-package problem as #383, one file over. The `diff` package never drove `DiffTables` through a persistence change, so the line that collects them counted as uncovered even though the fixtures exercise it from the root package. `TestDiffTables_PersistenceStmts` covers it and pins that the statements stay out of `Stmts`. With that and a case for a nil `RefTable` and a reference outside the set, `persistence.go` is at 100% and every line this PR touches in `diff/tables.go` is covered; what is left uncovered there is pre-existing `if err != nil` returns.

Full suite on 15 and 18, `make lint` (0 issues) and `make test-scenario` (11 passed).

winebarrel (commented):
Re-reviewed at 475b128. Verified with `make lint` (0 issues), the full suite on PostgreSQL 15 and 18, the persistence fixtures on 17, and coverage: all four functions in `diff/persistence.go` are at 100%, the `diff` package at 96.0%.

Nothing here is a correctness problem. I walked every consumer of `TableDiffResult` and none drops `PersistenceStmts` (`orderStatements`, `fallbackOrder`, the bulk-alter merge, disallowed drops, `ObjectCount`). The claim that the transitions land after any rename holds: a table `RENAME TO` goes into `Stmts`, which becomes `createStmts`, which is emitted before them. Splitting them out of `Stmts` also makes the bulk-alter hazard structural rather than a matter of the negatives list, since `mergeAlterTable` only ever sees `Stmts`. Under a schema map the edges still resolve, because `remap.go` rewrites `fk.RefSchema` alongside the table schemas. The `visited []bool` simplification is right: a cycle is deliberately not an error, so open and done carry nothing.

One thing beyond the inline notes. `dump` writes a foreign key unqualified, as `REFERENCES parent(id)` -- the `applied` block of `testdata/apply/alter_table_set_logged_fk_order.yml` shows exactly that. So the unqualified form is what a round trip hands back, yet all four plan fixtures write `REFERENCES public.parent(id)`, and only the unit tests reach `resolveRef`'s unqualified path. Given that the round trip is what the design policy is built around, one plan fixture with a bare reference would be cheap and would pin the path the tool's own output takes.

diff/persistence_test.go
```diff
@@ -0,0 +1,152 @@
...
+	stmts := orderPersistenceChanges([]*persistenceChange{changeFor(a), changeFor(b)})
+
+	assert.ElementsMatch(t, []string{
+		"ALTER TABLE public.a SET UNLOGGED;",
+		"ALTER TABLE public.b SET UNLOGGED;",
+	}, stmts)
+}
+
+// A reference written without a schema is read as the owning table's, the way
+// normalizeFKSchema reads it, so the edge is still found.
```
winebarrel: This still names `normalizeFKSchema`, which the commit message of 475b128 says was the wrong rule to follow. `resolveRef` and the test right below it both name `toposort.resolveUnqualified` instead.

plan_test.go
```diff
@@ -22,7 +22,11 @@ type planTestCase struct {
 	// MinPG skips the fixture on a server older than this major version, for
 	// syntax that does not exist there. PostgreSQL 16 added the SQL/JSON
 	// constructors, 17 the query functions.
-	MinPG                    int             `yaml:"min_pg,omitempty"`
+	MinPG int `yaml:"min_pg,omitempty"`
+	// MaxPG skips the fixture on a server newer than this major version, for
+	// syntax a later one dropped. PostgreSQL 18 rejects an unlogged
+	// partitioned table.
```
winebarrel, line 28: "syntax a later one dropped" does not describe what happens here. PostgreSQL 18 did not drop syntax; it rejects an unlogged partitioned table, which is a semantic restriction. The CHANGELOG already puts it as "a server newer than the shape it pins", which covers both and matches `min_pg`'s wording.

diff/persistence.go
```diff
@@ -0,0 +1,139 @@
...
+	return &persistenceChange{table: desired, stmt: stmt}
+}
+
+// orderPersistenceChanges puts the transitions in an order PostgreSQL accepts.
+// It refuses to leave a logged table referencing an unlogged one and checks it
+// on every ALTER, so a table turning unlogged waits for the tables that
+// reference it, and a table turning logged waits for the tables it references.
+// The two directions therefore run in opposite orders, and the tables turning
+// logged go first; a cross-direction pair either has no constraint between it
+// or describes a desired schema PostgreSQL rejects whichever order this picks.
```
winebarrel: The conclusion is right, but half of the case is stronger than "a desired schema PostgreSQL rejects". Take X turning unlogged and Y turning logged with X referencing Y: the current state would be a logged X referencing an unlogged Y, which PostgreSQL will not let you build in the first place, so that orientation cannot reach this function at all. Only the mirror -- X turning logged and Y turning unlogged -- is a reachable state with a desired schema that fails whichever order is picked.

winebarrel:
All four in a2fbd23, and the fixture request turned up something.

**Bare reference.** Adding it showed that the `public` fallback I put in last round is for a case that cannot happen. The fixture passes with the unqualified branch deleted: `parseFKConstraint` fills `RefSchema` from the owning table's schema when `Pktable.Schemaname` is empty, and `catalog.ListConstraintsByTable` reads it from `pg_namespace`, so a bare name never arrives here bare. `resolveRef` is gone, the lookup is direct again, and the comment records why `toposort.resolveUnqualified` has no counterpart rather than imitating it. The two unit tests that reached the branch built a nil `RefSchema` by hand and went with it.

The fixture stays, on the logged direction, and it is load-bearing. Checked by dropping every edge: `alter_table_set_logged_fk_order` and its bare-ref twin fail, the unlogged ones pass either way. That is worth recording, because it means my unlogged fixtures never pinned the edges at all: with the desired file listing the referenced table first, reversing the input order lands on the right answer whether or not an edge was found. They pin the reversal and, in the apply case, that the server accepts the result. The asymmetry is now written on the fixture.

**Cross-direction paragraph.** Rewritten along the lines you set out. With X referencing Y, X turning unlogged next to Y turning logged would need a logged X already referencing an unlogged Y, which is the state the rule forbids, so it never reaches the function. Only the mirror is reachable, and it asks for a schema PostgreSQL rejects whichever order is picked.

**`normalizeFKSchema` in the test comment.** Gone.

**`MaxPG` wording.** Now "a shape a later one no longer allows", since 18 dropped no syntax.

`persistence.go` is still at 100% after the deletions. Full suite on 15 and 18, `make lint` (0 issues), `make test-scenario` (11 passed).

One note on the earlier run: a batch of unrelated failures I saw locally was my test run colliding with another session on the same database, not the change. Re-run clean on a fresh server.

diff/persistence.go
```diff
@@ -0,0 +1,128 @@
...
+		index[c.table.FQTN()] = i
+	}
+	refs := make([][]int, len(changes))
+	for i, c := range changes {
+		for _, fk := range c.table.ForeignKeys.CollectValues() {
+			// A reference written without a schema does not arrive here
+			// without one: the parser fills RefSchema from the owning table's
+			// schema, and the catalog reads it from pg_namespace. So there is
+			// no bare name to resolve, and nothing like
+			// toposort.resolveUnqualified is needed.
```
winebarrel: The parser does not fill `RefSchema` from the owning table's schema. `parser/parser.go:1526` and `parser/parser.go:1587` both fall back to `defaultSchema`, which `diff_all.go:83` passes as `client.Schemas[0]`, the first target schema. The two coincide only when the owning table happens to sit in that schema.

They come apart on `--schema public,app` with

```sql
CREATE TABLE app.child (
    ...
    CONSTRAINT child_p_fkey FOREIGN KEY (p) REFERENCES parent(id)
);
```

where this comment predicts `app.parent` and the parser produces `public.parent`.

The conclusion survives -- `RefSchema` is never nil, so there is still no bare name to resolve -- and what the parser actually does is the more useful rule, since resolving to the first target schema is what `search_path` does and it finds the edge to `public.parent` that the owning-table rule would miss. But this comment is the entire justification for dropping `resolveRef`, so it should describe the rule the code implements. The commit message says the same thing. Worth noting too that the new fixture is all in `public`, so it cannot tell the two rules apart; nothing pins the cross-schema case either way.

diff/persistence.go
```diff
@@ -0,0 +1,128 @@
...
+			if fk.RefSchema == nil || fk.RefTable == nil {
```
winebarrel, line 105: Nit, and it follows from the comment above. If a bare name really never arrives, then `RefSchema == nil` is unreachable: the parser sets it whenever `Pktable` is non-nil, which it always is for a foreign key, and the catalog reads it from `pg_namespace`. The guard is then kept alive only by a hand-built `model.ForeignKey` in `TestOrderPersistenceChanges_ReferencesOutsideTheSet`, which is the shape AGENTS.md says not to write tests for. Either the guard is load-bearing, in which case "does not arrive here without one" is too strong, or it is not and the comment can say it is there as a cheap nil check rather than as resolution. Not worth changing the code over.

winebarrel (commented):
Re-reviewed at a2fbd23. All four points from the last round are addressed: the `normalizeFKSchema` citation went away with the test, `MaxPG` now reads "a shape a later one no longer allows", the cross-direction paragraph is rewritten around the unreachable orientation, and the bare-reference fixture exists.

Deleting `resolveRef` rather than keeping it is the better answer, and the new fixture earns its place. It is load-bearing for the reason the header claims: the desired file lists `child` first and the expected plan puts `parent` first, so with no edge found `sortByRefs` would return the input order and the fixture would fail. That also makes it a live check that `RefSchema` arrives non-nil, since the `nil` branch would drop the edge.

Verified with `make lint` (0 issues) and the full suite on PostgreSQL 15 and 18. Coverage is unchanged at 97.5% for the root package and 96.0% for `diff`, and all three functions in `persistence.go` stay at 100%. I could not check 17 this round -- the container is not running locally (compose has only pg15 and pg18 up), so unlike the previous rounds nothing here was exercised against it.

One inaccuracy in the comment that carries the whole argument, plus a nit, inline.

winebarrel:
Both in b1c9f7d.

**The rule the comment states.** You are right, and the commit message that introduced it was wrong the same way. `parseCreateStmt` and `parseAlterTableStmt` fall back to `defaultSchema`, and `diff_all.go:83` passes `client.Schemas[0]`. Your example is the case: `--schemas public,app` with `CREATE TABLE app.child (... REFERENCES parent(id))` gives `public.parent`, not `app.parent`.

The comment now describes that rule and says why it is the better one to rely on: resolving to the first target schema is where `search_path` lands a bare name, and it finds the `public.parent` edge the owning-table rule would miss. The conclusion for `resolveRef` is unchanged, since `RefSchema` is non-nil under either rule.

On the cross-schema case being unpinned: the plan harness runs a single schema, so a fixture cannot reach it. `TestOrderPersistenceChanges_CrossSchemaReference` covers it at the level this code works at, an `app.child` whose key carries `public.parent`, which is the shape the parser hands over.

**The nil guard.** Following it through turned up a real flaw in the test rather than just a wording question. `TestOrderPersistenceChanges_ReferencesOutsideTheSet` built its stray key with a `RefTable` but no `RefSchema`, so it hit the nil guard and never reached the index at all: it was not testing the miss it was named for. The key is qualified now, and the hand-built nil key is gone, since AGENTS.md says not to write tests for unreachable defensive code. The guard stays as a cheap nil check, the comment says so rather than claiming resolution, and it is the one uncovered line in `persistence.go` now. That seems like the honest state rather than propping it up.

**PostgreSQL 17.** Started this project's `pg17` and ran the full suite there, so this round is covered on 15, 17 and 18. `make lint` at 0 issues.

winebarrel (commented):
Re-reviewed at b1c9f7d. Both points from the last round are settled. The comment now names the rule the code implements, `client.Schemas[0]` rather than the owning table's schema, and the nil test is described as a guard rather than as resolution, with the synthetic key gone and the resulting gap left visible instead of papered over. Catching that the old outside-the-set test was exercising the guard rather than the miss it was named for is a good find, and pointing it at `public.elsewhere` fixes it.

Verified with `make lint` (0 issues) and the full suite on PostgreSQL 15, 17 and 18 -- all three this round, since pg17 is up again. Coverage holds at 97.5% for the root package and 96.0% for `diff`. `diffPersistence` and `orderPersistenceChanges` are at 100%, `sortByRefs` at 96.0%, and the single uncovered block is `persistence.go:106`, the guard's `continue`, exactly as the commit says.

One finding, inline.

diff/persistence_test.go
```diff
@@ -0,0 +1,126 @@
...
+	fk.Name = "child_p_fkey"
+	child.ForeignKeys.Set(fk.Name, fk)
+
+	stmts := orderPersistenceChanges([]*persistenceChange{changeFor(parent), changeFor(child)})
+
+	assert.Equal(t, []string{
+		"ALTER TABLE app.child SET UNLOGGED;",
+		"ALTER TABLE public.parent SET UNLOGGED;",
+	}, stmts)
+}
```
winebarrel, line 102: This does not pin the lookup it is named for. I measured it in a throwaway worktree by pointing the reference at a schema that is not in the set, so no edge can be found:

```
edge found:   [ALTER TABLE app.child SET UNLOGGED;  ALTER TABLE public.parent SET UNLOGGED;]
edge missing: [ALTER TABLE app.child SET UNLOGGED;  ALTER TABLE public.parent SET UNLOGGED;]
```

Identical, so the assertion holds even if cross-schema resolution were removed entirely. The cause is the asymmetry the header of `alter_table_set_logged_fk_order_bare_ref.yml` already writes down: the unlogged group is reversed, so an input that lists the referenced table first arrives at the right answer with no edge at all.

The logged direction separates them:

```
edge found:   [ALTER TABLE public.parent SET LOGGED;  ALTER TABLE app.child SET LOGGED;]
edge missing: [ALTER TABLE app.child SET LOGGED;      ALTER TABLE public.parent SET LOGGED;]
```

So building both tables logged and passing `child` first makes it load-bearing, which is the same reason the bare-reference fixture chose that direction.

While here: `TestOrderPersistenceChanges_BothDirections` has the same property in its "each group in its own order" half. The grouping half of that assertion is real; the intra-group order of the unlogged pair would hold with no edge either. Lower stakes, since grouping is what that test is for.

winebarrel:
Fixed in 923a57c, and your measurement matches mine.

`TestOrderPersistenceChanges_CrossSchemaReference` was the same trap I had just written the header of `alter_table_set_logged_fk_order_bare_ref.yml` about, and I walked into it one file over. It moves to the logged direction with the referencing table first, for the reason the fixture chose that direction.

`TestOrderPersistenceChanges_BothDirections` too. Rather than leave the intra-group half decorative, swapping the input so the unlogged pair arrives referencing-table-first makes it turn on the edge, since the reversal alone would then put it the wrong way round. The expected output is unchanged, and the grouping half it was written for still holds.

Checked the same way you did, by dropping every edge:

```
before: both pass
after:  both fail
```

`TestOrderPersistenceChanges_MutualForeignKeys` and `_ReferenceOutsideTheSet` still pass with no edges, which is what they are for: they assert what happens when no usable edge exists.

Full suite on 15, 17 and 18, `make lint` at 0 issues.

Worth noting the general shape, since it caught me twice: in this ordering, any unlogged-direction assertion whose input already lists the referenced table first is satisfied by the reversal alone and pins nothing about the edges. The logged direction is where an edge has to be found.

winebarrel (commented):
Re-reviewed at 923a57c. Nothing to raise this round.

Both tests now turn on the edge, and taking `BothDirections` along rather than leaving it as the lower-stakes half was the right call. Checked it by mutation rather than by reading: I disabled the edge lookup in `sortByRefs` in a throwaway worktree and ran everything that touches the ordering.

Fails with no edges, so it pins the ordering:

- `TestOrderPersistenceChanges_BothDirections`
- `TestOrderPersistenceChanges_CrossSchemaReference`
- `TestPlan/alter_table_set_logged_fk_order`
- `TestPlan/alter_table_set_logged_fk_order_bare_ref`
- `TestApply/alter_table_set_logged_fk_order`

Still passes with no edges, by design:

- `TestOrderPersistenceChanges_MutualForeignKeys` and `TestOrderPersistenceChanges_ReferenceOutsideTheSet`, which are about the absence of a constraint
- `TestPlan/alter_table_set_unlogged_fk_order` and `TestApply/alter_table_set_unlogged_fk_order`, for the reason their own headers give: the reversal alone lands on the right order there. The apply one still earns its keep, since what it checks is that a real server accepts the unlogged leg of `ATPrepChangePersistence`, not that the sort did anything.

So the logged direction carries the pinning at every level, unit and fixture alike, and that is now written down in both places.

Also verified with `make lint` (0 issues) and the full suite on PostgreSQL 15, 17 and 18. Coverage is unchanged at 97.5% for the root package and 96.0% for `diff`, with `sortByRefs` at 96.0% and the guard's `continue` at `persistence.go:106` still the only gap.

PR #394, opened 2026-08-20
Shorten an auto-generated constraint name the way PostgreSQL does

winebarrel:
`autoNameConstraint` built `{table}_{col}..._{label}` without a length limit, so a name over 63 bytes was handed to the server in full and truncated there. PostgreSQL truncates the identifier from the right, which cuts into the label: a `CHECK` on a 59-character table name went in as `{table}_quantity_check` and came back as `{table_48}_quantit`. The next plan no longer found the name it had written, so it added the constraint again on every run and the second apply failed with `constraint "..." already exists`.

```
$ pista apply schema.sql   # first run succeeds
$ pista plan schema.sql
ALTER TABLE public.customer_... ADD CONSTRAINT customer_..._quantity_check CHECK (quantity > 0);
-- skipped: ALTER TABLE public.customer_... DROP CONSTRAINT customer_..._quantit;
```

PostgreSQL builds the name with `makeObjectName` (`src/backend/catalog/catalog.c`), which shortens the table and column parts - the longer one first - until the whole name fits, and never touches the label. This ports that, clipping each part on a character boundary the way `pg_mbcliplen` does so a multibyte name is not cut in half, and names every constraint kind through it. A domain's unnamed `CHECK` took the same shortcut and goes through it too.

Duplicate names are still not resolved: PostgreSQL appends a number to the label and rebuilds the name from there (`t_a_check1`), which the desired schema alone cannot predict. That limit is unchanged.

## Tests

- `parser`: a table of 12 auto-naming cases covering name2 present/absent, shortening the table part, the column part, or both, the 63-byte boundary with nothing to shorten, a cut landing on a character boundary and one landing inside a character, and a four-column `UNIQUE`; plus separate cases for a foreign key, a domain constraint, and a `NULL` column constraint, which is a column attribute rather than something pg_constraint records.
- `testdata/apply/long_auto_constraint_name.yml`: adds an unnamed `CHECK` to a 59-character table alongside an unnamed `PRIMARY KEY`, with `verify_no_drift: true` so the next plan has to come back empty. It fails without the fix.

Every expected name was taken from a running PostgreSQL 16, not derived from the implementation.

Coverage of the `parser` package goes from 90.3% to 90.5%; `makeObjectName`, `clipIdent` and `autoNameConstraint` are each at 100%.

## Verification

Against a local PostgreSQL 16: `go vet ./...`, `go test -p 1 ./...` and `test/scenario/run.sh` all pass, and `golangci-lint run` reports no issues.

The last commit reformats `diff/policies_test.go`, which this change does not otherwise touch. The CI lint step installs the latest golangci-lint, and gofumpt 2.13.0 wants those two long one-line helpers out of the aligned group, so the lint step failed before reaching the tests. The reformatting is the formatter's own output.

## Background

Found while checking whether anything in [sqldef v3.11.20](https://github.com/sqldef/sqldef/releases/tag/v3.11.20) applies here. Of the six changes there, only sqldef/sqldef#1324 (unnamed `CHECK` constraints) overlaps: owner filtering, `CREATE OR REPLACE FUNCTION` and the mysqldef fix have no counterpart, `ENUM` renames are already covered by `-- pista:renamed-from`, and `NULLS NOT DISTINCT` round-trips cleanly since constraint definitions come from the catalog as text.


winebarrel:
## Review

I checked the port line by line against `makeObjectName` and `pg_mbcliplen` in the PostgreSQL tree, recomputed every expected name in the test table by hand, and ran the parser tests on the branch. The port is faithful and the fix is right.

Details I verified:

* The overhead computation matches: `strlen(label) + 1`, plus one more for the separating underscore when `name2` is non-NULL. Treating an empty `name2` as PostgreSQL's NULL is safe, because the only callers that pass an empty column part are the ones where PostgreSQL passes NULL (`pkey` always, `check` when the expression references zero or several columns). `ChooseIndexNameAddition` never returns an empty string for `key`/`excl`/`fkey`, since those always have at least one key column.
* The shorten-the-longer-one-first loop and the untouched label match, and both sides count bytes (`strlen` / `len`), not characters.
* `clipIdent` backs up from byte `n` while it is a continuation byte, where `pg_mbcliplen` walks forward accumulating whole characters. For valid UTF-8 the two agree on every input.
* Building the column part with a plain `strings.Join` is equivalent to `ChooseIndexNameAddition` / `ChooseForeignKeyConstraintNameAddition`, even though those cap the joined list at `NAMEDATALEN`. The cap can never bite: `availchars` is at most 57 for these labels, so the cut always happens inside `makeObjectName`, before the cap would.
* `pg_query` already truncates an over-long identifier to 63 bytes at parse time (I confirmed both an ASCII and a multibyte name), so `name1` and `name2` never arrive longer than PostgreSQL would have them.
* No other site predicts a server-generated name. The four `autoNameConstraint` call sites and the domain `CHECK` all go through `makeObjectName` now; the PG18 `_not_null` name is matched by suffix in `catalog/columns.go`, so shortening does not affect it; and there is no auto-naming for an unnamed `CREATE INDEX` or for a `serial` sequence.

### Shortening creates duplicates that were not duplicates before

"Duplicate names are still not resolved ... That limit is unchanged" understates the change a little. Before, two distinct constraints always produced two distinct names, and the collision only happened on the server. Now the collision happens in the parser:

```sql
CREATE TABLE public.t1 (
    cccccccccccccccccccccccccccccccccccccccccccccccccccccc1 integer CHECK (cccccccccccccccccccccccccccccccccccccccccccccccccccccc1 > 0),
    cccccccccccccccccccccccccccccccccccccccccccccccccccccc2 integer CHECK (cccccccccccccccccccccccccccccccccccccccccccccccccccccc2 > 0)
);
```

Both columns shorten to the same 54-byte prefix, and the file is now rejected with `duplicate constraint: t1_cccc..._check`. PostgreSQL accepts it and names the second one `..._check1`. So the set of schemas that hit the limit grows from "the same name written twice" to "two names that collide after shortening".

Trading a plan that never converges for a clear parse error is the right way round, and `pista dump` writes constraint names explicitly, so the round trip in the design policy is unaffected. Just worth saying so in the changelog rather than implying nothing moved.

### Suggestion: pin the `key` and `fkey` paths against a real server

The apply fixture covers `check` (and `pkey` indirectly, through the constraint name in the `applied` block), so those two are checked against a running PostgreSQL in CI. `key`, `excl` and `fkey` are only covered by the parser table, which asserts what pistachio predicts, not what the server stores. The expected names were taken from PostgreSQL 16, but nothing keeps them honest afterwards.

Adding an unnamed `UNIQUE` over two long columns and an unnamed foreign key to `testdata/apply/long_auto_constraint_name.yml`, with the existing `verify_no_drift: true`, would close that. The multi-column `UNIQUE` is the most valuable of the two, since it is where the column part is assembled differently from a naive join.

### Minor

* `makeObjectName` drops PostgreSQL's `Assert(availchars > 0)`. With a label over 57 bytes the loop would drive `name1chars` negative and `clipIdent` would panic on `s[:n]`. Every caller passes a 3 to 5 byte literal, so it is unreachable today; noting it in case a longer label is ever added.
* `clipIdent` assumes UTF-8, while `pg_mbcliplen` uses the database encoding. On a single-byte-encoding database PostgreSQL clips at the byte with no backing up, so a high-byte name could come out one or two bytes longer there. Not worth chasing.
* The `diff/policies_test.go` reformat is unrelated to this change. It is a real fix (`main` currently fails gofumpt on that file), and it is already its own commit, so this is only an observation.

---

winebarrel:
## Re-review of bb8cfda

All three follow-ups are handled: the changelog now says shortening can bring two names together, the fixture covers `key` and `fkey` against a real server, and `makeObjectName` carries the note where PostgreSQL asserts. I recomputed the two new expected names and both are right:

* `key`: `availchars` is 58, the column part `region_code_channel_code` is 24 bytes, so the table part clips to 34, giving `customer_subscription_billing_hist_region_code_channel_code_key` at 63 bytes.
* `fkey`: `availchars` is 57, `snapshot_id` is 11 bytes, so the table part clips to 46, which lands on the underscore after `archive` and produces the doubled underscore in `customer_subscription_billing_history_archive__snapshot_id_fkey`. That is what PostgreSQL writes too.

### `test (18)` is red on the new fixture

Not a flake, and not an expectation that just needs updating. The two table names are too close together for PG18:

```
--- Expected
+++ Actual
     id integer NOT NULL,
+    id integer CONSTRAINT customer_subscription_billing_history_archive_2024_id_not_null1 NOT NULL,
```

PG18 records an unnamed `NOT NULL` in `pg_constraint` and names it through the same `makeObjectName`. For label `not_null` the overhead is 10, so `availchars` is 53 and the table part clips to 51 bytes: `customer_subscription_billing_history_archive_2024_`. `..._snapshot` and `..._event` are identical for those 51 bytes, so both tables want the same name for their `id` column. `ChooseConstraintName` checks `ConstraintNameExists(conname, namespaceid)`, which is schema-wide rather than per-relation, so the second one retries with label `not_null1`: `availchars` drops to 52, the table part clips to 50, and the stored name is `customer_subscription_billing_history_archive_2024_id_not_null1`.

`catalog/columns.go` recognises an auto-generated not-null name by the `_not_null` suffix, and `_not_null1` does not match, so the name survives into the dump and the `applied` block gains a `CONSTRAINT` clause. PG15 to PG17 have no named not-null constraints, which is why only `test (18)` fails.

The fixture is what needs to change, not the code. Making the child table diverge before byte 51 fixes it, for instance `customer_subscription_billing_event_archive_2024_registry`. At 57 bytes it still forces the foreign key name to shorten (57 + 11 is over the 57 `availchars`, so the table part clips to 46), so the coverage the commit was added for is kept.

I would not reach for `applied_pg18` here. It would pin `_not_null1` as expected output, which is incidental to what this fixture is testing and would hide the collision rather than avoid it.

### Related, pre-existing, out of scope

The suffix test in `catalog/columns.go` silently misses every disambiguated name. On PG18 any two tables in a schema sharing a long enough prefix give the second one a `_not_null1` name, and `pista dump` then writes an explicit `CONSTRAINT ..._not_null1 NOT NULL` for a column the user wrote as plain `NOT NULL`.

The round trip itself holds: `parser.go:586` reads the name back, and `diff/tables.go:428` only emits a rename when both sides are named, so a dump fed back plans clean. Under the design policy that makes it low priority, and it is not this PR's doing. Worth a note somewhere, not a fix here.

Everything else from the first pass still stands. With the fixture table renamed this is good to merge.

---

winebarrel:
## Re-review of c6fdec2

The rename is the right fix, and dropping the columns first (d733c4b) then putting them back was the right call to reverse - the fixture reads like a real schema again.

I recomputed the new names against `makeObjectName`:

* `customer_subscription_billing_billing_event_archive_2024_registry` is 57 bytes, so `pkey` fits whole at 62 bytes. That label's shortening is still covered by the snapshot table's own `pkey`.
* `fkey`: `availchars` is 57 and `snapshot_id` is 11 bytes, so the table part clips to 46, landing mid-`2024`. `customer_subscription_billing_event_archive_20_snapshot_id_fkey` at 63 bytes is right.
* The collision is genuinely gone rather than moved. A `not_null` name clips the table part to 51 bytes, and the two names now diverge at byte 31 (`event` against `history`), well inside that cut. Nothing else in the fixture shares a prefix that far.

`test (15)`, `(16)`, `(17)`, `(18)`, `samples`, `keywords`, `ascii` and CodeQL are green; `windows` was still running when I wrote this.

The TODO.md entry states the gap accurately, including that the round trip holds.

### Nit

The new paragraph sits between "A second limitation" and "A third limitation", so the third is now the fourth thing on the list. It opens with "The same suffix test", which reads as a continuation of the second, so folding it into that paragraph or moving it below the third would keep the count honest.

LGTM otherwise.

---

winebarrel:
Typo in the first bullet above: the table is `customer_subscription_billing_event_archive_2024_registry`, not `..._billing_billing_event_...`. 57 bytes and the rest of the arithmetic are unchanged.

---

winebarrel:
## Re-review of 358982a

The only change since c6fdec2 is the TODO.md fold, which is exactly the nit and reads better as one paragraph - the disambiguation case is the same suffix test, so it belongs with the sentence before it. "A third limitation" now counts correctly again.

All 14 checks are green, including `windows` and both codecov gates.

Nothing left from my side. LGTM.

---

PR #395, opened 2026-08-20
Name an index written without one

winebarrel:
An index written without a name reached the diff with an empty name, which matched no index in the database. Every run planned the index again, and PostgreSQL numbered each copy rather than refusing, so the same file applied three times left three identical indexes behind and the plan never came back empty.

```
$ pista apply schema.sql   # three times, same file
$ psql -c '\di'
 items_qty_idx
 items_qty_idx1
 items_qty_idx2
$ pista plan schema.sql
CREATE INDEX ON public.items USING btree (qty);
-- skipped: DROP INDEX public.items_qty_idx;
-- skipped: DROP INDEX public.items_qty_idx1;
-- skipped: DROP INDEX public.items_qty_idx2;
```

Name it the way PostgreSQL does, following `ChooseRelationName` (`src/backend/commands/indexcmds.c`): the table, then one name per index element joined with underscores, then `idx`, all through the `makeObjectName` that the constraint names already go through, so an over-long name is shortened rather than cut from the right.

An element takes its column name. An expression takes the name `FigureIndexColname` picks (`src/backend/parser/parse_target.c`), which is the function it calls, the type of a cast with nothing under it, `case`, `coalesce`, `greatest`, `least`, `nullif` or `array`, and `expr` for anything with no name of its own. PostgreSQL's strength rule is kept, so `(a + b)::text` is named after the type while `coalesce(a, b)::text` keeps the function name. The `INCLUDE` list joins the name, and a name repeated within one index takes a number, so `(lower(a), lower(b))` reads `lower_lower1`.

Sort order, an operator class, a `WHERE` clause, the access method and `UNIQUE` leave the name to the columns, and the name carries no schema.

Two unnamed indexes that produce one name are now rejected as a duplicate index name, so `validateNamespaces` no longer skips an unnamed index. PostgreSQL would have numbered the second, which the desired schema alone cannot predict; that is the same limit the constraint names carry.

## Tests

- `parser`: a table of 23 naming cases, plus separate ones for the shapes that need a different statement (unique, schema-qualified), the 63-byte shortening, the duplicate error, and two feature interactions - `CREATE INDEX CONCURRENTLY` without a name, and `-- pista:renamed-from` onto one, which previously renamed to nothing.
- `testdata/apply/unnamed_index.yml`: a plain and an expression index with `verify_no_drift: true`, so the plan has to come back empty after the create. It fails without the fix.
- `testdata/parser/unnamed_index.yml`, `testdata/parser/unnamed_index_schema_qualified.yml`, `testdata/plan/rename_unnamed_index.yml`.

Every expected name was taken from a running PostgreSQL 16, not derived from the implementation.

Coverage of the `parser` package goes from 90.5% to 90.7%. `figureIndexColname` and `autoNameIndex` are at 100%; what is left uncovered in the other two is a guard against an empty name list and a nil `IndexElem`, neither reachable from a parse.

## Verification

Against a local PostgreSQL 16: applying the same file three times now leaves one index per statement and the next plan is empty. `go vet ./...`, `go test -p 1 ./...`, `test/scenario/run.sh`, `make lint`, `make deadcode` and the ASCII check all pass.

## Background

Found while checking sqldef v3.11.20 against pistachio (#394). It is the same class of problem as the constraint names fixed there: a server-generated name the parser did not predict.

---
_Generated by [Claude Code](https://claude.ai/code/session_018RmEfXs5W27TADi5HGHu9X)_


parser/parser.go
```diff
@@ -711,6 +712,125 @@ func autoNameConstraint(tableName string, cols []string, contype pg_query.Constr
...
+// as an expression, following FigureIndexColname and FigureColnameInternal
+// (src/backend/parser/parse_target.c). The second return is PostgreSQL's
+// strength: a name found deeper in the tree wins over one an enclosing cast or
+// CASE falls back to, so `(a + b)::text` is named after the type while
+// `coalesce(a, b)::text` keeps the function name. An empty name means
+// PostgreSQL finds none and the element is called "expr".
+//
+// The node kinds below are the ones an index expression realistically holds.
+// Anything else takes "expr", which is also what PostgreSQL does for every
+// kind it has no name for.
```
winebarrel, line 725: `T_A_Indirection` is one of the kinds an index expression realistically holds, and it is missing. `FigureColnameInternal` takes the last field name of the indirection, and when there is none (a bare subscript) it recurses into the argument. Both shapes end up as `expr` here. Confirmed on PostgreSQL 16:

| index | PostgreSQL 16 | this PR |
| --- | --- | --- |
| `((arr[1]))` | `items_arr_idx` | `items_expr_idx` |
| `(((comp).x))` | `t_x_idx` | `t_expr_idx` |

This is not cosmetic. pistachio now writes the name it picked, so the index is stable once pistachio creates it, but a database that already carries the name the server chose never converges, and the plan is a real drop rather than a skipped one:

```
DROP INDEX public.items_arr_idx;
CREATE INDEX items_expr_idx ON public.items USING btree ((arr[1]));
```

That is the same drift the PR sets out to remove, pointed the other way and now destructive. Roughly:

```go
case *pg_query.Node_AIndirection:
	// A subscript carries no name, so the argument is named instead.
	if name := lastNodeName(n.AIndirection.Indirection); name != "" {
		return name, 2
	}
	return figureIndexColname(n.AIndirection.Arg)
```

`lastNodeName` already ignores non-String nodes by returning `""` for a subscript, so it carries over as is.

The two other kinds that reach an index expression are `T_RowExpr` -> `row` and `T_XmlSerialize` -> `xmlserialize`; both confirmed on 16 (`((ROW(a, a)::ct))` gives `t_row_idx`, this PR gives `t_ct_idx`). Those are corner cases and I would leave them, but a subscript is a shape people write.

---

testdata/apply/unnamed_index.yml
```diff
@@ -0,0 +1,35 @@
...
+  -- public.items
+  CREATE TABLE public.items (
+      id integer NOT NULL,
+      name text,
+      qty integer,
+      CONSTRAINT items_pkey PRIMARY KEY (id)
+  );
+  CREATE INDEX items_lower_idx ON public.items USING btree (lower(name));
+  CREATE INDEX items_qty_idx ON public.items USING btree (qty);
+verify_no_drift: true
```
winebarrel, line 35: This fixture proves the name pistachio picks is stable, not that it is the name PostgreSQL picks: `desired` names the index, so `verify_no_drift` would hold for any name the parser invented. Every other assertion in the PR compares against a name transcribed by hand.

A plan fixture puts the server on the other side of the comparison. Let `init` create the indexes unnamed, so PostgreSQL names them, repeat the same statements in `desired`, and expect an empty plan:

```yaml
init: |
  CREATE TABLE public.items (
      qty integer,
      name text
  );
  CREATE INDEX ON public.items (qty);
  CREATE INDEX ON public.items ((lower(name)));
  CREATE UNIQUE INDEX ON public.items (name, qty);
  CREATE INDEX ON public.items (qty) INCLUDE (name);
desired: |
  <the same>
plan: ""
```

That passes here, and adding `CREATE INDEX ON public.items ((arr[1]))` to it is what surfaced the `A_Indirection` gap. It is the assertion the feature rests on and it costs one fixture.

---

winebarrel (commented):
Reviewed against a local PostgreSQL 16. `go vet ./...` and `go test -p 1 ./...` pass, and I re-ran the whole naming table in `parser_test.go` on 16: every expected name reproduces, including `coalesce(a, b)::text` -> `coalesce` against `(a + b)::text::varchar` -> `varchar`, the `CASE` else branch, and `lower_lower1`. The strength rule is modelled correctly, and the round trip works: a plan fixture that lets the server name the indexes in `init` and repeats the statements in `desired` comes back empty for plain columns, `lower(name)`, `UNIQUE` and `INCLUDE`.

Two things worth acting on, both left inline.

1. `figureIndexColname` has no `A_Indirection` case, so an index over an array element or a composite field is named `expr` where PostgreSQL names it after the column. Because pistachio now writes the name, a database already carrying the server's name plans an unskipped `DROP INDEX` plus a create, on every run.

2. Nothing compares the generated name against a server. `testdata/apply/unnamed_index.yml` pins stability, not agreement. One plan fixture with unnamed indexes in `init` closes that, and it is what found the case above.

## Docs

README carries a "Constraint naming" section with the auto-naming rules and the duplicate-name limitation spelled out. Index naming now has the same rules and the same limitation, and lives only in CHANGELOG. It deserves the same treatment, either its own section or by widening that one to cover both.

While in the area, the second bullet of that section is now false:

> PostgreSQL truncates identifier names to 63 bytes (NAMEDATALEN - 1). pistachio does not apply this truncation, so very long table/column names may produce mismatched constraint names.

The `makeObjectName` change in the same `[Unreleased]` block did apply it. The bullet should go.

## Checked and fine

- `ChooseIndexNameAddition` stops appending column names once the buffer reaches `NAMEDATALEN`, while `autoNameIndex` joins the whole list. The two agree: `makeObjectName` clips the joined part to at most 58 bytes, and PostgreSQL's is at least 64 when it stops, so the surviving prefix is identical. Confirmed with four 30-character columns on a one-character table, both give `u_aaa..._idx` byte for byte.
- The numbering loop does not clip `origname` before appending the digit the way `ChooseIndexColumnNames` does. Same reason, the result cannot differ after `makeObjectName`.
- Naming before the deparse is right, and `CONCURRENTLY` being stripped first keeps `Definition` canonical. The rename directive lands on the generated name, as the new plan fixture shows.
- Dropping the skip in `validateNamespaces` is the correct consequence: an unnamed index now occupies its name like any other, and the duplicate is caught by `setUnique` before it reaches the server.

---

parser/parser.go
winebarrel: Added in cce2368, as suggested. Confirmed on 16: `((arr[1]))` and `((arr[1:2]))` give `items_arr_idx`, `(((comp).x))` gives `items_x_idx`, and all three are in the naming table now.

Left `RowExpr` and `XmlSerialize` alone.

---

testdata/apply/unnamed_index.yml
winebarrel: You are right that the apply fixture only pins stability. Added `testdata/plan/unnamed_index_server_names.yml` in cce2368: `init` leaves all six indexes unnamed for the server to name, `desired` repeats the statements, `plan` is empty. It covers plain columns, `lower(name)`, `UNIQUE`, `INCLUDE`, a subscript and a field selection.

It does catch the `A_Indirection` gap. Reverting that case makes the file fail to parse, since both expression indexes then take `expr` and collide.

---

parser/parser.go
```diff
@@ -711,6 +712,132 @@ func autoNameConstraint(tableName string, cols []string, contype pg_query.Constr
...
+			return name, 2
+		}
+		return figureIndexColname(n.AIndirection.Arg)
+	}
+
+	return "", 0
+}
+
+// lastNodeName returns the last element of a dotted name, which is the part
+// PostgreSQL names a column after.
```
winebarrel: Nit, take it or leave it. The rewrite did more than remove an unreachable guard: skipping every node that is not a String is what makes the two new callers work. A subscript returns the empty string so `A_Indirection` falls back to its argument, and `a.*` gives `a` where the old indexed read gave nothing. That is `FigureColnameInternal`'s "find last field name, if any, ignoring `*`" loop, written out.

The comment still describes it as the last element of a dotted name, which reads like the old indexed version. Something like "the last name in a dotted or subscripted list, skipping subscripts and `*` the way FigureColnameInternal does" would keep the reason visible to whoever touches it next.

---

winebarrel (commented):
Both points are addressed, and I re-checked them against a local PostgreSQL 16 rather than reading the diff. Looks good to merge from my side.

**`A_Indirection`.** Correct, and it holds past the three cases the test table now covers. Every shape I threw at it agrees with the server:

| index | PostgreSQL 16 | this PR |
| --- | --- | --- |
| `((arr[1]))` | `arr` | `arr` |
| `((arr[1:2]))` | `arr` | `arr` |
| `(((comp).x))` | `x` | `x` |
| `((carr[1].y))` | `y` | `y` |
| `((arr[1]::text))` | `arr` | `arr` |

The nested `carr[1].y` is the one I would have expected to slip, since the indirection list mixes a subscript and a field, and it lands on `y` because `lastNodeName` keeps scanning past the subscript. Left a nit there about the comment.

**The server-side fixture.** `unnamed_index_server_names.yml` is a real guard, not just a passing test: dropping the `A_Indirection` case makes it fail with `duplicate index: items_expr_idx`, since both indirection indexes collapse to `expr`. Worth noting that this is the failure mode rather than a name mismatch, so if it ever goes red the message points at the collision and not at the name that moved - still unambiguous enough to debug.

**Verified here:** `go vet ./...` clean, `go test -p 1 ./...` green against 16 including the two new fixtures, and CI is green on 15, 16, 17 and 18, so the naming agrees with every version in the matrix, not only the one the expected names came from. `make lint` I could not run - the local golangci-lint is built against go1.25 and the module targets 1.26 - but the CI job covers it.

**README** now reads right, and dropping the truncation bullet is the correct call rather than rewording it: the shortening is described a paragraph above and no longer belongs in a limitations list.

`T_RowExpr` still differs (`((ROW(a, a)::ct))` gives `t_row_idx` on the server and `t_ct_idx` here), which is the corner case I said I would leave. Noting it only so it is on the record; nothing to do.

---

parser/parser.go
winebarrel: Taken, in 9544025. The comment now says it skips subscripts and `*` the way `FigureColnameInternal` does, and that an empty result is what lets the `A_Indirection` case fall back to the argument.

---

winebarrel (commented):
The nit is addressed and the new wording is accurate on both halves: the skipping is what `FigureColnameInternal` does in its `ColumnRef` and `A_Indirection` loops, and naming the bare-subscript case as the reason the fallback exists is the part that was missing. Nothing else changed - `git diff cce2368..9544025` is four comment lines in `parser/parser.go` and no code.

Re-ran `go vet ./...` and the `parser` package here, both clean, and confirmed the file is still ASCII. CI on `9544025` is green on 15, 16, 17, 18, `samples`, `ascii`, `keywords`, CodeQL and both Analyze jobs; only `windows` was still running when I looked, and it passed on the previous head.

Done from my side. I would merge it. Recording that as a comment rather than an approval only because GitHub will not let this account approve its own pull request.

---

PR #415, opened 2026-08-24
Keep a constraint trigger out of the table's constraints

winebarrel:
`CREATE CONSTRAINT TRIGGER` writes a `pg_constraint` row with `contype='t'` alongside the `pg_trigger` row, and `pg_get_constraintdef` renders it as the bare word `TRIGGER`, followed by the deferral clause when it has one. `ListConstraintsByTable` excluded only `contype='n'`, so the trigger reached the model as a table constraint.

Two consequences, both reproduced on 15 through 18:

* `dump` wrote `CONSTRAINT x TRIGGER DEFERRABLE INITIALLY DEFERRED` inside `CREATE TABLE`. That does not parse, so feeding the dump back failed with `syntax error at or near "TRIGGER"`, breaking the round trip the design policy requires.
* `plan` proposed `ALTER TABLE ... DROP CONSTRAINT` for the trigger. The default drop policy suppressed it, but `--allow-drop all` dropped the user's trigger.

Triggers are unmanaged, so the query skips `contype='t'` too. The other `pg_constraint` readers already restrict themselves: `catalog/domains.go` to `contype='c'` and `catalog/indexes.go` to `contype IN ('p','u','x')`.

## Tests

Written first and confirmed failing against the old query.

* `catalog/constraints_test.go`: a table carrying both a deferrable and an immediate constraint trigger reports only its primary key and its foreign key.
* `testdata/dump/constraint_trigger.yml`: the dump leaves the trigger out.
* `testdata/plan/constraint_trigger_no_diff.yml`: the plan is empty under `allow_drop: [all]`, so the drop would land in the plan rather than in the skipped list.
* `testdata/apply/constraint_trigger_idempotent.yml`: apply runs no SQL and leaves no drift.

`make lint`, `make vet`, `make test` and `make test-scenario` pass. The suite was run against 15, 16, 17 and 18. Coverage is unchanged: the diff adds a SQL predicate and no Go branch, and the three blocks still uncovered in `constraints.go` are the `Query`, `Scan` and `rows.Err` error wrappers, matching every other catalog reader.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Keep a constraint trigger out of the table's constraints. `CREATE CONSTRAINT TRIGGER` also writes a `pg_constraint` row, which the catalog read as a table constraint, so `dump` emitted `CONSTRAINT x TRIGGER ...`, which does not parse, and `plan` proposed dropping the trigger. Triggers stay unmanaged.
+
 ## [1.28.0] - 2026-08-23
 
 * Require Go 1.27 to build.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-* Keep a constraint trigger out of the table's constraints. `CREATE CONSTRAINT TRIGGER` also writes a `pg_constraint` row, which the catalog read as a table constraint, so `dump` emitted `CONSTRAINT x TRIGGER ...`, which does not parse, and `plan` proposed dropping the trigger. Triggers stay unmanaged.
+* Keep a constraint trigger out of the table's constraints. `CREATE CONSTRAINT TRIGGER` also writes a `pg_constraint` row, and the catalog read it as a table constraint. `dump` then emitted `CONSTRAINT x TRIGGER ...`, which does not parse, and `plan` proposed dropping the trigger. Triggers remain unmanaged.
 
 ## [1.28.0] - 2026-08-23
 
--- a/catalog/constraints.go
+++ b/catalog/constraints.go
@@ -59,10 +59,12 @@ func (c *Catalog) ListConstraintsByTable(ctx context.Context, table *model.Table
 			-- query only returns table-level constraints.
 			AND con.contype <> 'n'
 			-- CREATE CONSTRAINT TRIGGER records the trigger in pg_constraint
-			-- too (contype='t'), and pg_get_constraintdef renders it as the
-			-- bare word TRIGGER. It is a trigger, not a table constraint, so
-			-- reading it here made dump write invalid SQL and plan propose a
-			-- DROP CONSTRAINT for it. Triggers are not managed.
+			-- too (contype='t'), where pg_get_constraintdef renders it as the
+			-- bare word TRIGGER, followed by the deferral clause when it has
+			-- one. Reading that as a table constraint made dump write a
+			-- CREATE TABLE that does not parse, and plan propose a DROP
+			-- CONSTRAINT for the trigger. Triggers are unmanaged, so this
+			-- query skips the row.
 			AND con.contype <> 't'
 		ORDER BY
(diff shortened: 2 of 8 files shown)


winebarrel:
LGTM.

The diagnosis matches the catalog: `contype='t'` is only ever a constraint trigger, `pg_get_constraintdef` renders it as the bare word `TRIGGER`, and nothing in the tree reads `pg_trigger`, so a trigger has no business reaching `model.Table.Constraints`. Dropping it in `ListConstraintsByTable` fixes both the broken dump round trip and the `DROP CONSTRAINT` in the plan at the one place the row enters the model, without touching the diff. The `column_t` CTE still sees the row but joins on `con.oid`, so nothing leaks there either. On the desired side a `CREATE CONSTRAINT TRIGGER` is an unsupported statement the parser ignores with a warning, so both directions now agree that the trigger is unmanaged.

Test coverage is the right shape: the catalog test pins the row count at the source, the dump fixture pins the round trip, and the plan fixture with `allow_drop: [all]` makes the drop visible rather than suppressed, which is exactly what would have caught the bug.

Two optional notes, neither blocking:

* The `WHERE` clause is now two exclusions, while the query already carries the positive set in its `ORDER BY` (`'{p,u,c,x,f}'`). `AND con.contype IN ('p', 'u', 'c', 'x', 'f')` would say the same thing in one line, keep the two lists in sync, and match how `catalog/domains.go` and `catalog/indexes.go` already restrict themselves. It also stays correct if a future major version adds another `contype`, where the current form would let it through and sort it to the front with a NULL `array_position`.
* `model.ConstraintType.IsConstraintTrigger()` is referenced only by its own test now; it was already unused before this PR, but this change makes it unreachable by construction, so it could go with it.

Reviewed by reading the code; the suite was not run here (no PostgreSQL in this environment).

---

winebarrel:
Both notes taken.

The `WHERE` clause now names the set instead of excluding two types: `AND con.contype = ANY ('{p,u,c,x,f}'::"char"[])`, the same list the `ORDER BY` uses, so the two cannot drift apart. The comment above it still records where the other two types go, since that is not obvious from the list alone. `array_position` can no longer return NULL either, so the ordering is fully determined.

`ConstraintType.IsConstraintTrigger` is gone, along with its assertion in `model/constraint_test.go`.

One more test gap turned up while re-reading. The catalog test only had a `DEFERRABLE INITIALLY DEFERRED` trigger, whose definition carries the deferral clause; a trigger without one renders as the bare word `TRIGGER`. Both shapes are covered now, next to a foreign key that has to keep coming through, and the assertion compares the whole key list rather than a count so over-filtering fails too. Added `testdata/apply/constraint_trigger_idempotent.yml` as well: the plan fixture only pinned the diff, and this one pins that apply executes nothing and leaves no drift.

`make lint`, `make vet`, `make test` and `make test-scenario` pass, and the suite was run against 15, 16, 17 and 18.

winebarrel:
LGTM. Both notes are addressed, and the added coverage is more than I asked for.

`con.contype = ANY ('{p,u,c,x,f}'::"char"[])` now names the same set the `ORDER BY` sorts by, so a `contype` a later major version adds is excluded rather than admitted and sorted to the front on a NULL `array_position`. `IsConstraintTrigger` is gone along with its assertion, and nothing else referenced it.

The catalog test is stronger than the count it replaced: the immediate trigger covers the bare `TRIGGER` definition with no deferral clause, and asserting `CollectKeys()` on both `Constraints` and `ForeignKeys` shows the primary key and the `contype='f'` row still come through rather than just that the total is right.

The apply fixture closes the gap plan leaves. Worth noting for anyone reading it later that it works because `AppliedSQL` is a `*string`: `applied_sql: ""` decodes to a non-nil pointer, so the empty-SQL assertion actually runs. With `verify_no_drift` on top, the fixture pins that the run executes nothing and the schema still matches afterward.

CI is green on 15 through 18 plus samples and CodeQL (windows still running as I write this). Nothing further from me.

---

PR #416, opened 2026-08-24
Manage triggers

winebarrel:
Triggers were the largest object still left to `-- pista:execute`. They are now read from `pg_trigger`, written by `dump`, and diffed: plain triggers, `CREATE CONSTRAINT TRIGGER`, and `INSTEAD OF` triggers on views.

## Reading

The catalog reads `pg_get_triggerdef` in its pretty form, which keeps the `WHEN` expression free of the parentheses the plain form wraps every operator in and adds no line breaks of its own. Three kinds of row are left out:

* The internal triggers a foreign key installs. PostgreSQL creates and drops them with the constraint.
* The clones a partitioned table's trigger leaves on each partition. `DROP TRIGGER` on one fails with `cannot drop trigger ... because trigger ... requires it`, and the parent's definition already covers them.
* Anything an extension owns, the same `pg_depend` check `ListIndexes` uses.

## Comparing

The two sides render the same trigger differently. `pg_get_triggerdef` spells out `FOR EACH STATEMENT` and the `AS` in `REFERENCING`, adds casts in the `WHEN` expression, and drops a schema `search_path` reaches; the deparser does none of that. So the comparison parses both definitions, fills in an implicit relation schema the way `normalizeFKSchema` does for a foreign key, clears `OR REPLACE`, runs the `WHEN` expression through `normalizeCheckExpr`, and compares the deparsed result. Comparing the trees directly does not work: every node carries the byte offset of its token, so two spellings of the same trigger differ as soon as their lengths do.

The schema a definition gives the function it calls is left alone, so it follows `search_path` like every other name the catalog's deparsers write. `--search-path=` qualifies everything, as documented.

## Writing

A definition change is applied with `CREATE OR REPLACE TRIGGER`, which PostgreSQL has carried since 14 and which takes `SHARE ROW EXCLUSIVE` rather than the `ACCESS EXCLUSIVE` that `DROP TRIGGER` takes. Turning a trigger into a constraint trigger or back is the one change PostgreSQL rejects in place (`trigger "x" for relation "t" is a constraint trigger`), so that runs as `DROP` and `CREATE` under the new `trigger` drop policy.

Either statement leaves the trigger enabled, so a state other than the default is set again behind it. `CREATE TRIGGER` cannot express a disabled trigger, so the state travels as the `ALTER TABLE ... DISABLE TRIGGER` that PostgreSQL uses and `dump` writes, the way RLS already travels as `ALTER TABLE ... ENABLE ROW LEVEL SECURITY`. A view has no state to carry: PostgreSQL rejects `DISABLE TRIGGER` on one.

`DROP VIEW` takes the view's triggers with it, so the recreate path puts them back, next to where it already re-emits a materialized view's indexes. `CREATE OR REPLACE VIEW` keeps them, so that path diffs them instead.

`-- pista:renamed-from` renames a trigger. The rename also rewrites the name inside the definition, since that is what the catalog reports afterwards and the comparison would otherwise read the old name as a definition change.

`-m` and `--omit-schema` reach a trigger's schema and its definition, alongside the indexes and policies they already reached.

## Out of scope

The function a trigger calls stays unmanaged; `-- pista:execute-first` exists for exactly this and the README now says so. Event triggers belong to the database rather than to a schema. The `ALL` and `USER` forms of `ENABLE TRIGGER` name no single trigger.

A column rename does not rewrite a trigger's `UPDATE OF` list or `WHEN` expression, so the first plan after one emits a redundant `CREATE OR REPLACE TRIGGER` and the second is clean. Same class as the view and cross-table FK entry already in TODO.md, and recorded there.

## Tests

* `catalog/triggers_test.go`: definitions and enable state, internal triggers, partition clones, a view's trigger.
* `parser/trigger_test.go`: the default-schema fallback, constraint triggers, `OR REPLACE` normalization, a trigger on a view, duplicates, the same name on two tables, the four enable states, the rename directive.
* `diff/triggers_test.go`: add, drop, suppressed drop, replace, replace keeping a non-default state, each state transition, the constraint switch, four rename cases, and the parse-error paths.
* `model/trigger_test.go`, plus 13 plan fixtures, 3 apply fixtures with `verify_no_drift`, and 3 dump fixtures including `--omit-schema`.

`make lint`, `make vet`, `make test` and `make test-scenario` pass. The suite was run against 15, 16, 17 and 18. `make test-samples` passes all 46 real-world schemas, which between them carry 55 triggers (pagila 15, dvdrental 15, ledgersmb 11, dotcms 10, discourse 4).

Coverage: `model` 99.6 -> 99.7, `diff` 96.6 -> 96.3, `catalog` 88.9 -> 88.8, `parser` 92.0 -> 91.9. What is left uncovered in the new files is the deparse-after-a-successful-parse and `Query`/`Scan`/`rows.Err` wrappers, the same defensive paths already uncovered in every neighbouring file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* Manage triggers. `CREATE TRIGGER`, `CREATE CONSTRAINT TRIGGER` and `INSTEAD OF` triggers on views are read from `pg_trigger`, written by `dump`, and diffed. A definition change uses `CREATE OR REPLACE TRIGGER`; switching a trigger to or from a constraint trigger runs as `DROP` and `CREATE`, gated by `--allow-drop trigger`. The enable state is carried as `ALTER TABLE ... ENABLE/DISABLE TRIGGER`, and `-- pista:renamed-from` renames a trigger. The function a trigger calls stays unmanaged, so it belongs in a `-- pista:execute-first` statement.
+
 * Keep a constraint trigger out of the table's constraints. `CREATE CONSTRAINT TRIGGER` also writes a `pg_constraint` row, and the catalog read it as a table constraint. `dump` then emitted `CONSTRAINT x TRIGGER ...`, which does not parse, and `plan` proposed dropping the trigger. Triggers remain unmanaged.
 
 ## [1.28.0] - 2026-08-23
--- a/README.md
+++ b/README.md
@@ -869,7 +869,7 @@ CREATE TRIGGER events_stamp BEFORE UPDATE ON public.events
 
 A trigger names the table or view it sits on, so that relation has to be declared earlier in the file, or in an earlier file when the schema is split across several.
 
-A definition change is applied with `CREATE OR REPLACE TRIGGER`, which takes a lighter lock than dropping and recreating. Turning a trigger into a constraint trigger, or back, is the one change PostgreSQL rejects in place; it runs as `DROP TRIGGER` and `CREATE TRIGGER` and needs `--allow-drop trigger`.
+A definition change is applied with `CREATE OR REPLACE TRIGGER`, which takes a lighter lock than dropping and recreating. Turning a trigger into a constraint trigger, or back, is the one change PostgreSQL rejects in place; it runs as `DROP TRIGGER` and `CREATE TRIGGER` and needs `--allow-drop trigger`, the same as any other trigger drop. Without it the trigger keeps its current definition.
 
 `CREATE TRIGGER` has no syntax for a disabled trigger, so a desired schema asks for one the way `dump` writes it:
 
--- a/README.md
+++ b/README.md
@@ -108,7 +108,8 @@ pista apply ./schema/*.sql         # apply it
 - Indexes (unique, partial, expression, hash, multi-column)
 - Comments (on tables, columns, views, types, domains, composite types, composite attributes, sequences)
 - Row-level security (`ALTER TABLE ... ENABLE/DISABLE/FORCE/NO FORCE ROW LEVEL SECURITY`, policies via `CREATE POLICY` / `ALTER POLICY` / `DROP POLICY`)
-- Renaming (tables, views, enums, enum values, domains, composite types, composite attributes, sequences, columns, constraints, foreign keys, indexes, policies via `-- pista:renamed-from` directive)
+- Triggers (`CREATE TRIGGER`, `CREATE CONSTRAINT TRIGGER`, `INSTEAD OF` triggers on views, and the enable state via `ALTER TABLE ... ENABLE/DISABLE TRIGGER`). The function a trigger calls is not managed; see [Triggers](#triggers).
+- Renaming (tables, views, enums, enum values, domains, composite types, composite attributes, sequences, columns, constraints, foreign keys, indexes, policies, triggers via `-- pista:renamed-from` directive)
 - Array, JSON, UUID, and other built-in types
 - Quoted identifiers
 
@@ -518,7 +519,7 @@ pista plan --assume-validated schema.sql
 pista apply --assume-validated schema.sql
 ```
 
-By default, `plan` and `apply` do not drop tables, views, enums, domains, composite types, columns, constraints, foreign keys, or indexes. Use `--allow-drop` to enable dropping specific object types (`all`, `table`, `view`, `enum`, `domain`, `composite_type`, `column`, `constraint`, `foreign_key`, `index`, `policy`). Also available as `$PISTA_ALLOW_DROP`. `constraint` covers CHECK / UNIQUE / PRIMARY KEY / EXCLUSION; foreign keys are governed by `foreign_key` separately. `composite_type` also gates `DROP ATTRIBUTE` on a composite type.
+By default, `plan` and `apply` do not drop tables, views, enums, domains, composite types, columns, constraints, foreign keys, or indexes. Use `--allow-drop` to enable dropping specific object types (`all`, `table`, `view`, `enum`, `domain`, `composite_type`, `column`, `constraint`, `foreign_key`, `index`, `policy`, `trigger`). Also available as `$PISTA_ALLOW_DROP`. `constraint` covers CHECK / UNIQUE / PRIMARY KEY / EXCLUSION; foreign keys are governed by `foreign_key` separately. `composite_type` also gates `DROP ATTRIBUTE` on a composite type.
 
 ```bash
 # Allow all drops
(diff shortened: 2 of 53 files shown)


winebarrel:
Pushed a follow-up after re-reading the coverage report line by line.

One real gap turned up. The diff unit tests had no trigger carrying a `WHEN` expression, so the line that runs it through `normalizeCheckExpr` never executed and the hardest part of the comparison rested on the fixtures alone. That is now covered directly: the `::text` cast `pg_get_triggerdef` adds, the `= ANY (ARRAY[...])` it writes where the file says `IN`, an expression that really differs and has to produce a replace, and the arguments it renders as strings.

Four shapes the suite did not reach are now fixtures.

* A trigger on a partitioned parent. The clone on the partition has to stay out of the dump, or feeding the dump back would try to drop something PostgreSQL refuses to drop.
* A trigger declared on a partition, which is the only thing that takes the child branch of the table diff.
* A view and its trigger arriving together, which is the branch of `DiffViews` the earlier fixture skipped.
* A view keeping its definition while losing a trigger.

A parser test pins `EXECUTE PROCEDURE`, the spelling several of the sample schemas still use.

The README section gains the two things a reader would otherwise learn by running it: a trigger's relation has to be declared before the trigger, in the same file or an earlier one; and the function name follows `search_path` like every other name the catalog reports, so a `pg_dump`-style file that writes `public.stamp()` wants `--search-path=`.

Eleven blocks in the four new files are still uncovered, all of them defensive: `Query` / `Scan` / `rows.Err` in the catalog reader, four deparse failures on a tree that has just parsed, one error propagation that cannot fire because the definition parsed a line earlier, and the `Relation == nil` and `cmd == nil` guards. `parser/policy.go` carries nine of the same kind on its own, so the shape matches what is already there. Reaching them needs a broken connection or a hand-built parse tree, which CLAUDE.md asks not to write tests for.

`make lint`, `make vet`, `make test` and `make test-scenario` pass, on 15, 16, 17 and 18.

winebarrel:
Reviewed the whole diff (catalog, parser, model, diff, dump, remap, docs, fixtures) against `main`. The shape follows the existing policy/index paths closely and the fixture coverage is good. CI is green on 15-18 plus samples. LGTM.

One real discrepancy and a few nits.

## Constraint switch and `--allow-drop trigger`

The docs and the code disagree about the DROP a constraint-trigger switch emits.

README:

> it runs as `DROP TRIGGER` and `CREATE TRIGGER` and needs `--allow-drop trigger`

CHANGELOG:

> switching a trigger to or from a constraint trigger runs as `DROP` and `CREATE`, gated by `--allow-drop trigger`

In `diff/triggers.go` the recreate DROP is unconditional; `triggerAllowed` only guards the pure removal branch:

```go
	if recreated[name] {
		dropName := name
		...
		stmts = append(stmts, dropTriggerSQL(fqtn, dropName))
	}
```

So with no `--allow-drop` at all, a constraint switch still drops the trigger. Nothing in the suite pins this either way: the plan harness defaults to `AllowDrop: ["all"]` when a fixture omits `drop_policy` (`plan_test.go:374`), so `testdata/plan/trigger_constraint_switch.yml` does not exercise the default policy, and `TestDiffTriggers_ConstraintSwitch` passes `allowTriggerDrops()`.

The code as written matches `diffIndexes`, which documents exactly this split ("Pure removals honor the index-drop policy; definition changes still run DROP+CREATE"), so keeping it ungated is defensible. But `DiffViews` takes the other side and gates a definition-forced recreate on the view drop policy. Either reading is fine; the docs should just say what the code does. Whichever way it goes, a fixture with `allow_drop: []` over a constraint switch would pin it.

## Nits

* `model/trigger.go`: the `Definition` comment says "the diff compares parse trees rather than this text", while `equalTriggerDef` says the opposite - the trees carry token offsets, so it compares the deparsed text. Worth aligning, since the offset problem is the non-obvious part.
* `renameTriggerDef` and `replaceTriggerSQL` index `result.Stmts[0]` directly. `parseTriggerDef` checks `len(result.Stmts) != 1` first; the other two check `ct == nil` but would panic on a definition that parses to zero statements. Unreachable with a catalog or parser definition, but the check is one line and matches the neighbour.
* `Trigger.StateSQL()` and `diff.alterTriggerStateSQL()` render the same `ALTER TABLE ... TRIGGER` from two places, one off `trg.FQTN()` and one off the passed `fqtn`. They always agree (the parser only attaches a trigger to the relation its definition names, and `fqtn` is `desired.FQTN()`), so this is duplication rather than a bug.

## Test suggestion

`testdata/plan/trigger_partition_child.yml` covers a trigger declared on a partition, and the dump fixture covers the parent's clone staying out of the model. What is not covered is a definition change on a partitioned parent's trigger: `CREATE OR REPLACE TRIGGER` there has to recurse into the clones, and only the catalog's parent row is compared afterwards. An apply fixture with `verify_no_drift` over that case would rule out a per-run replace loop on partitioned tables.

The unqualified function name following `search_path`, and the redundant replace after a column rename, both read as deliberate given the design policy and are documented in README / TODO.md.

---

winebarrel:
All of this addressed, pushed as 2c49875.

## Constraint switch and --allow-drop trigger

Fixed in code, kept the docs as they were: the switch's DROP now honors triggerAllowed the same as the removal branch, matching what README and CHANGELOG already said.

Went with the DiffViews precedent over diffIndexes. Losing a trigger, even for the one statement between DROP and CREATE, drops the behavior it runs; an index rebuild has no equivalent gap. So a denied switch now leaves the trigger exactly as it was: no DROP, no CREATE (which would otherwise apply the desired definition without ever removing the old one), and a bundled rename stays suppressed too, the same as when the switch goes through. Added the denied case, a denied-plus-rename case, and a plan fixture with allow_drop: [] to pin it against a genuine default-policy run rather than the harness's default all.

## Nits

* renameTriggerDef and replaceTriggerSQL now check len(result.Stmts) != 1 before indexing, matching parseTriggerDef. Covered with a two-statement input.
* Pulled the shared "ALTER TABLE fqtn ACTION TRIGGER name;" text into model.TriggerStateSQL; Trigger.StateSQL and diff.alterTriggerStateSQL both call it now. They still differ in when they call it (StateSQL suppresses the default state, alterTriggerStateSQL always fires, since going back to default after a replace needs the explicit ALTER), so the branching stayed where it was.
* Reworded the Definition comment to point at equalTriggerDef instead of asserting the opposite of what it does.

## Test suggestion

Added testdata/apply/trigger_partition_replace.yml: CREATE OR REPLACE TRIGGER on a partitioned parent, checked against a live server first to confirm PostgreSQL itself recurses the replace into the clone, then verify_no_drift over the applied result.

make lint, make vet, make test and make test-scenario pass on 15, 16, 17 and 18.

winebarrel:
Re-reviewed 2c49875. All four points are addressed, and the gating is implemented the way the commit message describes. CI is green on 15-18 including samples, windows and codecov. LGTM.

What I checked on the gating:

* `recreateDenied` suppresses the DROP, the CREATE, and the state statement, so a denied switch leaves the trigger untouched rather than applying the desired definition without clearing the old one.
* A bundled rename stays suppressed: `recreated[name]` is true either way, so the RENAME loop skips it and the skipped-DROP names the trigger that actually exists (`events_old`). `TestDiffTriggers_RenameWithRecreateDenied` pins it.
* Only `-- skipped: DROP TRIGGER ...` goes to `DisallowedDropStmts`, with no CREATE alongside it. Same shape as the denied view recreate in `DiffViews`, so the two read alike.
* `testdata/plan/trigger_constraint_switch.yml` now sets `allow_drop: [trigger]` explicitly and the new `trigger_constraint_switch_disallowed.yml` covers the default policy, so neither direction rests on the harness default any more.

`testdata/apply/trigger_partition_replace.yml` passing on all four versions is the confirmation I was after: the parent's replace reaches the clone and comparing only the parent's row afterwards finds no drift.

One leftover nit, in the new `model.TriggerStateSQL` comment:

> Exported so diff can reuse it for the case StateSQL suppresses: a trigger going back to the default state after CREATE OR REPLACE TRIGGER reset it, which still needs the statement CREATE TRIGGER alone would not.

The first half is right - `StateSQL` returns "" for the default state, so the diff needs its own renderer to emit `ENABLE TRIGGER`. The second half does not match the call site: `alterTriggerStateSQL` runs only under `if !changed[name]`, where the definition is identical and just the state moved. After a `CREATE OR REPLACE TRIGGER` the diff takes the `des.StateSQL()` path, and the replace already leaves the trigger in the default state, so no statement is needed there at all. Trimming it to "a trigger whose definition did not change going back to the default state" would say it.

---

winebarrel:
Fixed in 7192236: trimmed the comment to "a trigger whose definition did not change going back to the default state", dropping the CREATE OR REPLACE TRIGGER claim that did not match the call site.

make lint and the relevant tests pass; full suite green on 15.

winebarrel:
Re-reviewed `790783c..7192236`. Everything raised is addressed, and the fix went the way that matches `DiffViews`. LGTM, nothing further from me.

**Gating.** `recreateDenied` now suppresses the whole recreate, not just the DROP: the CREATE is skipped in the desired loop, the RENAME stays suppressed (so the skipped-DROP comment names the relation that actually exists), and the state re-application goes with it. That is exactly the shape `DiffViews` uses for a denied view recreate, down to emitting only the commented DROP. The README no longer overstates the old behavior, and CHANGELOG was already correct.

The two new unit tests pin both halves - the plain switch and the switch bundled with a rename - and `testdata/plan/trigger_constraint_switch_disallowed.yml` pins it through the harness with `allow_drop: []`, which is what the old fixture could not do since the harness defaults to `all`. `trigger_constraint_switch.yml` now declares `allow_drop: [trigger]` explicitly, so the allowed path is no longer riding on the default either.

One consequence worth knowing, not a request: a rename bundled with a denied switch reports only the skipped DROP, so the plan says nothing about the rename it also held back. `DiffViews` behaves the same way, so consistency seems like the right call here.

**Partitioned parent.** `testdata/apply/trigger_partition_replace.yml` covers what I asked about, and it answers the open question rather than just guarding it: `CREATE OR REPLACE TRIGGER` on the parent recurses into the clone, the clone stays out of the dump, and `verify_no_drift` shows no per-run replace loop. Green on 15-18.

**Nits.** `len(result.Stmts) != 1` added to both helpers with multi-statement cases in the tests, the `Definition` comment now points at `equalTriggerDef`, and the duplicated `ALTER TABLE ... TRIGGER` collapsed into `model.TriggerStateSQL`. The comment on the exported helper explains why `StateSQL` cannot serve the default-state transition, which is the part that would otherwise look redundant.

Locally: `go build ./...`, `go vet ./...`, and `go test ./diff/... ./model/... ./parser/...` pass at `7192236`. All 14 checks on the PR are green, windows and samples included, and codecov/patch and codecov/project both pass now.

---

PR #424, opened 2026-08-25
Manage functions and procedures behind --manage-routine

winebarrel:
Functions and procedures were the one thing a schema file could not declare. README pointed at `-- pista:execute` with a version tag embedded in `COMMENT ON FUNCTION`, which asks the user to compute the diff by hand.

## Opt-in

Routines are off by default. Without `--manage-routine` (`$PISTA_MANAGE_ROUTINE`) `pg_proc` is never read, so a schema maintained with `-- pista:execute` keeps working and plan output is unchanged down to the summary line: `ObjectCount.Routines` is a pointer and a nil leaves the slot out entirely.

```console
$ pista plan schema.sql
-- Plan for schema public (2 tables, 0 views, 0 enums, 0 domains, 0 composite types, 0 sequences)

$ pista plan --manage-routine schema.sql
-- Plan for schema public (2 tables, 0 views, 0 enums, 0 domains, 0 composite types, 0 sequences, 3 routines)
```

`routine` also joins the `--enable` / `--disable` type list, where it narrows what `--manage-routine` turned on rather than turning routines on by itself, and the `--allow-drop` list, where it covers functions and procedures alike.

## One parse path

The catalog reads each routine as the `CREATE` statement `pg_get_functiondef` renders and hands it to the same parser the desired schema goes through, so both sides build one `model.Routine` and render through one SQL writer.

```
desired SQL ────────┐
                    ├─→ parseCreateFunctionStmt → *model.Routine → compare
pg_get_functiondef ─┘
```

That is what keeps a dump fed back as the desired schema planning clean, without the definition-text normalization views and triggers need. The normalization that remains is small and exists so the two spellings meet: a parameter's type modifier is dropped the way PostgreSQL drops it, an attribute left at its default is not written back (`pg_get_functiondef` prints neither `VOLATILE` nor `PARALLEL UNSAFE` nor the default `COST`), and `SET x TO y` and `SET x TO 'y'` parse to the same value.

## Identity and replacement

A routine is keyed by name **and identity argument types**, so an overload set is several independent objects and changing an argument type is a drop and a create. Changes apply with `CREATE OR REPLACE`; the ones PostgreSQL refuses in place run as `DROP` and `CREATE`, gated by `--allow-drop routine`:

- changing the return type, including the columns of a `RETURNS TABLE`
- renaming a parameter
- removing a parameter default
- turning a function into a procedure, or back

Adding or removing a parameter is not a modification either, since the signature is the identity.

## Ordering

Two edges. A routine depends on the types its signature names. Every table and view depends on every routine, so routines come first for the `CHECK` constraints, `GENERATED` expressions, index expressions, policies and triggers that call them. The edge is drawn wholesale rather than by reading each expression — the answer is the same "routine first" either way.

The graph keys a routine by name alone under a `routine:` prefix. A `CREATE` statement spells its parameters with names and defaults rather than identity types, so no FQRN can be recovered from the statement text, and a schema may hold a table and a function of the same name, which without the prefix would collapse into one node and read as a cycle.

The reverse edge is deliberately absent, so a `LANGUAGE sql` body that reads a table created in the same run fails to apply. Verified and documented; `plpgsql` is unaffected.

## Not managed

Left out on **both** sides of the diff, so neither `dump` writes them nor `plan` proposes dropping them: aggregates, window functions, extension-owned routines, and a body written as `BEGIN ATOMIC`. The last one is not a parsing limitation — pg_query handles it — but such a body records real `pg_depend` entries on what it reads, which cannot hold together with creating routines before tables. Renaming a routine is not supported and is an error rather than a silent drop-and-create.

## Verification

Run against PostgreSQL 16.

- `make test` — new fixtures under `testdata/{plan,apply,dump}`, plus Go tests for `model`, `parser`, `catalog`, `diff` and `toposort`. The apply fixtures carry `verify_no_drift: true` with one per normalization rule, since that round trip is the design
- `make test-scenario` — new `routine` scenario, 10 steps: an overload set through a body change, a `CHECK` constraint calling a routine, a trigger function, a procedure, a return-type change that has to run as a drop and a create, and the drop gate
- `make test-samples` — the runner now exports `PISTA_MANAGE_ROUTINE=1`, so routines are part of the real-world round trip. **39 samples passed with 0 drift, 779 routines in total**: chado 94, gitlab 337, musicbrainz 130, kea 130, sourcegraph 37, dotcms 17, dvdrental and pagila 9 each for their trigger functions

Three bugs were found and fixed during implementation, each by a test rather than by reading: `RETURNS SETOF SETOF int` (`deparseTypeName` folds `setof` into the type name), `SET work_mem TO "64MB"` (double quotes read as an identifier), and the toposort node name (independent nodes sort alphabetically, so insertion order decides nothing).

## Not done

- `make lint` could not run here: the installed golangci-lint is built with Go 1.25 and cannot load a `go 1.27` config. `go vet` and `gofumpt` are clean, and the config enables only `errorlint`, `unconvert`, `gofumpt` and `goimports`. **Needs CI to confirm**
- Eight samples could not be loaded from this environment and are left to CI: five pgFoundry tarballs and clubdata behind unreachable hosts, discourse and osm needing pgvector and PostGIS
- The coverage table in `sample-db-test.md` has no Routines column. It would mean loading and counting ~45 sample databases, which those eight block, so the numbers are not invented. Filed in `TODO.md` along with routine renaming, body-dependency ordering, `-m` not rewriting bodies, `BEGIN ATOMIC`, and aggregates/window functions

https://claude.ai/code/session_01MUvq6kGhcXcMv5dWXejp3j

---
_Generated by [Claude Code](https://claude.ai/code/session_01MUvq6kGhcXcMv5dWXejp3j)_


winebarrel:
Reviewed at 8895715, built the branch and ran it against PostgreSQL 16.

The design holds up well: one parse path for both sides, opt-in by default with the nil `Routines` slot keeping the summary line byte-identical, and the dump round trip planned clean in every check I ran. `golangci-lint run` is clean here (0 issues), so the lint item under "Not done" looks settled.

Three things I think need a fix, plus one nit.

## 1. A signature that names a table cycles the graph, and plan silently emits the wrong order

`addRoutineDeps` draws `table -> routine` for every pair, and `resolveTypeDep` resolves against `defined`, which holds tables and views as well as types. So `RETURNS SETOF <table>`, or a table row type as a parameter, closes a cycle.

```sql
CREATE TABLE users (id integer NOT NULL, name text);

CREATE OR REPLACE FUNCTION all_users()
    RETURNS SETOF users
    LANGUAGE sql
    AS $$SELECT * FROM users$$;
```

`dump --manage-routine --sort-by-deps` fails outright:

```
pista: error: failed to order dump by dependency: schema dependency sort failed: cycle detected: sorted 1 of 4 nodes
```

`plan` swallows the error and takes `fallbackOrder`, which puts the routine first:

```
$ pista apply --manage-routine dump.sql   # against an empty schema
CREATE OR REPLACE FUNCTION pistest.all_users()
    RETURNS SETOF pistest.users
    ...
pista: error: failed to execute SQL: ...: ERROR: type "pistest.users" does not exist (SQLSTATE 42704)
```

The input above is `pista dump` output, so this is the round trip breaking on a fresh database, not a hand-written spelling. `RETURNS SETOF <table>` is a common shape. One way out is to skip the wholesale `table -> routine` edge for the tables a routine's own signature names, so the routine lands after its own table and before the rest. `toposort/routines_test.go` has no case where a signature names a relation.

## 2. A schema-qualified type in a signature plans a `CREATE OR REPLACE` and a `DROP` of the same routine

The catalog reports a type reached by `search_path` bare, the desired schema may write it qualified, and the FQRN is built from the raw string, so the two spellings become two keys:

```sql
-- current: CREATE DOMAIN dom AS text; CREATE TABLE t(c dom);
--          CREATE FUNCTION f(x dom) RETURNS text LANGUAGE sql AS $$SELECT x$$;

CREATE DOMAIN public.dom AS text;
CREATE TABLE public.t (c public.dom);
CREATE OR REPLACE FUNCTION public.f(x public.dom)
    RETURNS text
    LANGUAGE sql
    AS $$SELECT x$$;
```

```
$ pista plan --manage-routine --allow-drop routine schema.sql
CREATE OR REPLACE FUNCTION public.f(x public.dom)
    ...
DROP FUNCTION public.f(dom);
```

Creates run before drops, and both statements name the same function, so the plan replaces it and then deletes it. Every run repeats. Without `--allow-drop routine` it is a permanent `-- skipped: DROP FUNCTION` that never converges. Note the table in the same file plans clean: `stripTypeSchema` in `diff/tables.go` already normalizes exactly this, and the routine path needs the same treatment on `RoutineArg.Type` and `ReturnType`. The README line "write them the way dump does" is a fair policy, but a plan that drops a routine the desired schema still declares is worse than drift.

## 3. An unsupported routine option is a hard error, including with `--manage-routine` off

`parseRoutineOptions` returns a plain error for any defname it does not handle, and `SUPPORT` and `TRANSFORM FOR TYPE` are two real ones. On main such a statement is warned about and dropped:

```
$ pista plan schema.sql          # main
pistachio: ignored unsupported statement: CREATE FUNCTION public.f_support(...) ... SUPPORT public.f_supp ...
-- Plan for schema public (...)
CREATE TABLE public.t1 (...);

$ pista plan schema.sql          # this branch, no --manage-routine
pista: error: failed to parse SQL file: unsupported routine option: support
```

That contradicts "plan output is unchanged" for anyone who has not opted in. The catalog side has the same shape: with `--manage-routine`, a database holding one such function makes `dump` fail entirely with `catalog: failed to parse routine definition: unsupported routine option: support`. Returning `errUnsupportedRoutine` the way the `window` case does would degrade both sides to warn and skip.

## 4. Nit: `SET x FROM CURRENT` renders invalid SQL

`parseRoutineConfig` reads a `VariableSetStmt` with no args and `RoutineConfig.SQL` renders the trailing `TO` with nothing after it:

```
$ pista plan --manage-routine schema.sql
CREATE OR REPLACE FUNCTION public.g()
    RETURNS text
    LANGUAGE sql
    SET search_path TO 
    AS $$SELECT 'x'$$;
```

`pg_get_functiondef` never writes `FROM CURRENT`, so this is desired-side only, but the statement reaches the plan and fails at apply. Rejecting it reads better than emitting it.

winebarrel:
All four reproduced exactly as described, including the apply failing on pistachio's own dump. Fixed in 698bb99.

## 1. Signature naming a relation

You were right about the mechanism. `resolveTypeDep` resolves against `defined`, which `collectDefined` builds from all six maps, so a relation in a signature resolves like a type and the wholesale `table -> routine` edge points back at it.

`addRoutineDeps` now records what each signature names and skips those pairs when drawing the wholesale edge. The routine lands after its own relation and before every other one, which is what your suggestion asked for. Overloads share a graph node, so the sets merge per node.

`toposort/routines_test.go` covers `RETURNS SETOF <table>` and a table row type as a parameter, both asserting the routine sits between its own table and an unrelated one. `testdata/apply/function_returns_setof_table.yml` pins the end to end case with `verify_no_drift`.

## 2. Schema-qualified type in a signature

Taken as a correctness bug rather than drift, for the reason you gave: a plan that drops a routine the desired schema still declares is not the same thing as a redundant statement.

`FQRN` now strips the routine's own schema from the identity types, so the two spellings key alike. `stripTypeSchema` moved from `diff/tables.go` to `model` and both call it, rather than a second copy. The diff compares on the stripped form too, or the pair would still read as a `needsDropCreate`.

One thing your sketch did not cover: a `DROP` or a `COMMENT` has to name the types as they were read, since PostgreSQL resolves them under the same `search_path` the catalog reported them through. Stripping there would break `--search-path=`, where the catalog qualifies everything. So the key and the rendered statement are now separate, `FQRN` and `Signature`, and `diff/routines_test.go` asserts the drop keeps the qualified form.

## 3. Unsupported option is a hard error

This was the worst of the four, since it broke people who never opted in. Both sides now skip: the parser returns `ErrUnsupportedRoutine` and warns, and `ListRoutines` skips the row instead of failing the read, so neither side of the diff sees the routine and no drop is proposed. `testdata/plan/routine_unsupported_option_ignored.yml` runs without `--manage-routine` and asserts the table still plans.

## 4. SET x FROM CURRENT

Joined the same unsupported set rather than erroring, for the reason above: the desired schema is read whether or not `--manage-routine` is set, so an error would land on someone who never asked for routines. The session value it captures is not in the statement, so there is nothing to compare either way.

## Verification

`make test`, `make test-scenario` and `go vet` are clean on PostgreSQL 16, and dvdrental, pagila and chado still round trip (9, 9 and 94 routines). README and CHANGELOG list the two new unmanaged cases and the relation-in-signature ordering rule.

The lint item under "Not done" is settled, thank you for confirming; CI has been running `golangci-lint` in the `test` job across all four PostgreSQL versions since the first push.

---

winebarrel:
Re-reviewed at 7c5753f, rebuilt and re-ran the same repros against PostgreSQL 16. `go vet`, `golangci-lint run` (0 issues) and the full `go test -p 1 ./...` are clean here.

All four are fixed:

1. The cycle is gone. `plan` now emits the table before the routine, `apply` on an empty schema succeeds, and the re-plan is clean. Skipping only the relations a routine's own signature names is a nice way out.
2. `f(x public.dom)` against a catalog `f(dom)` plans to no changes, and the `FQRN` / `Signature` split keeps `DROP` and `COMMENT ON` naming the type the way it was read. Moving `StripTypeSchema` into `model` is the right home.
3. `SUPPORT` and `TRANSFORM FOR TYPE` warn and skip on both sides, so a `dump --manage-routine` of a database holding one no longer fails.

Two things left.

## `dump --sort-by-deps` silently falls back to name order as soon as a routine is present

`dependencyOrderedSQL` keys `pos` by the graph node name, which for a routine is `RoutineNode(...)` = `routine:public.f`, but `appendDumpItems` looks the routine up by the map key, which is the FQRN `public.f()` (dump.go:294). The lookup always misses, `ok` comes back false, and the whole dump degrades to the category order. This predates the fix commits; the cycle error was hiding it.

```sql
CREATE TABLE z (id int PRIMARY KEY);
CREATE TABLE a (id int, z_id int REFERENCES z(id));
```

```
$ pista dump --manage-routine --sort-by-deps
-- pistest.z
-- pistest.a

$ psql -c "CREATE FUNCTION noop() RETURNS int LANGUAGE sql AS 'SELECT 1'"
$ pista dump --manage-routine --sort-by-deps
-- pistest.noop()
-- pistest.a          <- FK to z, now emitted before it
-- pistest.z
```

One unrelated function is enough to break the ordering guarantee for every object in the dump. It also puts a routine ahead of the table its own signature names, so the `RETURNS SETOF <table>` dump from item 1 above comes out in an order that cannot be applied:

```
$ pista dump --manage-routine --sort-by-deps -n pistest
-- pistest.all_users()
CREATE OR REPLACE FUNCTION pistest.all_users()
    RETURNS SETOF pistest.users
...
-- pistest.users
CREATE TABLE pistest.users (
```

The routine lookup needs to go through `toposort.RoutineNode(model.Ident(r.Schema, r.Name))` rather than the map key.

## `SET x FROM CURRENT` now plans a `DROP` of a routine the schema file declares

`parseRoutineConfig` returns `ErrUnsupportedRoutine`, so the desired side drops the routine. The catalog does not: `pg_get_functiondef` writes the resolved value, so the routine is read back normally and the two sides no longer leave out the same ones, which is the invariant the `ErrUnsupportedRoutine` comment rests on.

```sql
CREATE OR REPLACE FUNCTION public.g()
    RETURNS text
    LANGUAGE sql
    SECURITY DEFINER
    SET search_path FROM CURRENT
    AS $$SELECT 'x'$$;
```

```
$ pista plan --manage-routine --allow-drop routine schema.sql
pistachio: ignored unsupported statement: CREATE OR REPLACE FUNCTION public.g() ...
-- Plan for schema public (0 tables, ..., 1 routine)
DROP FUNCTION public.g();
```

The routine is declared right there in the file and pistachio deletes it. That is worse than the invalid `SET search_path TO ` it replaced, which at least failed loudly. `-- pista:ignore` semantics are what this case wants: setting `Ignore` on the routine instead of dropping it from the map makes `removeIgnored` take the key off both sides, so it is neither created, altered, nor dropped. The other unsupported cases do not need it, since the catalog skips those anyway.

## Nits

- `CLAUDE.md` lists the optional fixture fields per suite and does not mention `manage_routine`.
- The `dep != node` guards in `addRoutineDeps` can never be false: `resolveTypeDep` returns a key from `defined`, and nothing there carries the `routine:` prefix.

winebarrel:
Both fixed at 0bc1195. Each has a fixture that fails without the fix.

## `dump --sort-by-deps`

Your diagnosis is exactly right. `dependencyOrderedSQL` keys `pos` by the graph node name, `appendDumpItems` looks the routine up by the map key, and the two never match.

I kept `appendDumpItems` as it is and aliased each FQRN onto its node's position right after `pos` is built, so the other six call sites stay unchanged. An FQRN always carries its parentheses, so it cannot collide with another kind of key, and overloads share a node and therefore a position, which the stable sort preserves.

```
$ pista dump --manage-routine --sort-by-deps -n pistest
-- pistest.noop()
-- pistest.users
-- pistest.all_users()
-- pistest.z
-- pistest.a
```

`z` before `a` again, and `all_users()` after the table its signature names. Piping that dump into an empty schema now replays without error, which is the property that was actually broken.

You are right that `function_sort_by_deps.yml` could not see this: its dependency order and the fallback order render identically, so it passes either way. The two new fixtures put the orders in conflict. `routine_sort_by_deps_tables.yml` has a foreign key that disagrees with the name order plus one unrelated function, and `routine_sort_by_deps_setof_table.yml` covers `RETURNS SETOF <table>`. Both fail on the previous commit.

## `SET x FROM CURRENT`

Agreed, and the invariant argument is the right way to see it. `ErrUnsupportedRoutine` rests on both sides leaving the same routines out, and this is the one case where the catalog reads the routine back, so the desired side must not drop it alone.

Split it into its own sentinel and set `Ignore` instead. One thing your note did not mention that had to change with it: `parser.go` did `routine.Ignore = ignore`, which cleared the flag the routine had just set for itself. It is now `routine.Ignore = routine.Ignore || ignore`.

```
$ pista plan --manage-routine --allow-drop routine schema.sql
-- Plan for schema pistest3 (0 tables, ..., 0 routines)
-- ignored: pistest3.g()
-- No changes
```

README and CHANGELOG moved it out of the "not managed" list, since that list is defined by both sides leaving the object out and this one is no longer that.

## Nits

`AGENTS.md` now matches the structs. `manage_routine` was missing from all three, and so were `sort_by_deps` for dump and `ignored`, `bulk_alter`, `assume_validated` for plan. The apply line excludes `ignored` as well now.

Both `dep != node` guards are gone with a comment saying why no self-edge is possible. The `dep != k` in the composite-type loop stays, since an attribute can name the type itself.

## Checks

`go vet`, `gofmt`, the ASCII check, `make test` and `make test-scenario` are clean here against PostgreSQL 16. `make lint` still cannot run in this environment for the Go version reason, so CI remains the check for it.


---

winebarrel:
Re-reviewed at 0bc1195. Both are fixed, verified against PostgreSQL 16.

**`dump --sort-by-deps`.** Aliasing each FQRN onto its node position is the smallest fix and it holds up. An unrelated routine no longer disturbs anything else (`z` still precedes the table with the foreign key to it), a routine now sorts after the relation its signature names, and overloads keep their map order behind the shared position. The dependency-ordered dump applies to an empty schema and re-plans clean, with `--omit-schema` too.

```
$ pista dump --manage-routine --sort-by-deps
-- public.norm(text)
-- public.norm(text, boolean)
-- public.z
-- public.rows_of()       <- RETURNS SETOF z
-- public.a               <- FK to z
```

**`SET x FROM CURRENT`.** Marking the routine `Ignore` is the right call, and separating `errRoutineConfigFromCurrent` from `ErrUnsupportedRoutine` keeps the reason legible: the one invariant holds for the options the catalog also skips, and does not hold here. `plan --allow-drop routine` now reports `-- ignored: public.g()` and proposes nothing, and the `allow_drop: [all]` fixture pins the destructive case. `routine.Ignore || ignore` in the parser is the correct order too.

I re-ran the earlier repros and they all still hold: the `RETURNS SETOF <table>` apply from an empty schema, the qualified signature type, and `SUPPORT` warning and skipping on both sides. `go vet`, `golangci-lint run` (0 issues), `make test` and `make test-scenario` (including the new `routine` scenario) are all green here.

The remaining limitations are the documented ones: `-m` leaves a routine body alone, an expression that calls a routine defined later needs `-- pista:execute`, and a signature type in a third schema is not normalized, the same as for a column type. All three are in TODO.md or the README.

LGTM.

PR #428, opened 2026-08-25
Warn about ignored ALTER TABLE actions and COMMENT targets

winebarrel:
`ALTER TABLE` and `COMMENT ON` each have a parser case of their own, so neither reached the `ignored unsupported statement` warning that covers an unhandled statement type. The parser reads only some of what they carry and dropped the rest without a word, which turned into a plan proposing the opposite of what the file asked for: a column added by `ALTER TABLE ... ADD COLUMN` was absent from the desired schema, so the plan proposed dropping it from the database. `ALTER COLUMN ... SET DEFAULT` / `SET NOT NULL` / `TYPE`, `DROP COLUMN`, `OWNER TO`, `SET UNLOGGED` and `COMMENT ON INDEX` / `CONSTRAINT` / `SCHEMA` / `TRIGGER` went the same way.

The warning now follows a list of what the parser reads rather than a list of what to reject, so an action or comment target added later warns instead of vanishing. `alterTableSupportedCmds` names `AT_AddConstraint`, the four row-level security toggles and the four trigger states; `commentTargetSupported` names table, view, materialized view, column, sequence, type, domain, function and procedure.

`warnIgnoredAlterTableCmds` rebuilds a statement from the ignored actions alone, so one that mixes a read action with a dropped one reports only the dropped one, and keeps the original location for the snippet fallback a deparse failure uses. `COMMENT ON` is checked at the entry to the case rather than in the switch inside `parseCommentStmt`, because a target such as `SCHEMA` carries no object list and returned before reaching it.

```
pistachio: ignored unsupported statement: ALTER TABLE public.a ADD COLUMN x text
pistachio: ignored unsupported statement: ALTER TABLE public.a ALTER COLUMN id TYPE bigint
pistachio: ignored unsupported statement: ALTER TABLE public.a OWNER TO postgres
pistachio: ignored unsupported statement: COMMENT ON INDEX public.a_id_idx IS 'i'
pistachio: ignored unsupported statement: COMMENT ON SCHEMA public IS 's'
```

Nothing but the warning changes: the actions were dropped before and are dropped now. A statement marked `-- pista:execute` never reaches the case, so it stays silent.

An `ALTER TABLE` naming a relation the file does not declare is still skipped without a warning. `CREATE INDEX` and the trigger states treat such a relation the same way, deliberately, so changing it is a separate call. Clause-level omissions inside a supported statement, such as a table's `WITH (fillfactor = 70)`, are also untouched; catching those means enumerating every field the parser does not read.

## Testing

Ten tests in `parser/parser_test.go`, written before the change and confirmed failing first, and the existing `TestParseSQL_AlterTableNonAddConstraint` now asserts the warning. They cover a dropped action, a statement mixing a dropped action with a read one, several dropped actions in one statement, `IF EXISTS`, `-- pista:execute` silencing the warning, each unsupported comment target, and silence for every action and comment target the parser reads. Two paths in `parseCommentStmt` that had no test of their own are covered as well: a comment on a composite type attribute, and an explicit `NULL` comment clearing one set earlier in the file.

`go vet ./...` and `gofmt` are clean and `make test` is green against PostgreSQL 16. The parser package goes from 91.4% to 92.4% and the repository from 94.3% to 94.5%; `warnIgnoredAlterTableCmds` is fully covered. What is left uncovered in the code this touches is defensive: an empty snippet, an object list that a supported comment target cannot produce, and a `COMMENT ON COLUMN` with a single name, which is not valid SQL.

`pista dump` output fed back still parses without a warning. `ALTER TABLE ONLY ... ADD CONSTRAINT ... FOREIGN KEY`, `ENABLE ROW LEVEL SECURITY` and `DISABLE TRIGGER` are all in the read set, and a dump then plan round trip prints `-- No changes` with empty stderr.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-* Warn about an `ALTER TABLE` action or a `COMMENT ON` target the parser drops. Both statements are supported as a whole, so neither reached the `ignored unsupported statement` warning, and the parts pistachio does not read were dropped in silence: `ALTER TABLE ... ADD COLUMN` left the column out of the desired schema, so the plan proposed dropping it from the database, and `ALTER COLUMN ... SET DEFAULT` / `SET NOT NULL` / `TYPE` went the same way. The actions the parser reads (`ADD CONSTRAINT`, the row-level security toggles, the trigger states) and the comment targets it models (table, column, view, materialized view, sequence, type, domain, function, procedure) are now listed explicitly, so an action or target added later warns instead of vanishing. A statement that mixes a read action with a dropped one reports only the dropped one. An `ALTER TABLE` naming a relation the file does not declare is still skipped in silence, the same as a `CREATE INDEX` on such a relation.
+* Warn about an `ALTER TABLE` action or a `COMMENT ON` target the parser drops. Both have a parser case of their own, so neither reached the `ignored unsupported statement` warning: a column added by `ALTER TABLE ... ADD COLUMN` was absent from the desired schema and the plan proposed dropping it from the database, and `ALTER COLUMN ... SET DEFAULT` / `SET NOT NULL` / `TYPE` went the same way. The warning follows what the parser reads, so an action or target added later warns instead of vanishing. An `ALTER TABLE` naming a relation the file does not declare is still skipped without one.
 
 ## [1.30.0] - 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Warn about an `ALTER TABLE` action or a `COMMENT ON` target the parser drops. Both statements are supported as a whole, so neither reached the `ignored unsupported statement` warning, and the parts pistachio does not read were dropped in silence: `ALTER TABLE ... ADD COLUMN` left the column out of the desired schema, so the plan proposed dropping it from the database, and `ALTER COLUMN ... SET DEFAULT` / `SET NOT NULL` / `TYPE` went the same way. The actions the parser reads (`ADD CONSTRAINT`, the row-level security toggles, the trigger states) and the comment targets it models (table, column, view, materialized view, sequence, type, domain, function, procedure) are now listed explicitly, so an action or target added later warns instead of vanishing. A statement that mixes a read action with a dropped one reports only the dropped one. An `ALTER TABLE` naming a relation the file does not declare is still skipped in silence, the same as a `CREATE INDEX` on such a relation.
+
 ## [1.30.0] - 2026-08-25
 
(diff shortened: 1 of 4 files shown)


winebarrel:
LGTM.

I checked this out and verified the parts that matter for the round trip. Everything `dump` emits stays silent: the only `ALTER TABLE` forms it writes are `ADD CONSTRAINT` (`model/constraint.go`), the RLS toggles (`model/table.go`) and the trigger states (`model/trigger.go`), and the only `COMMENT ON` targets are table, column, view, materialized view, type, domain, sequence and function/procedure. All of those are in the two new allow lists. I also walked the `desired:` blocks of every plan/apply fixture and the `dump:` blocks of every dump fixture, plus `test/scenario/testdata`, and found no statement that would newly warn. `go vet`, `gofmt` and `go test ./parser/` are clean here (no local PostgreSQL, so I did not run the DB-backed suites; CI is green on all four versions).

Driving the check off what the parser reads rather than off a reject list is the right call, and rebuilding the statement from the ignored actions alone reads well - a mixed statement reporting only the dropped half is exactly what a user needs to see. Checking the comment target at the entry to the case rather than inside `parseCommentStmt` is also correct given that `SCHEMA` returns before the switch. Placing the call after the `tables.GetOk` lookup keeps the undeclared-relation case silent, consistent with `CREATE INDEX`.

Two minor points, neither blocking:

1. README line 119 says pistachio "drops any other statement in a schema file" and warns for each one. After this change the warning can also come from an action inside a statement the "Supported Objects" list presents as supported, which is the case a user coming from a `pg_dump` file (`ALTER TABLE ... ALTER COLUMN ... SET DEFAULT nextval(...)`, `OWNER TO`) will hit most. One sentence there would keep the docs matching the behavior, and it is the natural place to point at `-- pista:execute` as the way to silence it.

2. A table whose `CREATE TABLE` carries `-- pista:ignore` still warns for its `ALTER TABLE` actions:

```
-- pista:ignore
CREATE TABLE public.t (id integer);
ALTER TABLE public.t ADD COLUMN x text;
```
```
pistachio: ignored unsupported statement: ALTER TABLE public.t ADD COLUMN x text
```

The table is out of the diff, so the dropped action cannot mislead anyone, and the warning is noise. Low priority - it is a corner case, and the fix is a `t.Ignore` check.

---

winebarrel:
Re-reviewed at d657a29. Both points are addressed, and the extra test commit is a good addition. LGTM.

The `t.Ignore` guard sits in the right place - after the lookup that resolves the relation and before the three helpers, so an ignored table skips only the warning and nothing else changes. Leaving `COMMENT ON` alone is the correct call for the reason the commit message gives: an unsupported target such as `INDEX` or `SCHEMA` names no relation the parser resolves, so there is no `Ignore` flag to consult, and a comment on a table the file does declare is a supported target that never reaches the warning anyway. The README sentence now covers the case it did not, and naming the two forms a `pg_dump` file carries points the reader at what they will actually see. The shortened changelog entry reads better and picks up the ignore case.

The tests in 0ec48b6 pin the things I would have wanted pinned: several dropped actions collapsing to one line, `IF EXISTS` surviving the rebuild into the warning, and `-- pista:execute` silencing it. The two `parseCommentStmt` tests (composite attribute, `IS NULL` clearing an earlier comment) cover paths that had none.

Verified locally on d657a29: `go vet ./...`, `gofmt` and `go test ./parser/` are clean, and coverage is 92.4% as the commit message claims. Still no local PostgreSQL here, so the DB-backed suites are unrun; CI was green on the first commit and the run for this head is in flight.

---

PR #430, opened 2026-08-25
Accept a regular expression in --include / --exclude

winebarrel:
A pattern wrapped in slashes is a regular expression; anything else stays a wildcard, and the two forms mix in one invocation.

```bash
pista plan -E '/^posts_\d+$/' schema.sql        # numbered partitions, which a wildcard cannot express
pista dump -I 'user*' -I '/^audit_(log|trail)$/'
```

## Why this shape

Keeping both forms on the same two flags leaves everything around them untouched: `$PISTA_INCLUDE` / `$PISTA_EXCLUDE` and the `include:` / `exclude:` config keys carry a regular expression with no code change (both verified against a live database), and the include-then-exclude precedence stays as it was. A separate `--include-regex` would have turned two flags into four and raised a question this form never asks: what a pattern in `--include` and one in `--include-regex` mean together.

The delimiters do not collide with real names. `CREATE TABLE a/b` is a syntax error, so only a quoted identifier can hold a slash, and only one that both starts and ends with one would change meaning.

## Semantics

A regular expression matches the way Go matches one: anywhere in the name unless the expression anchors itself. A wildcard still has to match the whole name. Wrapping the expression in an implicit `\A(?:...)\z` would have made the two consistent, but then `/^posts_/` would match nothing, which is worse than the difference it removes. `README.md` states the rule.

An invalid expression is rejected by `ValidatePatterns` before the database is read, in the same shape as the existing wildcard error:

```
pista: error: invalid --exclude pattern "/[bad/": error parsing regexp: missing closing ]: `[bad`
```

## Note on cost

`matchPattern` compiles the expression per name rather than once per run. `FilterOptions` is used directly by library callers and by the existing tests without `AfterApply`, so precompiled state would need a second lazy path through `MatchName`. Compilation is bounded by object count and lost in the catalog round trips, so this takes the simpler version; say the word if you would rather have the precompiled field.

## Testing

`go test -p 1 ./...` passes against PostgreSQL 16.13. New cases in `filter_test.go` cover both flags, the unanchored semantics, the two forms mixed, `//`, a single leading slash staying a wildcard, a regexp metacharacter staying literal in a wildcard, and an invalid expression in each flag. `testdata/plan/include_regexp.yml` and `exclude_regexp.yml` cover the plan path. `go vet` is clean; `golangci-lint` could not run in this environment (its binary is built with Go 1.25 against the repo's 1.27 target).

---
_Generated by [Claude Code](https://claude.ai/code/session_014ujhGwnE1mcGpEJjtJbQvB)_


winebarrel:
LGTM.

Reviewed the diff and ran `go vet ./...` plus the root-package filter tests on `8afefe7` locally; both clean, and Codecov shows every new line covered.

The shape is right. Keeping both forms on the same two flags is what makes `$PISTA_INCLUDE` / `$PISTA_EXCLUDE` and the config keys work with no code change, and it avoids the "what do `--include` and `--include-regex` mean together" question that a separate flag would have created. `regexpPattern` requiring both delimiters keeps `/`, `/user` and `user/` as wildcards, and the tests pin each of those down, along with a slash inside the expression and a regexp metacharacter staying literal in a wildcard. Routing `ValidatePatterns` through the same `matchPattern` means the regexp is checked before the catalog is read, in the existing error shape.

Two notes, neither blocking:

* Compiling per name is the right call here. A precompiled field would need a second lazy path through `MatchName` for callers that skip `AfterApply`, and the cost is invisible next to the catalog round trips. Agreed with the simpler version.
* `//` compiles to the empty expression, so `-E '//'` filters out every object and `-I '//'` keeps every one. It is consistent with how Go reads that expression and the test states it, so nothing to change; just worth knowing it is silent rather than an error.

The one behavior change is a quoted identifier literally named `/x/`, which now reads as a regexp. That needs a name both starting and ending with a slash, so it is the kind of corner case not worth code.

---

PR #438, opened 2026-08-26
Make IS NOT DISTINCT FROM idempotent

winebarrel:
PostgreSQL stores a written IS NOT DISTINCT FROM as the negation of
IS DISTINCT FROM, and the expression normalization did not fold the two
together, so a desired schema that wrote the operator never converged: a
CHECK was dropped and re-added on every run, which revalidates every row,
and an index using it was rebuilt.

The fold sits in normalizeExprNode, which every expression comparison goes
through, so a view body, a policy USING and WITH CHECK, a trigger WHEN clause
and a domain CHECK are covered too. A generated column did not drift but
failed the run outright, since PostgreSQL cannot alter a generated expression
in place.

PostgreSQL rewrites the operator but leaves the resulting double negation
alone, and so does the fold: the walk visits children first, so a stored
NOT (NOT (a IS DISTINCT FROM b)) folds at the inner negation only and lands on
the tree the written form parses to. No reverse fold and no double-negation
elimination is needed. Folding towards the operator keeps the generated DDL in
the spelling the author used.

The fixtures cover each expression the normalization reaches, the double
negation, the dump-form desired schema that has to keep planning clean now
that the fold runs on the desired side too, and a real IS DISTINCT FROM to
IS NOT DISTINCT FROM change that must not be swallowed. Every fixture that
pins the fix was run against a build with the fold removed and fails there.
A column DEFAULT carries the operator into the catalog verbatim and is
covered by the same fold, but gets no fixture: a DEFAULT cannot reference a
column, so both operands are constants and nobody writes this.

What stays out of scope is the operator against NULL, wherever it is written.
Parse analysis rewrites `a IS NOT DISTINCT FROM NULL` to `a IS NULL`, so every
expression the catalog stores comes back as a NullTest rather than the
negation, and no fold over the negation reaches it. `dump` writes `IS NULL`,
so the round trip that matters plans clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Stop `IS NOT DISTINCT FROM` from drifting. PostgreSQL stores the operator as the negation of `IS DISTINCT FROM`, so a desired schema that wrote it never matched the catalog: a `CHECK` was dropped and re-added on every run, which revalidates the whole table, and an index using it was rebuilt. The fold sits in the shared expression normalization, so a view body, a policy `USING` / `WITH CHECK`, a trigger `WHEN` clause and a domain `CHECK` get it too. A generated column did not drift but failed the run, since a generated expression cannot be altered in place.
+
 ## [1.31.0] - 2026-08-25
 
 * Accept a regular expression in `--include` / `--exclude`. A pattern wrapped in slashes is one, so `-E '/^posts_\d+$/'` reaches a numbered partition set that a wildcard cannot; anything else stays a wildcard, and the two forms mix in one invocation. A regular expression matches anywhere in the name unless it is anchored; a wildcard still has to match the whole name. An invalid expression is rejected before the database is read. The flags did not change, so `$PISTA_INCLUDE` / `$PISTA_EXCLUDE` and the config file carry both forms.
--- a/TODO.md
+++ b/TODO.md
@@ -583,6 +583,38 @@ Workaround: write the call unqualified.
 
 Origin: found while adding `-- pista:execute-first`, 2026-07-31.
 
+## Perpetual drift on the DISTINCT operators against NULL
+
+Priority: low.
+
+A comparison written against a NULL literal, as
+`CHECK (a IS NOT DISTINCT FROM NULL)`, is re-emitted on every plan. Applying
+it succeeds and changes nothing, so `plan --check` stays at exit code 2
+forever. Parse analysis rewrites the comparison to `a IS NULL` before it is
+stored, so `pg_get_constraintdef` and `pg_get_expr` hand back a `NullTest`
+while the desired side keeps what the user wrote. `IS DISTINCT FROM NULL`
+against `IS NOT NULL` goes the same way.
+
+The rewrite happens in parse analysis, so it lands on every expression the
+catalog stores rather than on one node kind: a CHECK, an index predicate and
+a view body written that way all come back as `IS NULL`. A column DEFAULT
+does too, though a DEFAULT cannot reference a column, so both operands there
+are constants.
+
+`foldNotDistinct` (`diff/tables.go`) folds
+`NOT (a IS DISTINCT FROM b)` into `a IS NOT DISTINCT FROM b`, which is the
+shape the catalog stores for every other operand, but a `NullTest` carries no
+negation to fold and is a different node. pg_query_go stops at the raw
+parser, so the desired side never reaches the rewrite on its own. Matching it
+means rewriting an `A_Expr` of kind `AEXPR_NOT_DISTINCT` whose operand is a
+NULL `A_Const` into a `NullTest` on the desired side, and `AEXPR_DISTINCT`
+into its negation.
+
+`dump` writes `IS NULL`, so a schema kept through `pista dump` never hits
+this.
+
+Workaround: write `IS NULL` / `IS NOT NULL`.
+
 ## Perpetual drift on a serial column written as an explicit default
 
 Priority: low.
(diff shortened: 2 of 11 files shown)


winebarrel:
LGTM.

I checked out the branch and ran the plan/apply suites against a local PG16, then re-ran them with the `foldNotDistinct` call short-circuited: `check_is_not_distinct_from_idempotent`, `no_diff_is_not_distinct_from`, `..._double_negation`, `..._expressions` and `..._generated_column` all fail without the fold, so the fixtures really do pin the fix. The other two (`alter_check_to_is_not_distinct_from`, `..._stored_form`) pass either way, which is what you want from a regression guard.

The fold itself reads right. `NOT (a IS DISTINCT FROM b)` and `a IS NOT DISTINCT FROM b` are exactly equivalent since `IS DISTINCT FROM` is strict-free and never returns NULL, and both spellings parse to an `A_Expr` with the same `Name`, so flipping `Kind` lands on the tree the written form produces rather than something merely deparse-equal. Children-first in `pgast.Walk` gives the double-negation behaviour the comment describes, and PG16 confirms the premise: a written `CHECK (NOT (a IS NOT DISTINCT FROM b))` comes back from `pg_get_constraintdef` as `CHECK ((NOT (NOT (a IS DISTINCT FROM b))))`. Placing the fold before `alignCurrentCasts` also matters and is done right: the pair walk only descends where both sides hold the same node kind, so without the fold running first a `BoolExpr` on one side and an `A_Expr` on the other would stop the cast alignment dead. `CHECK ((NOT (a IS DISTINCT FROM '0'::integer)))` vs `CHECK (a IS NOT DISTINCT FROM 0)` compares equal on this branch, which is that path working.

One correction to the description, not the code: the reason given for leaving out a column `DEFAULT` fixture does not hold. PostgreSQL does not constant-fold stored defaults, so

```sql
CREATE TABLE d1 (b bool DEFAULT ('x' IS NOT DISTINCT FROM 'y'), c integer DEFAULT (1+1));
```

stores `(NOT ('x'::text IS DISTINCT FROM 'y'::text))` and `(1 + 1)` verbatim. So the operator does survive into a stored default and did drift there before this change. The good news is that the fix already covers it for free: `equalDefault` runs `normalizeCheckExpr` over both sides, and I confirmed the two forms above compare equal on this branch. Given a `DEFAULT` cannot reference a column, both operands are constants and nobody writes that, so no fixture is warranted; just worth not leaving the wrong reason in the commit message.

---

winebarrel:
You are right, and my reason was wrong. Checked on PG15:

```
b|(NOT ('x'::text IS DISTINCT FROM 'y'::text))
c|(1 + 1)
e|(1 IS NULL)
```

No constant folding. What misled me is column `e`: I tested with a NULL
operand, which PostgreSQL rewrites to `x IS NULL`, and I read that as folding.
The general case stores the negation verbatim, so the operator did drift in a
stored DEFAULT before this change, and `equalDefault` running
`normalizeCheckExpr` over both sides covers it now. Confirmed against the
branch: column `b` drifts with the fold short-circuited and does not with it in
place. Column `e` still drifts either way, which is the NULL-operand rewrite
and not something this fold reaches.

The commit message does not mention DEFAULT; the wrong reason was in the PR
description, which I have corrected. No fixture added, for the reason you give.

winebarrel:
Thanks, the corrected description reads right. One scoping note on the NULL operand, then I am done: it is not confined to `DEFAULT`. The rewrite happens in parse analysis, so it lands on every expression the catalog stores. On PG16:

```sql
CREATE TABLE t (a int, b int, CONSTRAINT c1 CHECK (a IS NOT DISTINCT FROM NULL));
CREATE INDEX i1 ON t (b) WHERE (a IS NOT DISTINCT FROM NULL);
CREATE VIEW v AS SELECT b FROM t WHERE a IS NOT DISTINCT FROM NULL;
```

comes back as `CHECK ((a IS NULL))`, `WHERE (a IS NULL)` and `WHERE a IS NULL`. pg_query_go stops at the raw parser, so the desired side keeps `A_Expr`/`AEXPR_NOT_DISTINCT` while the catalog side is a `NullTest`, and on this branch `equalConstraintDef("CHECK ((a IS NULL))", "CHECK (a IS NOT DISTINCT FROM NULL)")` is false. `IS DISTINCT FROM NULL` against `IS NOT NULL` is false the same way.

I would leave it. `dump` writes `IS NULL`, so the round trip that matters plans clean, and it is the drift-from-another-spelling case the design policy puts last. Just worth saying "the operator against NULL, wherever it is written" rather than tying it to `DEFAULT`, so the next person does not re-derive it from a CHECK. Still LGTM.

---

winebarrel:
Reproduced on PG15, same result:

```
CHECK ((a IS NULL))
CREATE INDEX i1 ON nl.t USING btree (b) WHERE (a IS NULL)
 WHERE t.a IS NULL
```

All three drift on this branch, so it is the whole stored-expression surface
and not a DEFAULT quirk. `dump` writes `IS NULL` and feeding that back plans
clean, so I agree it should stay as is.

Description reworded: the NULL note now stands on its own as "the operator
against NULL, wherever it is written", with the parse-analysis rewrite as the
mechanism, and the DEFAULT paragraph only covers why that one gets no fixture.

Say the word if you would rather have it as a TODO.md entry too, so it does not
live only in a PR description.

winebarrel:
Added it as a TODO.md entry in f52c47c, under "Perpetual drift on the DISTINCT
operators against NULL". It notes that the rewrite happens in parse analysis
and so lands on every stored expression rather than on DEFAULT, why the fold
cannot reach a NullTest, what matching it would take on the desired side, and
that dump writes IS NULL so the round trip stays clean. Workaround is to write
IS NULL / IS NOT NULL.

PR #439, opened 2026-08-26
Make the DISTINCT operators against NULL idempotent

winebarrel:
Parse analysis rewrites a comparison against a bare NULL literal into an
IS NULL test before the expression is stored, so pg_get_constraintdef and
pg_get_expr hand back the test while a desired schema that wrote the operator
kept it. The two never matched: a CHECK was dropped and re-added on every run,
which revalidates every row, and an index using it was rebuilt.

The fold added for IS NOT DISTINCT FROM cannot reach this one, since the stored
form is another node rather than a negation. The rewrite joins it in
normalizeExprNode, which every expression comparison goes through, so an index
expression and predicate, a view body, a policy USING and WITH CHECK, a trigger
WHEN clause, a domain CHECK, a column DEFAULT and a generated column are
covered too.

Only a bare literal is rewritten, which is what PostgreSQL does. It keeps a
negation the definition wrote around the test it rewrote, and it stores a cast
literal that way as well, so the two sides can differ by a NOT wrapper once the
text-like cast is stripped. Inverting a test under a NOT brings them back
together. A row operand is left alone: IS NULL on a row asks whether every
field is null rather than whether the row itself is, so neither test is the
other's negation there.


winebarrel:
Reviewed the change and ran it against a local PostgreSQL 16: `go test -p 1 ./...` is green, `go vet` is clean (`golangci-lint` could not run here, its build targets an older Go than the module). The rewrite sits in `normalizeExprNode`, so it is symmetric and reaches every expression, and the fixtures cover CHECK, index predicate and expression, view, policy, trigger, domain, DEFAULT and generated column. I verified each claim in the description against the catalog and they all hold: `NOT (a IS DISTINCT FROM NULL)` is stored as `NOT (a IS NOT NULL)`, a cast literal keeps the operator, and a `NULL::text` operand comes back as the negation once the text cast is stripped.

One thing that I think deserves a decision before merge, and two nits.

## `isRowOperand` misses a composite-typed column reference

`isRowOperand` catches a `RowExpr` and a whole-row `tbl.*`, which are the two forms that are recognizable in the raw parse tree. A column whose type is composite is a plain `ColumnRef` with no star, so the rewrite fires on it. PostgreSQL keeps the two apart there, because `ruleutils` prints a row-typed `NullTest` with `argisrow = false` back as the operator rather than as `IS NULL`:

```sql
CREATE TYPE cpt AS (x int, y int);
CREATE TABLE ct (p cpt);
ALTER TABLE ct ADD CONSTRAINT c1 CHECK (p IS NOT DISTINCT FROM NULL);
ALTER TABLE ct ADD CONSTRAINT c2 CHECK (p IS NULL);
ALTER TABLE ct ADD CONSTRAINT c3 CHECK (NOT (p IS NULL));
ALTER TABLE ct ADD CONSTRAINT c4 CHECK (p IS NOT NULL);

-- pg_get_constraintdef, PG16:
-- c1: CHECK ((p IS NOT DISTINCT FROM NULL))
-- c2: CHECK ((p IS NULL))
-- c3: CHECK ((NOT (p IS NULL)))
-- c4: CHECK ((p IS NOT NULL))
```

`c1` round trips fine, since both sides normalize the same way. What changes is that two genuinely different constraints now compare as one, so a real change is dropped from the plan. Both of these pass on `main` and fail on this branch:

```go
assert.False(t, equalConstraintDef(
    "CHECK ((p IS NULL))",
    "CHECK (p IS NOT DISTINCT FROM NULL)",
))
assert.False(t, equalConstraintDef(
    "CHECK ((NOT (p IS NULL)))",
    "CHECK (p IS NOT NULL)",
))
```

The normalization is syntactic and the type is not available there, so catching this properly is not on the table. A composite column tested against NULL is rare enough that accepting it looks reasonable to me under the corner-case policy. What I would not leave as is is the wording: the comments on `nullTestFromDistinct` and `foldNotNullTest`, the `isRowOperand` doc and the CHANGELOG all say a row operand is left alone, which reads as a semantic guarantee while the code only recognizes a syntactic row value. Narrowing that to a row value the parse tree shows, with a line saying a composite column is not one and what that costs, would keep the next reader out of the trap.

## `nullTestFromDistinct` does not need to run after `foldNotDistinct`

The comment says running after the fold picks up `NOT (a IS DISTINCT FROM NULL::text)`, "which arrives here as the operator once the cast is stripped". The walk visits children first, so that inner `A_Expr` is already rewritten to `a IS NOT NULL` at its own visit, and the `NOT` node reaches `foldNotNullTest` with a `NullTest` under it; `foldNotDistinct` never sees it. The chained path cannot fire in general either: the fold only changes the kind, so the second `nullTestFromDistinct` call looks at the same operands the child's call already rejected. I confirmed it by panicking on the chained rewrite and running the suite, which stays green. `node = folded` is therefore equivalent to the previous `return folded`; the comment is the part worth fixing.

## `TestEqualConstraintDef_distinctFromNullBothSidesKept`

The current side is spelled `CHECK ((NULL IS NULL))`, but the catalog emits `CHECK ((NULL::unknown IS NULL))` for `CHECK (NULL IS NOT DISTINCT FROM NULL)`. The guard itself is fine, and that constraint drifts on `main` too, so this is only about the fixture matching what PostgreSQL actually hands back.

---

winebarrel:
All three addressed in 537e125.

**Composite column.** Taking the corner-case reading, and narrowing the wording rather than the behavior. `isRowOperand` now says it recognizes a row value the parse tree shows, that a composite-typed column is a plain `ColumnRef` and goes through the rewrite, and what that costs: both sides normalize alike so nothing drifts, but `p IS NULL` and `p IS NOT DISTINCT FROM NULL` compare equal on such a column and a change between them is dropped from the plan. The `nullTestFromDistinct` and `foldNotNullTest` comments and the changelog say row constructor / `tbl.*` instead of row operand.

**The chained call.** You are right, and I confirmed it the same way: panicking when `nullTestFromDistinct` fires on a folded node leaves the suite green. The fold only changes the kind, so the second call sees the operands the child's own visit already rejected. Back to `return folded`, and the comment now points at `foldNotNullTest` for the stored negation instead of claiming the chain.

**Both-literal test.** Spelled `CHECK ((NULL::unknown IS NULL))` now.

`go test -p 1 ./...` green on PostgreSQL 16, `go vet` clean, coverage unchanged at 95.5% with the four new functions at 100%.

---

winebarrel:
Re-reviewed 537e125. All three are addressed and the code change is the one line back to `return folded`, which is the equivalence we both measured. `go test -p 1 ./...` green on PostgreSQL 16, `go vet` clean. I also confirmed the cost the new `isRowOperand` comment documents is real rather than theoretical, so the reader is being warned about something that can actually bite:

```
 p       | p IS NULL | p IS NOT DISTINCT FROM NULL
---------+-----------+-----------------------------
 <null>  | t         | t
 (,)     | t         | f
 (1,)    | f         | f
```

LGTM.

One correction to the mechanism the comments give, worth a follow-up rather than a change here, since the behavior they describe is right and only the "why" is off. PostgreSQL does not keep a row value out of the rewrite. `transformAExprDistinct` tests the operands before it transforms them and before it looks at any type:

```c
	if (exprIsNullConstant(rexpr))
		return make_nulltest_from_distinct(pstate, a, lexpr);
	if (exprIsNullConstant(lexpr))
		return make_nulltest_from_distinct(pstate, a, rexpr);
```

and `make_nulltest_from_distinct` sets `argisrow = false` unconditionally, with the comment `/* argisrow = false is correct whether or not arg is composite */`. A `ROW(...)`, a `tbl.*` and a composite column all become a scalar `NullTest` there. What keeps the two forms apart is the deparse: `get_rule_expr` prints `IS NULL` only when `ntest->argisrow || !type_is_rowtype(exprType(ntest->arg))`, and otherwise prints `IS NOT DISTINCT FROM NULL` to preserve the scalar meaning. That is why `pg_get_constraintdef` hands back the operator for every row-typed operand, composite column included.

Nothing follows for the code: the catalog never emits `IS NULL` on a row-typed argument, so skipping the rewrite on the row values the parse tree shows is exactly right. But the split matters for the composite-column note, which currently reads as PostgreSQL declining the rewrite by type when it is the deparse doing the work. Saying the rewrite is unconditional and the type check lives in the deparse would also explain, in one line, why the catalog side is always the operator there.

---

winebarrel:
You are right, and I took it now rather than as a follow-up since it is comment-only: 1944177.

I checked both halves before rewriting. `make_nulltest_from_distinct` sets `argisrow = false` with exactly the comment you quote, and `get_rule_expr` gates on `ntest->argisrow || !type_is_rowtype(exprType(...))` with a note that a rowtype input under a scalar test must print `IS NOT DISTINCT FROM NULL` to stay semantically correct. The split shows up in the catalog as it predicts, on a composite column and on a row constructor alike:

```
c1 | CHECK (p IS NOT DISTINCT FROM NULL)        -> CHECK ((p IS NOT DISTINCT FROM NULL))
c2 | CHECK (p IS NULL)                          -> CHECK ((p IS NULL))
c3 | CHECK (ROW(q, q) IS NOT DISTINCT FROM NULL)-> CHECK ((ROW(q, q) IS NOT DISTINCT FROM NULL))
c4 | CHECK (ROW(q, q) IS NULL)                  -> CHECK ((ROW(q, q) IS NULL))
c5 | CHECK (q IS NOT DISTINCT FROM NULL)        -> CHECK ((q IS NULL))
```

The `isRowOperand` doc now says the rewrite is unconditional and the type check is in the deparse, which is also the one-line reason the catalog side is always the operator on a row-typed argument. The composite-column note keeps the cost it had. `nullTestFromDistinct` points at `isRowOperand` for the same reason instead of arguing from row semantics; `foldNotNullTest` keeps its own reason, since what makes that inversion wrong on a row is the semantics rather than the deparse.

Suite green and `go vet` clean on PostgreSQL 16.

---

winebarrel:
Re-reviewed 1944177. Comments only, no behavior change; suite green on PostgreSQL 16 and `go vet` clean. The `isRowOperand` doc now has the mechanism right: the rewrite is unconditional and the type check lives in the deparse, `IS NULL` when `argisrow` is set or the argument is scalar and the operator otherwise.

The conclusion drawn from it went one step too far, though, and I would fix it before merge, because it says something the tests right below it contradict. A row-typed argument does not come back as the operator either way: a written `ROW(...) IS NULL` gets `argisrow = true` from `transformExprRecurse`, which is the first branch of that same deparse condition, so it comes back as `IS NULL`. PG16:

```sql
CREATE TABLE rr (a int, b int);
ALTER TABLE rr ADD CONSTRAINT r_op      CHECK (ROW(a,b) IS NOT DISTINCT FROM NULL);
ALTER TABLE rr ADD CONSTRAINT r_test    CHECK (ROW(a,b) IS NULL);
ALTER TABLE rr ADD CONSTRAINT r_notop   CHECK (ROW(a,b) IS DISTINCT FROM NULL);
ALTER TABLE rr ADD CONSTRAINT r_nottest CHECK (ROW(a,b) IS NOT NULL);
ALTER TABLE rr ADD CONSTRAINT r_star    CHECK (rr.* IS NULL);

-- pg_get_constraintdef:
-- r_op      -> CHECK ((ROW(a, b) IS NOT DISTINCT FROM NULL))
-- r_test    -> CHECK ((ROW(a, b) IS NULL))
-- r_notop   -> CHECK ((ROW(a, b) IS DISTINCT FROM NULL))
-- r_nottest -> CHECK ((ROW(a, b) IS NOT NULL))
-- r_star    -> CHECK ((rr.* IS NULL))
```

So the reason to skip a row value is the opposite of "either way": the catalog keeps the two spellings apart on a row-typed argument, the operator staying the operator and the row test staying `IS NULL`, and folding them together here would invent an equality the catalog never shows. That is exactly what `TestEqualConstraintDef_distinctFromNullRowKept` asserts, but its comment now reads "the catalog hands a row-typed argument back as the operator whichever test it stored, so the two forms must not compare equal", which argues for the opposite of what it asserts: if both spellings really came back as the operator, they would compare equal and the test would be wrong.

The same sentence is in the `nullTestFromDistinct` doc ("which the catalog hands back as the operator whichever test it stored") and in the changelog ("since the catalog hands a row-typed argument back as the operator either way"). Something like "the catalog keeps the two apart on a row-typed argument: the rewritten operator carries `argisrow = false` and prints as the operator, while a written row test carries `argisrow = true` and prints as `IS NULL`" covers all three. The `isRowOperand` paragraph above it already sets that up, so the last sentence just has to land on "apart" rather than "the operator either way".

Code-wise I have nothing left.

---

winebarrel:
Right, and my own measurements said so two comments up: `ROW(q, q) IS NULL` came back as `IS NULL` there while the operator stayed the operator. I wrote a conclusion the data in the same comment contradicts. Fixed in e6f2428, and I re-ran your five constraints to confirm:

```
r_op      | CHECK ((ROW(a, b) IS NOT DISTINCT FROM NULL))
r_test    | CHECK ((ROW(a, b) IS NULL))
r_notop   | CHECK ((ROW(a, b) IS DISTINCT FROM NULL))
r_nottest | CHECK ((ROW(a, b) IS NOT NULL))
r_star    | CHECK ((rr.* IS NULL))
```

The `isRowOperand` paragraph now lands on the split: the catalog keeps the two spellings apart on a row-typed argument, a rewritten operator printing as the operator and a written row test as `IS NULL`, so folding them together here would invent an equality it never shows. `nullTestFromDistinct` and the changelog say the same, and the test comment now argues for what the test asserts.

Suite green and `go vet` clean on PostgreSQL 16.

---

winebarrel:
Re-reviewed e6f2428. All four places now say the catalog keeps the two spellings apart on a row-typed argument, which matches both the deparse condition and what PG16 hands back, and the test comment argues for what the test asserts. Comments only again, so nothing to re-check behaviorally, but I ran it anyway: `go test -p 1 ./...` green on PostgreSQL 16, `go vet` clean.

LGTM.

---

PR #442, opened 2026-08-26
Manage column TOAST storage and compression

winebarrel:
Closes the `Column.StorageType` is dead code entry in TODO.md, by implementing it rather than dropping the field, and takes `attcompression` along since it is the same row, the same shape of statement and the same emission point.

## What was wrong

`pg_attribute.attstorage` and `attcompression` were read by neither side. `model.Column.StorageType` was declared and never written. On 16:

```
CREATE TABLE public.docs (id int, body text STORAGE external, note text COMPRESSION pglz);
```

- `pista dump` wrote `body text` and `note text`, losing both settings.
- An inline `STORAGE main` in a desired file was dropped without a warning, and the plan said `-- No changes`.
- The `ALTER TABLE ONLY ... SET STORAGE` a `pg_dump` file carries warned as an unsupported statement.

## What it does now

Both forms are read: the column definition (`body text STORAGE external COMPRESSION lz4`) and the separate `ALTER TABLE ... ALTER COLUMN ... SET STORAGE` / `SET COMPRESSION`, which `pg_dump` writes as `ALTER TABLE ONLY`. `dump` writes the separate form, which is what PostgreSQL 15 takes — a column definition accepts `STORAGE` only from 16 on — and what `pg_dump` does for the same reason. Neither statement rewrites the table; both decide how values inserted later are stored.

A definition that names neither asks for the defaults, so a column the database carries at `EXTERNAL` plans back to its type's own strategy.

## Three things that needed care

**Storage has no unset state.** `attstorage` always holds a strategy, so "never set" is only visible as it being equal to the type's own. The type's strategy is read alongside it and an unwritten one resolves against it, which is what keeps `body text STORAGE extended` on a text column from drifting. Resetting names the strategy rather than using `SET STORAGE DEFAULT`, which arrived in 16. A domain column reads its strategy through the domain, which carries the base type's.

**`SET DATA TYPE` resets both**, even when the type does not change, which a collation change goes through:

```
CREATE TABLE tc (a text STORAGE main COMPRESSION pglz);
ALTER TABLE tc ALTER COLUMN a SET DATA TYPE text COLLATE "C";
-- attstorage x, attcompression ''
```

`ATExecAlterColumnType` assigns `attstorage = tform->typstorage` and `attcompression = InvalidCompressionMethod` unconditionally, in 15 through 18. Without writing the settings out again, an apply drops them silently and the next plan re-emits them, so `plan --check` stays at exit code 2 until a second apply. `testdata/apply/column_storage_type_change.yml` covers a type change, a collation change and a column that leaves both out; its re-plan fails without the re-emit.

Under `--bulk-alter` the settings merge with a type change into one `ALTER TABLE`. That is safe in either order, since `AT_AlterColumnType` runs in `AT_PASS_ALTER_TYPE` and both settings in `AT_PASS_MISC`.

**A partition copies both off its parent as it is created**, and declares no columns of its own, so only the parent carries the statements. On a partition that already exists the two differ: `SET STORAGE` recurses through `ATSimpleRecursion`, `SET COMPRESSION` does not (`ATPrepCmd` carries `/* This command never recurses */` for it). On 16:

```
SET STORAGE MAIN     -> parent m, partition m
SET COMPRESSION pglz -> parent p, partition unset
```

So a method added to a table that already has partitions leaves them on the old one, and nothing re-emits it, since a partition child declares no columns and `diffTable` returns before `diffColumns`. Emitting per partition is more than this is worth, so TODO.md records it.

## Tests

Parser (inline, `DEFAULT`, the `ALTER TABLE ONLY` form, a column the file does not declare), dump (including a partitioned pair), plan (add, reset, no-diff including a definition that spells the type default out, `ADD COLUMN`, `CREATE TABLE`), apply (round trip, `--bulk-alter` with a merged type change, the type and collation changes), catalog (storage, compression, a domain column), unit tests for `StorageSQL` / `SetCompressionSQL` / `columnStorageSQL`, and one pinning that neither action reaches the ignored-action warning. `lz4` appears only in the parser fixture, since a server built without it rejects the method. Every setup writes `STORAGE` as its own statement, since a column definition takes it only from 16 on and the job for 15 fails otherwise.

Suite and `make test-scenario` green on PostgreSQL 16. Measured the way CI does, per package with no `-coverpkg`, coverage goes 94.588% -> 94.608%.


winebarrel:
Read the diff and checked the PostgreSQL side against `tablecmds.c` (16 and 18). The shape is right. Reading `typstorage` alongside `attstorage` is the only way to tell "never set" apart from "set to the type's own", naming the strategy instead of `SET STORAGE DEFAULT` is what keeps 15 working, and the re-emit after `SET DATA TYPE` matches `ATExecAlterColumnType`, which assigns `attTup->attstorage = tform->typstorage` and `attTup->attcompression = InvalidCompressionMethod` with no condition on the type moving. The `--bulk-alter` merge is safe for the reason given: `AT_PASS_ALTER_TYPE` is 1 and `AT_PASS_MISC` is 10, and both settings prep as MISC. `SET COMPRESSION default` resolves to `InvalidCompressionMethod` in `GetAttributeCompression`, so the catalog reads back as empty and the reset plans clean.

Two things.

**`SET COMPRESSION` does not recurse, `SET STORAGE` does.** `AT_SetStorage` calls `ATSimpleRecursion` in `ATPrepCmd`; `AT_SetCompression` carries `/* This command never recurses */` there instead, in 16 and still in 18. So "applying them there recurses" holds for storage only. A new partition is fine either way, since `MergeAttributes` copies both `attstorage` and `attcompression` off the parent attribute at creation, and the parent's statements come before the child's `CREATE` in the plan. What is missed is adding a compression method to a table that already has partitions: the parent row changes, the existing partitions keep their old method, and because `diffTable` returns before `diffColumns` for a partition child the next plan still reads clean. Same for a type change on a partitioned parent, where `SET DATA TYPE` recurses and resets the children but the compression re-emit does not follow it. Emitting per child looks like more than this is worth, but the comment on `StorageSQL` and the CHANGELOG paragraph both say the settings reach the partitions, and for compression they do not.

**Coverage.** Codecov reports 83.5% on the patch and -0.14% on the project, which is not noise here: CI collects `-coverprofile` per package with no `-coverpkg`, so anything exercised only by the root-level YAML fixtures counts as uncovered. Inside their own packages `model.SetStorageSQL` and `model.SetCompressionSQL` are at 0% and the four emitting lines of `columnStorageSQL` are at 0%. A `TestTable_StorageSQL` next to `TestTable_IdxSQL` / `TestTable_CommentSQL`, plus one `columnStorageSQL` case in `diff/tables_test.go`, covers it; the fixtures stay where they are.

Nits, take or leave:

- The two `LEFT JOIN LATERAL (SELECT CASE ...) ON true` in `ListColumnsByTable` are plain scalar expressions over `a` and `t`. Written inline in the select list they read the same and drop two joins; the `s` lateral above them earns its shape, these do not.
- `SET STORAGE EXTENDED` is upper case and `SET COMPRESSION default` lower, in the same plan.
- A new table or column whose definition names the type's own strategy gets a `SET STORAGE` it does not need. Harmless, and the desired side cannot know the type default, so probably the right trade.

Nothing blocking. LGTM once the partition wording is fixed.

---

winebarrel:
Both taken, in fcb52ff.

**`SET COMPRESSION` does not recurse.** You are right, and I had read `/* This command never recurses */` in `ATPrepCmd` while checking the pass order without registering what it meant. Confirmed on 16:

```
before               parent x, partition x
SET STORAGE MAIN     parent m, partition m
SET COMPRESSION pglz parent p, partition unset
a partition created afterwards: m, p
```

So the wording was wrong in the direction that matters. The `StorageSQL` comment and the changelog now say a partition copies both off the parent *as it is created*, and that `SET STORAGE` reaches the partitions that already exist while `SET COMPRESSION` stops at the parent. TODO.md records the gap, including the type-change case you point out, where `SET DATA TYPE` recurses and resets every partition's compression while the re-emit names the parent alone. Emitting per partition is more than this is worth, agreed.

**Coverage.** The measurement was mine, not noise: I used `-coverpkg=./...`, which credits the root fixtures to the package that owns the code, and CI does not. Per package, `SetStorageSQL` and `SetCompressionSQL` were at 0% and `columnStorageSQL` at 77.8%. `TestTable_StorageSQL` (with the empty and partition-child cases), `TestSetCompressionSQL_Default`, `TestTableToSQL_IncludesStorage` and `TestColumnStorageSQL` cover them; the fixtures are unchanged. Measured the way CI does, 94.588% -> 94.608%, and the four are at 100%.

**Nits.** The two lateral joins are gone, written in the select list. The case difference stays: the strategy is a keyword and the method is a name, which is also how `pg_dump` writes them. The needless `SET STORAGE` on a definition that spells the type default out stays too, for the reason you give.

---

winebarrel:
Re-read at fcb52ff. All three are addressed.

The partition wording is right now, in the CHANGELOG and on `StorageSQL` both, and recording the gap in TODO.md rather than walking the children in the column diff is the call I would have made: the entry says what is missed, why nothing catches it on the next run, and what closing it would cost. The `SET DATA TYPE` sentence is worth keeping in there, since that is the path that loses a method a user never touched.

The two laterals are gone and the CASEs read the same inline. I pulled the query text out of `columns.go` and put it through pg_query; it parses, and no `st.` / `ts.` reference is left behind.

Coverage is closed. In their own packages `SetStorageSQL`, `SetCompressionSQL`, `StorageSQL`, `TableToSQL` and `columnStorageSQL` all come out at 100% now, so the codecov drop should go with it. The `columnStorageSQL` table covers the four cases that matter, the reset pair included, and the partition-child case pins the early return that the dump fixture only shows the result of.

No DB here, so I ran parser, model and diff locally (green) and left catalog to CI, which is green on 15 through 18 with samples; only the windows job is still running. Nothing further from me. LGTM.

---

PR #443, opened 2026-08-28
Carry the same-table dependents through a column rename

winebarrel:
`-- pista:renamed-from` rewrites the column references in the same table's indexes, constraints and foreign keys before the comparison, and leaves the other three same-table dependents alone: triggers, policies and generated columns. PostgreSQL applies `RENAME COLUMN` to all of them, so none of them needs DDL of its own.

A trigger `UPDATE OF` list or `WHEN` expression, and a policy `USING` / `WITH CHECK` clause, read as a definition change, so the plan carried a redundant `CREATE OR REPLACE TRIGGER` or `ALTER POLICY`. The trigger statement takes a `SHARE ROW EXCLUSIVE` lock on the table, which holds writes off for as long as it waits; under `--with-tx` that lock is held to commit.

A stored generated expression was worse than noise. `diffColumns` reports an expression change as an error, since PostgreSQL has no `ALTER COLUMN ... SET GENERATED AS`, so a rename on a table carrying one failed the whole run.

```sql
-- before
ALTER TABLE public.products RENAME COLUMN qty TO quantity;
CREATE OR REPLACE TRIGGER products_audit AFTER UPDATE OF quantity ON public.products
    FOR EACH ROW WHEN (new.quantity IS DISTINCT FROM old.quantity) EXECUTE FUNCTION stamp();
ALTER POLICY docs_own ON public.docs USING (owner_name = current_user);
-- and, on a table with a generated column:
-- pista: error: failed to diff tables: column public.items.total:
--   cannot change GENERATED expression; DROP COLUMN + ADD COLUMN is required

-- after
ALTER TABLE public.products RENAME COLUMN qty TO quantity;
```

## Implementation

A trigger names columns in the `UPDATE OF` list and the `WHEN` expression. `NEW` and `OLD` are the only qualifiers a `WHEN` clause takes, so a two-field `ColumnRef` names a column there; `WalkExprColumnRefs` does not visit it, since it looks for the unqualified form an index or constraint expression carries. The arguments passed to the trigger function are left alone, as PostgreSQL does not rewrite them either.

A constraint trigger carries `UPDATE OF` and `WHEN` as well, and the deparse has to keep it one. Reading it back as a plain trigger would cross the line `CREATE OR REPLACE TRIGGER` refuses, and `diffTriggers` would fall back to a `DROP` and a `CREATE`.

A policy `USING` / `WITH CHECK` clause and a generated expression are stored bare and reference their own table unqualified, so `parseSelectExpr` and the existing `rewriteColumnRefsInExpr` cover them. The policy rewrite joins the others in `diffTable`; the generated one goes inside `diffColumns`, which runs before that clone is built. A plain `DEFAULT` is left alone, since it cannot reference a column.

In every case the rewritten text only reaches the comparison. `alterPolicySQL` and the trigger emitters build their statements off the desired side.

View triggers need nothing: `model.View` has no `Columns`, so no column rename reaches them.

## Tests

Six plan fixtures confirmed to fail without the change: the plain trigger case, a constraint trigger, a policy, a generated column, and, for each of triggers and policies, a rename alongside a change the desired schema does ask for. A seventh pins that a generated expression which also changes is still an error; it passes either way, so it guards the rewrite rather than proving it.

Three apply fixtures, which re-plan and require no drift, so the round trip is checked against a real server. Fourteen unit tests covering the rewrites, quoted identifiers, chained renames, the untouched function arguments and plain `DEFAULT`, and every fallback.

Run against PostgreSQL 15 and 18.

Measured the way CI does, coverage goes 94.608% -> 94.642%.

`make test`, `make test-scenario`, `make vet` and `make lint` all pass.

Closes the TODO.md entry for triggers. The remaining entry, for views and cross-table FKs, is updated to say what is now covered.


winebarrel:
LGTM.

The rewrite sits where the other three same-table rewrites do, has the same shape, and the same single-pass no-cascade property. Since `equalTriggerDef` parses and deparses both sides anyway, the extra round trip that every trigger definition takes when a rename is present is a no-op for triggers the rename does not touch, and the rewritten definition never reaches the emitted SQL, which comes off the desired side. Checked a few things locally against the branch:

* The constraint trigger keeps `CREATE CONSTRAINT TRIGGER` through parse/deparse, so it does not cross the line `diffTriggers` falls back to `DROP` and `CREATE` for. The fixture covers it.
* A composite reference such as `(new.addr).city` parses to an `A_Indirection` over a two-field `ColumnRef`, so it is rewritten too.
* Limiting the walk to a two-field `ColumnRef` misses nothing: `WalkExprColumnRefs` visits the one-field form, and an unqualified column reference is not valid in a `WHEN` clause to begin with. Deparse lowercases nothing that matters here, since `new` and `old` arrive lowercase from the parser and a quoted `"NEW"` would not resolve on the server either.

Two notes, neither blocking.

The nil guard on `rewriteColumnRefsInTriggers` is fine to keep, but the comment on it and the PR description are wrong about why: the catalog (`catalog/tables.go:156`) and the parser both initialize `Triggers` to an empty map, so a relation with no triggers carries an empty map rather than nil. `diffPolicies` states the convention the other way round, that callers always pass initialized maps and no nil guard is needed. So `_NilPassesThrough` covers a path `diffTable` cannot reach.

The larger one is for a later PR. `diffTable` now rewrites indexes, constraints, foreign keys and triggers, and still leaves two same-table dependents alone. A policy `USING` / `WITH CHECK` naming the renamed column is the same noise this PR removes, one redundant `ALTER POLICY` on the first plan. A stored generated column on the same table is worse than noise: the expression differs on the two sides, so `diffColumns` returns `cannot change GENERATED expression; DROP COLUMN + ADD COLUMN is required` (`diff/tables.go:321`) and the whole plan fails, even though PostgreSQL would have carried the expression through the rename. Worth a TODO.md entry at least, since the trigger one just went away.

---

winebarrel:
LGTM.

Both notes are addressed, and the nil guard one landed better than what I suggested: dropping the guard turned up the actual cause, `newTable` in the diff tests being the only caller that left `Triggers` unset, so the fix is in the helper rather than a guard standing in for it.

The two new rewrites read the same way as the four before them, and I checked the parts that decide whether a rewrite of the current side can leak into a plan:

* `alterPolicySQL` writes every clause off the desired side, and `diffColumns` skips the default branch when either side is generated, so neither deparsed text reaches an emitted statement. `exprChanged` goes through `equalSelectExpr`, which parses both sides, so the extra parse/deparse round trip on an expression the rename does not touch is a no-op.
* `rewriteColumnsInExprPtr` returns nil for nil, which matters beyond neatness: the recreate test at `diff/policies.go:139` reads a clause going from set to unset, and a rewrite that materialized an empty string there would have turned an `ALTER POLICY` into a `DROP` and a `CREATE`.
* Putting the generated rewrite inside `diffColumns` rather than next to the others is right, and recomputing the renames map there keeps it self-contained. The extra entry `collectColumnRenames` can carry for a directive whose rename has already been applied is a no-op on the expression side, since the old name is gone from it by then.

Worth recording that the set is now closed rather than merely longer: every expression-carrying member of `model.Table` is covered, `Comment` carries none, and `PartitionDef` is never compared at all, which has its own TODO entry. The remaining entry naming views and cross-table FKs is now accurate.

One thing I looked at and am happy to leave alone: the partition-child branch returns before `diffColumns`, so none of the six rewrites run there. A partition child cannot rename a column on its own, so there is nothing to carry, and the branch was already this way.

`go vet` and the non-DB unit tests pass here on `e0831eb`; CI is green on 15 through 18, with the windows job still running when I looked.

---

winebarrel:
One low-priority note, on the TODO.md wording rather than the code.

The entry now reads that the rewriter updates the same-table dependents, naming all six, which reads as covered. It is not covered on a partition child. A trigger or policy created directly on the child is in the model, since the catalog only excludes the clones the parent pushes down (`tgparentid = 0`, `catalog/triggers.go:52`) and policies are per-relation to begin with. But the child branch in `diffTable` diffs its policies and triggers and returns before the rewrite block, and would find nothing there anyway: a `PARTITION OF` child declares no columns, which is what `withInheritedColumns` exists for, so `collectColumnRenames` on its desired side is empty. Carrying the rename to a child needs the parent's map.

So a rename on the parent still leaves a redundant `CREATE OR REPLACE TRIGGER`, with the `SHARE ROW EXCLUSIVE` lock this PR is about, or an `ALTER POLICY`, on a child that has one of its own. The second run is clean, since `RENAME COLUMN` propagates to every descendant and PostgreSQL rewrites the child's definitions with it. Indexes and FKs on a child are in the same position and were before this PR, so nothing regressed; only the entry now claims more than holds. A line saying the rewrite does not reach a partition child would keep it accurate. Whether the rewrite should reach one is a separate question, and leaving it is reasonable for a shape this rare.

Two things I checked while looking at this, both fine. The drop ordering in `diffColumns` is a two-pass on `IsStoredGeneratedColumn()` and never reads the expression, so rewriting a generated expression on the current side cannot disturb a paired drop. And `alignCurrentCasts` strips current-only casts by walking against the desired tree, which a rename of a `String` node leaves structurally identical, so the comparison is unaffected; the only cost is that the exact-string fast path in `equalSelectExpr` no longer hits.

---

PR #444, opened 2026-08-28
Check the column refs in a DEFAULT or GENERATED expression

winebarrel:
`validateColumnRefs` reads the index, constraint and foreign-key definitions on a desired table and reports a name the table's columns do not carry. It does not read the two expressions a column carries itself, so the same mistake in one of those reaches the database and fails there.

```sql
CREATE TABLE public.items (
    id integer NOT NULL,
    -- pista:renamed-from qty
    quantity integer NOT NULL,
    total integer GENERATED ALWAYS AS (qty * 2) STORED
);
```

```
-- before: the plan is built and the ALTER fails on the server
-- after
pista: error: column qty referenced in the GENERATED expression on column total does not exist on table public.items
```

Closes the TODO.md entry for it.

## Implementation

Both expressions are stored as bare text rather than a statement, so the collector parses them behind a `SELECT`, the way the diff side reads a policy clause. The names then go through the existing `walkExprColumnRefs` and `reportMissing`, so the message and the aggregation match the three checks beside it.

A generated expression may reference only columns of its own table, so every name it carries belongs to the set being checked. A plain `DEFAULT` may reference no column at all, which PostgreSQL rejects on its own; the names it does carry are checked the same way rather than rejected outright. That leaves `DEFAULT upper(name)` for the server to reject, and catches `DEFAULT upper(nmae)` here.

The walk goes ahead of the index loop, so a report follows the order a `CREATE TABLE` declares things in. Deduplication stays per object, so the same missing name reached from two columns is reported once for each.

Only unqualified names are read, as they are for an index or a constraint. Checking this turned up a separate problem: PostgreSQL accepts a qualified reference in a generated expression and stores it stripped, so a desired schema that writes `GENERATED ALWAYS AS (items.qty * 2) STORED` fails with `cannot change GENERATED expression` on every plan with nothing changed. That belongs to the diff rather than the validator, so it gets a TODO.md entry.

## Tests

Two plan fixtures for the end-to-end path, and eight unit tests: each kind reporting a missing name, a valid expression, an expression naming no column at all (`nextval`, `CURRENT_DATE`), a quoted column name, per-expression dedup, per-column reporting, and an unparsable expression degrading to a no-op.

The risk here is a false positive failing a schema that was fine, so I ran the round trip over the 51 sample schemas from `make schema`, dumping and replanning each: 50 plan clean and the 51st is gitlab's existing `VALIDATE CONSTRAINT` drift. No schema newly fails to plan.

Measured the way CI does, coverage goes 94.642% -> 94.653%.

`make test` on PostgreSQL 15 and 18, `make test-scenario`, `make vet` and `make lint` all pass.


winebarrel:
LGTM.

The collector reuses the `SELECT <expr>` wrap, `walkExprColumnRefs` and `reportMissing` that the other three checks use, so the message shape, the per-object dedup and the aggregation all line up with what was there. Placing the loop ahead of the indexes keeps the report in `CREATE TABLE` order, and skipping partition children and ignored tables is inherited from the surrounding function rather than re-decided. The rename interaction is right: `rewriteColumnRefsInGenerated` rewrites the current side, so a desired expression still naming the old column is a real error and the new fixture is not fighting #443.

The risk is a false positive, so I probed the collector over about two dozen expressions of the kind a dump produces: `nextval(...)`, `'active'::status`, `ARRAY[]::text[]`, `EXTRACT(year FROM created_at)`, `xmlelement(name foo, bar)`, `to_char(created_at, 'YYYY')`, `pg_catalog.now()`, `((meta).total * 2)`, `(items.qty * 2)`, casts, `CASE`, `COALESCE`. Each one yields exactly the real column names and nothing else - type names, function names, the `xmlelement` name and a qualified reference all stay out, as the comment on `WalkExprColumnRefs` says they do. That matches the 51-schema round trip you ran. `go test ./parser -run ValidateColumnRefs` passes here.

Three minor things, none blocking:

1. `parser.collectColumnRefsInColumnExpr` and `diff.parseSelectExpr` (diff/tables.go:1483) now open with the same `pg_query.Parse("SELECT " + expr)`. Moving that wrap into `internal/pgast` would give both one place to hold the target-list shape, and the comment about walking the list rather than indexing it would live with it.

2. That comment, "Walking the list rather than indexing it keeps a shape that cannot occur from needing a guard of its own", took a couple of reads. Something like "the prefix gives exactly one target, and the loop avoids a guard for a shape that cannot occur" would say the same thing faster.

3. README "Column rename caveats" still lists only indexes, constraints and foreign keys as the dependents that must use the new name and error out at parse time. That went stale with #443 for triggers and policies, and this PR adds the generated expression to the same list. A line there would keep the docs with the behavior.

---

PR #447, opened 2026-08-28
Leave partition children unmanaged behind --skip-partition-child

winebarrel:
Where another tool owns the partitions, pg_partman for example, the desired
schema declares the partitioned parent alone, and every partition in the
database read as a table nothing declares: the plan carried a DROP TABLE for
each one. -E reaches a regularly numbered set, but nothing reaches a set whose
names carry no pattern, so the escape hatch ran out exactly where the children
are created on someone else's schedule.

The flag sits in FilterOptions next to --manage-routine, so it is read by dump,
plan and apply, and filterTables applies it to both sides of the diff: a
partition the database holds is not dropped, and one the desired schema
declares is not created. dump writes the parent without its partitions, which
keeps its output plannable under the same flag.

A partition is left out whether or not it is partitioned itself, so a
sub-partitioned level goes with the leaves under it. The parent stays managed
and PostgreSQL carries ADD COLUMN and an index created on it down to the
partitions, so the children do not go stale for being unmanaged. An INHERITS
child declares its own columns and carries no bound, so IsPartitionChild
separates the two and leaves it managed; the same predicate replaces the pair
of nil checks the diff and the comment parser already spelled out.

Measured the way CI does, 94.6% -> 94.7%.

Claude-Session: https://claude.ai/code/session_012GyB9Rhvc8MZSCoiAfFG9m

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,8 +67,8 @@ make fix            # golangci-lint run --fix (auto-fix lint errors)
 - Package-level tests generally use external test packages (e.g., `package catalog_test`, `package model_test`). Use same-package tests only when access to unexported identifiers is required (e.g., `package diff`).
 - Root-level integration tests use `package pistachio_test`.
 - Test fixtures are YAML files in `testdata/`. Required fields vary by test suite: `apply` uses `init`/`desired`/`applied`, `plan` uses `init`/`desired`/`plan`/`error`, `dump` uses `init`/`dump`, and `parser` uses `input`/`expected`. The plan/apply/dump harnesses also accept optional fields, but **the set differs per suite** (the lists below are not interchangeable):
-  - `dump`: `min_pg`, `omit_schema`, `sort_by_deps`, `manage_routine`, `include`/`exclude`/`enable`/`disable`, `dump_pg16`/`dump_pg17`/`dump_pg18`.
-  - `plan`: `min_pg`/`max_pg`, `count`, `drop_policy`, `disallowed_drops`, `ignored`, `disable_index_concurrently`, `force_index_concurrently`, `bulk_alter`, `assume_validated`, `manage_routine`, `include`/`exclude`/`enable`/`disable`, `pre_sql`/`pre_sql_file`/`concurrently_pre_sql`/`concurrently_pre_sql_file`.
+  - `dump`: `min_pg`, `omit_schema`, `sort_by_deps`, `manage_routine`, `skip_partition_child`, `include`/`exclude`/`enable`/`disable`, `dump_pg16`/`dump_pg17`/`dump_pg18`.
+  - `plan`: `min_pg`/`max_pg`, `count`, `drop_policy`, `disallowed_drops`, `ignored`, `disable_index_concurrently`, `force_index_concurrently`, `bulk_alter`, `assume_validated`, `manage_routine`, `skip_partition_child`, `include`/`exclude`/`enable`/`disable`, `pre_sql`/`pre_sql_file`/`concurrently_pre_sql`/`concurrently_pre_sql_file`.
   - `apply`: everything `plan` accepts other than `plan`/`error`, `max_pg` and `ignored`, plus `applied_sql`, `skip_drift_check` and `applied_pg16`/`applied_pg17`/`applied_pg18`.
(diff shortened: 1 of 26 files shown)


winebarrel:
LGTM.

I read the whole diff against `267be16`, built and vetted the branch, and ran the DB-free tests (`model`, `parser`, `internal/pgast`, `TestFilterTables_SkipPartitionChild`) locally; all pass. No PostgreSQL or Docker in this environment, so the plan/apply/dump fixtures and the scenario test ran only in CI, where 15/16/17/18, `samples`, `ascii` and CodeQL are green.

What convinced me:

- `filterTables` is the right seam. It is the one gate both `dump.go:534` and the two `diff_all.go` sites go through, so one predicate covers dump, the current side and the desired side without a second code path. Indexes, policies, RLS flags, triggers and comments live inside `model.Table`, so dropping the table entry takes them with it — `skip_partition_child_objects.yml` and `skip_partition_child_desired_objects.yml` pin exactly that, which is the part that would have been easy to miss.
- The bound is the correct discriminator. `pg_get_expr(c.relpartbound, c.oid)` is non-null for a declarative partition including a `DEFAULT` one, and null for an `INHERITS` child, so `IsPartitionChild` separates the two the way the help text claims. Folding the two spelled-out nil checks in `diff/tables.go` and `parseCommentStmt` into it is a clean consolidation, and the three remaining single `PartitionOf != nil` checks (`validate.go`, `toposort/schema.go`, `dump.go`) are deliberately INHERITS-inclusive and correctly left alone.
- The skip sits before `MatchName`, so `--include` cannot pull a partition back in; the unit test covers that case along with `--exclude` and `--disable table`.
- The round trip holds. `dump` writes the parent alone and scenario step 06 feeds that output straight back under the same flag for an empty plan, which is what the design policy asks for.
- `skip_partition_child_drop_parent.yml` is the fixture I most wanted to see: it pins that dropping the parent is still the parent's own statement and that PostgreSQL takes the unmanaged partitions with it, so the flag cannot leave orphans behind.
- Every apply fixture re-plans without `skip_drift_check`, so drift-freedom is asserted rather than assumed.

Two observations, neither blocking:

- A schema file that contradicts the database — declaring a relation as a plain table where the database holds it as a partition, or the reverse — falls out of the flag asymmetrically: one side is filtered, the other is not, so the plan carries a `CREATE TABLE` that would fail or a `DROP TABLE` (gated by `--allow-drop table`). That is a contradictory input under a flag the user opted into, so per the "do not chase corner cases" policy I would leave it exactly as is; noting it only so the behavior is on record.
- Nits: `CLAUDE.md` lines 70-72 enumerate the optional fixture fields per suite and do not yet list `skip_partition_child` (the text does point at the `*TestCase` structs as authoritative, so this is cosmetic). And in `skip_partition_child.test.sh`, steps 02-06 invoke `"$PISTA"` directly where `pista_plan --skip-partition-child ...` would work, since the helper forwards `"$@"` after `--allow-drop all`.

CHANGELOG, README prose and all three regenerated help blocks match the shipped flag. Nothing here needs changing before merge.

---

winebarrel:
Re-reviewed at `0757887`. LGTM, both nits are fixed and nothing else moved.

`ed5eabf..0757887` touches only `AGENTS.md` and `test/scenario/skip_partition_child.test.sh` — no Go changed, so the substantive review above still stands unchanged.

- `AGENTS.md` is the right file to edit; `CLAUDE.md` is a symlink to it, so my earlier reference to `CLAUDE.md` pointed at the same content. `skip_partition_child` now sits in the `dump` and `plan` lists, and the `apply` bullet inherits it through "everything `plan` accepts", which is accurate against `applyTestCase`.
- The scenario steps now go through `pista_plan` / `pista_apply`, and dropping the explicit `2>&1` is correct since both helpers already fold stderr in — no double redirect, and `--allow-drop all` still leads. Leaving steps 05 and 06's `dump` invocations direct is the right call; no helper wraps `dump`. `bash -n` is clean.

Re-verified locally: build, vet, and the DB-free tests (`model`, `parser`, `internal/pgast`, `TestFilterTables_SkipPartitionChild`) all pass. CI on this head is green across 15/16/17/18, `samples`, `keywords`, `ascii` and CodeQL, with only `windows` still running.

Ready to merge from my side.

---

PR #449, opened 2026-08-28
Emit policy renames in the desired schema's order

winebarrel:
`diffPolicies` read the renames out of `renamedFrom`, a Go map keyed by the new name, so the `ALTER POLICY ... RENAME TO` statements came out in a random order. One rename on a table cannot show it; two or more, and the same desired schema produced a different plan on each run. The statements are independent, so nothing broke, but a plan that is reviewed, or diffed against the previous run, has to be stable.

The renames now follow `desired`, which is an ordered map, the way `DiffViews` and `diffTriggers` already do. Rendering the statement moved to the caller, so `detectPolicyRenames` returns the rename map alone, and `diffPolicies` both suppresses the ones whose policy needs DROP+CREATE and keeps the rest in order. That also drops the needle matching the old code used to pair a statement with its entry.

## Tests

`TestDiffPolicies_RenameOrderFollowsDesired` repeats the diff, since a single run can line up by chance, and `testdata/plan/rls_rename_policies.yml` covers the same case end to end. Both fail on `main`: the unit test on every run, the fixture on about five runs in six.

`detectPolicyRenames` was reached only through `diffPolicies`, so its remaining branches are covered too: a directive naming the policy itself, a rename already applied, and a destination that already exists. It goes from 90.3% to 100%.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.714%. `make lint`, `make deadcode` and `make test` pass.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Emit the `ALTER POLICY ... RENAME TO` statements in the desired schema's order. The renames were read out of a Go map, whose iteration order is randomized, so a plan renaming two or more policies on one table printed them in a different order on every run. The statements are independent, so only the output moved, but a plan that is reviewed, or diffed against the previous run, has to be stable.
+
 ## [1.34.0] - 2026-08-28
 
 * Manage a partitioned table without its partitions, behind `--skip-partition-child` (`$PISTA_SKIP_PARTITION_CHILD`). Where another tool creates the partitions, pg_partman for example, a schema file that declares the parent alone planned a `DROP TABLE` for every partition, and `-E` only reached names that carry a pattern. With the flag `dump` writes the parent alone, and `plan` / `apply` skip a partition on both sides of the diff, including the indexes, policies and comments it owns. The parent stays managed, and PostgreSQL carries `ADD COLUMN` and an index created on it down to the partitions. A partition is skipped whether or not it is partitioned itself, so a sub-partitioned level goes with the leaves under it. An `INHERITS` child carries no partition bound and stays managed.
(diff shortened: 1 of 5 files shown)


winebarrel:
LGTM.

I traced the equivalence: `detectPolicyRenames` already appended its statements while iterating `desired.All()`, so `renameStmts` was ordered. The caller then threw that order away by iterating `renamedFrom` and matching a needle back into the slice, which is exactly where the randomness came from. Iterating `desired.Keys()` and rendering directly is the right shape, and it cannot lose a rename: every `renamedFrom` key is a `desired` key by construction, so the new loop visits all of them. The `needsRecreateRenamed` loop still walks the map, but it only builds a set, so its order does not matter.

Extracting `renamePolicySQL` and dropping the needle is worth it on its own. `strings.Contains` on a rendered statement is fragile: with `a` -> `b` and `xa` -> `by` on one table, the needle `a TO b` is a substring of the `xa TO by` statement.

That same pattern is still in `diffConstraints` (`diff/tables.go`, the loop over `renameStmts` with `strings.Contains(stmt, model.Ident(oldName)+" TO "+model.Ident(name))`). The order there is stable, since it iterates the ordered `renameStmts` and only tests membership, so this is not a bug in the same sense, but the substring match can suppress the wrong rename in the case above. Worth the same treatment in a follow-up, if you think it clears the corner-case bar.

Minor: this is the fifth open PR adding an `## [Unreleased]` heading at the top of CHANGELOG.md, so whichever of them merges after the first needs a trivial conflict resolution there.

---

winebarrel:
The `strings.Contains` point is fixed in #452 (91b4c41), which already reworks both functions.

It turned out to clear the corner-case bar on its own. The collision does not misorder the renames, it loses one: with `a` -> `b` (definition also changed) next to `xa` -> `by` (unchanged), the needle for the first suppresses the statement for the second and nothing replaces it, so `xa` keeps its name in the database and the same plan comes back next run. `diffForeignKeys` had the same hole.

---

PR #450, opened 2026-08-28
Filter the desired sequences by schema too

winebarrel:
`filterDesiredBySchemas` drops the objects a desired SQL file declares outside the schemas the run manages, so a file holding more than one schema can be planned one schema at a time. It named tables, views, enums, domains, composite types and routines, and left `ParseResult.Sequences` alone.

A sequence in another schema therefore reached the diff, where the current side is read from the target schemas alone and never holds it, so it read as new and the plan carried a `CREATE SEQUENCE` for a schema the run was not asked to touch. `apply` goes through the same `diffAll`, so it created it.

## Tests

`testdata/plan/desired_schema_filtered.yml` covered the case for a table alone. It now declares every object type outside the target schema. On `main` it fails with exactly one statement, the `CREATE SEQUENCE`, which is both the reproduction and the proof that the other six already worked.

`filterDesiredBySchemas` names each map explicitly, which is what let this one slip, so the doc comment now says what a missing name costs and why `ExecuteStmts` is not one.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.678%. `make lint`, `make deadcode` and `make test` pass.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Leave a sequence the desired schema declares outside the target schemas alone. The desired-side schema filter named every object type but the sequence, so a `CREATE SEQUENCE other.counter` sitting in a file planned with `-n public` was compared against a current side that never holds it, and the plan created it in a schema the run was not asked to manage. A table, view, enum, domain, composite type or routine in the same file was already dropped before the diff.
+
 ## [1.34.0] - 2026-08-28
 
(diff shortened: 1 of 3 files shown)


winebarrel:
LGTM.

I checked `parser.ParseResult` against the filter: it carries `Tables`, `Views`, `Enums`, `Domains`, `CompositeTypes`, `Sequences`, `Routines` and `ExecuteStmts`. With this change all seven object maps are named, so the list is now complete rather than complete-for-now, and the reasoning for leaving `ExecuteStmts` out is right, since raw SQL carries no schema to filter on.

The fixture is the part I like most. Declaring every object type outside the target schema turns `desired_schema_filtered.yml` into a guard against the next map going unnamed, not just a reproduction of this one, and `manage_routine: true` is needed for the routine to reach the diff at all, so that line is doing work.

One thought, no change requested: the doc comment tells the next reader what a missing name costs, but nothing fails if a new map is added to `ParseResult` and not to the filter. The fixture catches it only for object types someone remembers to add there. If this recurs, a reflect-based test asserting that every `*orderedmap.Map` field in `ParseResult` is emptied of a foreign schema would close it for good. Not worth it for one occurrence.

---

winebarrel:
Re-reviewed at `a809c1d`. Still LGTM.

The only new commit is the merge of `main`, and I diffed the two heads to confirm: everything it brings in is #449's content, and the sole conflict resolution is the CHANGELOG, where your bullet now sits under the `## [Unreleased]` heading that #449 landed. No code moved, so my earlier review stands unchanged.

---

PR #451, opened 2026-08-28
Walk the whole parse tree for a view's dependencies

winebarrel:
`collectRangeVars` was five levels of hand-written recursion over about a dozen node kinds. A relation named from a position none of them descended into was not seen, so the view got no edge and the topological order placed it wherever the name sort put it. Two new views in one run, one reading the other from such a position, were created in the wrong order and the apply failed on the relation that did not exist yet. A new view over an existing table was fine, since tables are created in an earlier phase.

The positions that went unseen are ordinary: a scalar sub-query under a cast, which `pg_get_viewdef` adds on its own, and one under `GREATEST`, an array or row constructor, a subscript, `ORDER BY`, `GROUP BY`, `LIMIT`, an `OVER` clause or an aggregate `FILTER`.

`pgast.Walk` already reaches every node kind through protoreflect, and its own comment names this failure mode, so the scan uses it and the recursion goes away. The visitor keeps the `RangeVar` rule the old one ended at, so a CTE name still resolves the way it did.

## Tests

`TestOrderFromSchema_ViewWithNestedSubquery` names one relation from each position that was missed. 12 of its 14 cases fail on `main`; the two that pass, `LATERAL` and `NOT EXISTS`, went through `RangeSubselect` and `BoolExpr`, which the old recursion did handle.

Deleting 59 well-covered statements would have dropped the percentage on its own, so the composite-type attribute edges, which had no test at all, get one. That leaves it above where it started.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.730%. `make lint`, `make deadcode`, `make test` and `make test-scenario` pass.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Order a view after every relation its body names. The dependency scan walked a hand-written list of node kinds, so a reference under a cast, `GREATEST`, an array or row constructor, a subscript, `ORDER BY`, `GROUP BY`, `LIMIT`, an `OVER` clause or an aggregate `FILTER` went unseen. Two new views in one run, one reading the other from such a position, were created in the wrong order and the apply failed on the relation that did not exist yet. The scan now uses the same walk the view comparison does, which reaches every node kind.
+
 ## [1.34.0] - 2026-08-28
 
 * Manage a partitioned table without its partitions, behind `--skip-partition-child` (`$PISTA_SKIP_PARTITION_CHILD`). Where another tool creates the partitions, pg_partman for example, a schema file that declares the parent alone planned a `DROP TABLE` for every partition, and `-E` only reached names that carry a pattern. With the flag `dump` writes the parent alone, and `plan` / `apply` skip a partition on both sides of the diff, including the indexes, policies and comments it owns. The parent stays managed, and PostgreSQL carries `ADD COLUMN` and an index created on it down to the partitions. A partition is skipped whether or not it is partitioned itself, so a sub-partitioned level goes with the leaves under it. An `INHERITS` child carries no partition bound and stays managed.
--- a/toposort/schema.go
+++ b/toposort/schema.go
@@ -7,6 +7,7 @@ import (
 
 	pg_query "github.com/pganalyze/pg_query_go/v6"
 	"github.com/winebarrel/orderedmap/v2"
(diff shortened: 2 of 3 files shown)


winebarrel:
LGTM. This is the best kind of diff: 120 lines of enumeration replaced by the traversal the repo already has, with the failure mode the enumeration produced turned into a table test.

Points I checked:

- `collectRangeVars` has no other caller, so nothing else loses coverage of those node kinds.
- `WalkOptions{}` leaves `SkipSubqueries` false, which is what a dependency scan wants. The one caller that would want it true is a per-table rule, and this is not one.
- The visitor returns the node unchanged in every path, so `Walk`'s rewrite machinery is inert here and the parse tree is not mutated. That matches `VisitFunc`'s contract about never returning nil.
- The CTE behavior is preserved for the reason you give: a CTE name reaches `qualifyRangeVar` as a `RangeVar` under the old code too, and `defined` is what filters it out, not the traversal.
- The walk now also reaches a `RangeVar` inside a data-modifying CTE, which the old recursion skipped. That is a real dependency, so seeing it is the right answer.

The test is well chosen. Naming the relation `z_defaults` so the name sort would put it after the view means each case can only pass on the edge, not by accident, and the two cases that pass on `main` are the two the old recursion genuinely handled, which is a good check that the fixture is measuring what it claims.

The protoreflect walk is slower per node than hand recursion, but it runs once per view definition when building the order, and the enumeration was returning wrong answers, so that is not a trade worth thinking about.

Backfilling the composite-type attribute edges is a fair way to hold the percentage, and that path having had no test at all is worth fixing regardless of this PR.

---

winebarrel:
Re-reviewed at `7f61701`. Still LGTM.

The only new commit is the merge of `main`, and diffing the two heads confirms it brings in nothing but #449's content, with the CHANGELOG as the sole conflict resolution: your bullet now sits under the `## [Unreleased]` heading #449 landed. `toposort/schema.go` and its test are untouched, so my earlier review stands.

---

winebarrel:
Re-reviewed at `e7f7794`. Still LGTM.

The only new commit is another merge of `main`, this time bringing in #450. Diffing against the head I last reviewed confirms the sole non-CHANGELOG content it adds is `schema_filter.go` and `desired_schema_filtered.yml` from that PR. `toposort/schema.go` and its test are untouched, so my review stands.

`go build ./...` and `go vet ./...` are clean on the merged head, and the CHANGELOG still carries exactly your one bullet on top of what main already has.

---

PR #452, opened 2026-08-28
Compare a constraint, foreign key and index definition once, and render their renames from the desired order

winebarrel:
Two changes to the same two functions. The first is what the PR opened with; the second came out of review and is a bug fix.

## Compare once

`diffConstraints` asked the same question in three places. The rename loop, the drop loop and the add loop each called `equalConstraintDef` on the same pair, and two of them called it twice to separate a NOT VALID transition from a real definition change. That is three parses per unchanged constraint and six per changed one. `diffForeignKeys` had the same shape around `equalFKDef`, and `diffIndexes` and `diffViewIndexes` called `equalIndexDef` twice per index.

Each comparison parses and deparses both definitions through pg_query, which is the most expensive thing a table diff does: 136us for a constraint, 91us for an index, 30us for a foreign key on the machine this was measured on.

The comparison now runs once per object and the loops read the answer. The two questions the old code re-derived are named: `changed`, meaning the object is dropped and added back, and `validateOnly`, meaning the definition matches and only the NOT VALID flag differs. Naming them also takes the double negatives out of the loop conditions, which is where the duplicate calls came from.

No output change. On a synthetic 500-table schema (9 columns, a primary key, a CHECK, two indexes and a foreign key each), `plan` output is byte-identical before and after, and runs a few percent faster. Reading the catalog dominates a plan, so the comparison is not where most of the time goes; the point is that two of the three loops in each function no longer parse at all.

## Render the renames from the desired order

Review of #449 and of this PR both pointed at the `strings.Contains` needle these functions still used. It is not only a style problem.

A renamed constraint whose definition also changes cannot use `RENAME`, so the statement is suppressed and the constraint is dropped and re-added under the new name instead. The suppression found the statement to drop by testing `strings.Contains` on the rendered SQL. With `a` renamed to `b` next to `xa` renamed to `by` on one table, the needle `a TO b` is a substring of `... RENAME CONSTRAINT xa TO by;`, so the second rename was suppressed as well and nothing replaced it. `xa` kept its name in the database and the same plan came back on the next run. `diffForeignKeys` was built the same way and had the same hole.

The fix is the shape #449 landed for policies. `detectConstraintRenames` and `detectForeignKeyRenames` return the rename map alone, the caller walks `desired`, which is ordered, and renders through one helper. Both emit `ALTER TABLE ... RENAME CONSTRAINT`, so the helper is shared. The dead `oldName != newName` guard above the destination check goes with it, the way the domain one does in #454.

## Tests

The compare-once half is a refactor, so the existing suite is the assertion; `diffConstraints`, `diffForeignKeys`, `constraintChanges`, `foreignKeyChanges`, `equalIndexDefs` and `newDefinitionChange` all report 100% statement coverage, so every branch it collapsed is exercised.

The rename fix has a unit test through each function and `testdata/plan/rename_constraint_with_recreated_peer.yml` end to end. The first fixture I wrote used `chk_a` / `chk_xa` and passed without the fix, since `chk_xa` does not contain `chk_a`; the committed one uses names that actually collide. `detectConstraintRenames` and `detectForeignKeyRenames` reach 100%.

Inverting the add-loop condition in `diffViewIndexes` made a gap visible: nothing covered a materialized view whose definition and indexes are both unchanged. `TestDiffViews_matviewIndexUnchanged` covers it.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.685%. `make lint`, `make deadcode`, `make test` and `make test-scenario` pass.


winebarrel:
LGTM. I worked through the truth table for each rewritten loop rather than trusting the description, and they are equivalent.

Constraint drop loop. Old: drop when `!ok || !sameDef || validatedDiff`, minus the `ok && sameDef && !cur.Validated && des.Validated` case. That is `!ok || (changed && !validateOnly)`. New: skip when `ok && (!ch.changed || ch.validateOnly)`, which is the same predicate. When `!ok`, `changes[name]` is the zero value, so `ch.changed` is false and the skip guard is short-circuited by `ok` before it is read. Same for the FK loop.

Constraint add loop. `validateOnly` implies `changed` (`sameDef && !curV && desV` forces `curV != desV`), so testing `ok && !ch.changed` before `ch.validateOnly` cannot swallow a VALIDATE, and `!ok` falls through to ADD with a zero-value `ch` as before.

`needsRecreation`. Iterating `renamedFrom` instead of `current` is right because every `renamedFrom` key is set by `detect*Renames` from a `desired` key and re-inserted into the adjusted current map, so the two loops range over the same names. It also reads better as "of the renames, which need recreating".

The one thing I wanted to confirm: `constraintChanges` and `foreignKeyChanges` are called after `current` has been reassigned to the rename-adjusted map, so a renamed object is compared under its new name, which is what the old inline comparisons did. Same for `equalIndexDefs` in both `diffIndexes` and `diffViewIndexes`.

Naming `changed` and `validateOnly` is the real win. The double negatives in the old conditions are what made it natural to re-derive the answer instead of hoisting it, so this should stay collapsed.

Two small things, neither blocking:

- `equalIndexDefs` returns `map[string]bool` while the other two return `map[string]definitionChange`. A missing key reading as false is safe only because every use is guarded by the `ok` from `GetOk`, which holds in all four loops today. Fine as is.
- The needle matching in the `diffConstraints` rename-emission loop is still there, a few lines above your change. #449 removes the same pattern for policies. `strings.Contains` on the rendered statement can match across renames (`a TO b` is a substring of `xa TO by`), so it may be worth the same treatment while you are in this function.

Your note about overlapping with #454 in `diff/tables.go` and `diff/views.go` is right; the two touch adjacent lines in the same drop loops, and the resolution either way is mechanical.

---

winebarrel:
Done in 91b4c41, and it clears the corner-case bar: the collision does not just misorder, it loses the rename outright.

`diffConstraints("public.t", ...)` with `a` -> `b` (definition also changed) next to `xa` -> `by` (unchanged) emitted only:

```
ALTER TABLE public.t DROP CONSTRAINT a;
ALTER TABLE public.t ADD CONSTRAINT b CHECK ((x > 1));
```

The `xa` -> `by` rename is suppressed by the needle and nothing replaces it, so `xa` keeps its name in the database and the same plan comes back on the next run. `diffForeignKeys` was the same.

`detectConstraintRenames` and `detectForeignKeyRenames` now return the rename map alone, and the caller walks `desired` and renders through one `renameConstraintSQL`, since both spell `ALTER TABLE ... RENAME CONSTRAINT`. The dead `oldName != newName` guard above the destination check went with it, matching what #454 does for domains. Both functions reach 100%.

On your other point: I left `equalIndexDefs` returning `map[string]bool`. There is no NOT VALID equivalent for an index, so a `definitionChange` there would carry one always-false field.

Tests: `TestDiffConstraints_RenameSuppressionKeepsOtherRenames`, the foreign-key twin, and `testdata/plan/rename_constraint_with_recreated_peer.yml` end to end. The first fixture I wrote used `chk_a` / `chk_xa` and passed without the fix, since `chk_xa` does not contain `chk_a`; the names in the committed one are chosen so the substring actually collides.

Inverting the add-loop condition in `diffViewIndexes` made a gap visible: nothing covered a materialized view whose definition and indexes are both unchanged. `TestDiffViews_matviewIndexUnchanged` covers it.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.685%, so uncovered statement count is now back to the baseline.

---

winebarrel:
Re-reviewed at `4a00d79`. LGTM.

The new commit turns what I flagged as an optional cleanup into the more valuable half of this PR, and you are right that it is worse than a style problem. I had only worked out that the substring match could suppress the wrong statement; the consequence you name is the part that matters, that `xa` keeps its old name, nothing replaces the suppressed statement, and the same plan comes back forever. That is a silent non-convergence, not a cosmetic slip, so it deserves the CHANGELOG entry it got.

What I checked on the new commit:

- `detectConstraintRenames` and `detectForeignKeyRenames` each have exactly one caller, so dropping the `[]string` return has nowhere else to land.
- The `oldName != newName` guard removed from both is dead for the same reason as the domain one in #454: the loop head does `if oldName == newName { continue }` about ten lines above. Confirmed in both functions.
- `renameConstraintSQL` being shared is right rather than convenient. PostgreSQL really does spell both as `ALTER TABLE ... RENAME CONSTRAINT`, and a foreign key is a constraint, so the two cannot drift apart.
- The emission loops match the shape #449 landed, and the resulting order is the desired schema's in both.
- `strings` is still used elsewhere in `diff/tables.go`, so the import stays.
- `go build ./...` and `go vet ./...` are clean on the merged head. I have no PostgreSQL here, so I did not run the suite; CI is green on 15 through 18, windows and samples.

The tests are the right ones. Picking `a` -> `b` alongside `xa` -> `by` so the needle is a genuine substring means the fixture would have caught this on `main` and will catch a regression, rather than just documenting it, and having both the unit tests and `rename_constraint_with_recreated_peer.yml` is worth it here because the two failures look different: the unit test shows the missing statement, the fixture shows the plan that never converges.

Adding `TestDiffViews_matviewIndexUnchanged` is a good catch. An index loop with no case for "nothing changed" is exactly the kind of gap an inverted condition makes visible, and it is worth having independently of this PR.

One nit, non-blocking: `detectForeignKeyRenames`'s doc comment still reads `Returns (renameStmts, adjustedCurrent, renamedFrom map[newName]oldName, error)`, but it no longer returns `renameStmts`. Its constraint counterpart got rewritten in the same commit, so this one just got missed.

The CHANGELOG merge came out right: your entry sits above #449's under the shared `## [Unreleased]`, and the refactor itself correctly has no entry since it changes no output.

---

winebarrel:
Re-reviewed the branch against current `main` and found one thing, fixed in 14312d8.

`detectForeignKeyRenames` stopped returning the rename statements in 91b4c41, and its doc comment kept naming them: `Returns (renameStmts, adjustedCurrent, renamedFrom map[newName]oldName, error)`. `detectConstraintRenames` changed the same way and had its comment rewritten; this one was missed.

The rest holds up. Walking the three loops again:

- Constraint drop loop: `!ok || (changed && !validateOnly)`. For a name absent from `desired`, `changes[name]` is the zero value, and `ok` short-circuits before it is read.
- Constraint add loop: `validateOnly` implies `changed`, so testing `ok && !ch.changed` first cannot swallow a VALIDATE; `!ok` falls through to ADD with a zero-value `ch`.
- Rename emission: `renamedFrom` keys are all `desired` keys, so walking `desired.Keys()` visits every one, and the ones `needsRecreation` holds are the ones the ADD covers under the new name.
- `dropName` still resolves through `renamedFrom`, so a renamed-and-recreated constraint is dropped by its old name, which is what the database still has.

`equalIndexDefs` and both `changes` builders run after `current` has been reassigned to the rename-adjusted map, so a renamed object is compared under its new name.

---

winebarrel:
Re-reviewed at `e0727bd`. LGTM, nothing outstanding.

`14312d8` fixes the stale `detectForeignKeyRenames` doc comment, and it goes one better than restating the signature: pointing at `detectConstraintRenames` as the sibling that leaves rendering to its caller is what a reader needs, since the two are now deliberately the same shape and the comment is the only thing saying so. That was my only open point on this PR.

The other new commit is the merge of `main`, which brings in #450. Diffing the heads confirms the only non-CHANGELOG content it adds is `schema_filter.go` and `desired_schema_filtered.yml` from that PR; nothing in `diff/` moved. Your bullet still sits alongside #449's and #450's under the shared `## [Unreleased]`.

`go build ./...` and `go vet ./...` are clean on the merged head. CI is green on 15 through 18, samples, CodeQL and the lint jobs; windows is still running as I write this.

This is ready from my side.

---

PR #453, opened 2026-08-28
Evaluate a plan-time execute check under the target schemas

winebarrel:
`apply` sets `search_path` to the target schemas plus `public` before it runs the managed DDL and the `-- pista:execute` statements. `plan` evaluated the same check SQL under the connection's `search_path`, which is `public` unless `--search-path` says otherwise.

So a check that names a table in a non-public target schema without qualifying it failed during `plan` with `relation does not exist`. `plan` treats an unevaluable check as undetermined and keeps the statement, with the reason recorded, while `apply` answered the same check and skipped the statement. The plan advertised a statement that never ran.

The `SET` the two commands need is now one method on `Client`, and `plan` issues it after `diffAll`, so the catalog has already been read and nothing about the diff can move. Only a run that carries a check pays for the statement, which keeps the round trip off every other plan.

## Tests

`TestPlan_ExecuteCheckUsesTargetSearchPath` plans a schema in `myschema` whose check reads `flags` unqualified. On `main` the plan keeps the statement with `relation "flags" does not exist`; with the change it is empty, matching what `apply` does. `TestPlan_ExecuteCheckUnevaluable` still passes, so a check that genuinely cannot be answered at plan time is still reported rather than failing the run.

`plan.go:117` is the one statement this adds without a test: the `SET` itself failing, which needs a broken connection to reach. `apply.go:157` is the same statement and is untested for the same reason. That is the whole of the coverage move.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.668%. `make lint`, `make deadcode` and `make test` pass.

---


winebarrel:
LGTM. The asymmetry this fixes is a real one: a plan that advertises a statement `apply` then skips is worse than either answer alone, because the reviewer has no way to tell which command is right.

Points I checked:

- The `SET` lands after `diffAll`, so the catalog read still happens under the connection's `search_path`. That matters here, since `connect` deliberately leaves the startup parameter at `public` so an object outside `public` stays qualified in the dump and in the diff. Moving the `SET` any earlier would have quietly changed that.
- Nothing else uses `conn` after the execute-statement evaluation, so the altered session state cannot reach anything but the checks.
- `plan` connects with `default_transaction_read_only=on`. `SET` is not a write, so it runs fine on that session.
- `hasCheckSQL` scans all of `ExecuteStmts` rather than one `First` group, which is right, since both groups are evaluated in the same run.

The one behavior worth naming out loud: when `--search-path` is given, the check now runs under the target schemas rather than under what the user asked for. That is the correct call, because `apply` already issues its own unconditional `SET` over the same startup parameter, so this makes `plan` match rather than diverge. It is just a thing someone reading the flag's documentation might not predict.

Folding the `SET` into `Client.searchPathSQL` is the right home for it, and the doc comment there is now the single place that explains why `public` is appended. Gating on `hasCheckSQL` keeps the round trip off every plan that has no check, which is most of them.

The test is the right one: it fails on `main` with `relation "flags" does not exist` recorded as the note, and `TestPlan_ExecuteCheckUnevaluable` still covering the genuinely unanswerable case is what keeps this from having quietly turned a diagnostic into a hard failure.

---

winebarrel:
Re-reviewed at `bac5c23`. Still LGTM.

The only new commit is the merge of `main`, and diffing the two heads confirms it brings in nothing but #449's content, with the CHANGELOG as the sole conflict resolution: your bullet now sits under the `## [Unreleased]` heading #449 landed. `plan.go`, `apply.go`, `client.go` and `directives.md` are untouched, so my earlier review stands.

---

winebarrel:
Re-reviewed at `f0030d9`. Still LGTM.

The only new commit is another merge of `main`, this time bringing in #450. Diffing against the head I last reviewed confirms the sole non-CHANGELOG content it adds is `schema_filter.go` and `desired_schema_filtered.yml` from that PR. `plan.go`, `apply.go`, `client.go` and `directives.md` are untouched, so my review stands.

`go build ./...` and `go vet ./...` are clean on the merged head, and the CHANGELOG still carries exactly your one bullet on top of what main already has.

---

PR #454, opened 2026-08-28
Stabilize a schema-map error message and drop two dead branches

winebarrel:
Three small things, none of which changes a plan.

`ValidateSchemaMap` walked `o.SchemaMap` to find two sources mapping to one destination. With three sources on one destination it named whichever pair the random map order reached first, so the same command printed a different rejection each time it ran. It walks the sources in name order now.

`detectDomainRenames` guarded its destination-exists check with `oldKey != newKey`, which the loop head has already established by continuing on the equal case. `detectCompositeTypeRenames` spells the same sequence without the guard, and that is the shape kept.

The drop loops in `diffIndexes` and `diffViewIndexes` initialised `useConcurrently` to `false` and then assigned it in both arms of the `if` that followed. The initial value is now the fallback the `else` arm held, and the `else` goes away. `desiredIdx` is nil when the index is a pure drop, so the fallback has to be the current side, not the desired one.

## Tests

The new `ValidateSchemaMap` case repeats the call, since one run can line up by chance; it fails on `main` most runs. The `useConcurrently` and domain-rename changes are behavior-neutral, so the existing tests are the assertion.

The domain rename directive has two branches that rename nothing, the directive naming the domain itself and a rename already applied. Only the second had a test; the first now has one too.

Measured the way CI does on PostgreSQL 16, 94.677% -> 94.694%. `make lint`, `make deadcode` and `make test` pass.

Note: this touches the same loops in `diff/tables.go` and `diff/views.go` as #452, a few lines apart. Whichever merges second may need a trivial rebase.

---


winebarrel:
LGTM. All three hold up.

`ValidateSchemaMap`. Sorting the sources fixes it, and with three or more sources on one destination the message now names the two lowest-sorting ones, which is stable and also the pair a user would most likely name themselves. The `len(o.SchemaMap) <= 1` early return keeps the sort off the common case. The repeated call in the test is the right way to assert this; a single run passes on `main` often enough to be useless.

`detectDomainRenames`. Confirmed the guard is dead: the loop already does `if oldKey == newKey { continue }` about ten lines above, so `oldKey != newKey` is invariant by the time the destination check runs. Matching `detectCompositeTypeRenames`, which spells the same sequence without it, is the right direction for the two to converge.

`useConcurrently`. Behavior-identical, and the initial value is now the fallback rather than a placeholder that every path overwrites. You are right that it has to be the current side: `desiredIdx` is nil on a pure drop, which is the only branch that reads the initial value at all.

Adding the same-name test for the domain rename directive is worth having on its own. Both of the branches that rename nothing now have one, and they are easy to break together because they look redundant to a reader who has not traced the loop.

The CHANGELOG entry covering only the first item is the right choice, since the other two produce no output change.

Your note about overlapping with #452 in `diff/tables.go` and `diff/views.go` is right; the two touch adjacent lines in the same drop loops and the resolution either way is mechanical. The `## [Unreleased]` heading will want the same treatment, since several of the open PRs add it at the same spot.

---

winebarrel:
Re-reviewed at `a8ef0e7`. Still LGTM.

The only new commit is the merge of `main`, and diffing the two heads confirms it brings in nothing but #449's content, with the CHANGELOG as the sole conflict resolution: your bullet now sits under the `## [Unreleased]` heading #449 landed. `options.go`, `diff/domains.go`, `diff/tables.go` and `diff/views.go` are untouched, so my earlier review stands.

Worth noting for merge order: #452 has since removed the dead `oldName != newName` guard from `detectConstraintRenames` and `detectForeignKeyRenames`, the same shape you remove from `detectDomainRenames` here, so the three rename detectors end up consistent whichever way round they land. #452 also still touches the `useConcurrently` lines in `diff/tables.go` and `diff/views.go`, so the conflict you flagged is still there and still mechanical.

---

winebarrel:
Re-reviewed at `089163b`. Still LGTM.

The only new commit is another merge of `main`, this time bringing in #450. Diffing against the head I last reviewed confirms the sole non-CHANGELOG content it adds is `schema_filter.go` and `desired_schema_filtered.yml` from that PR. `options.go`, `diff/domains.go`, `diff/tables.go` and `diff/views.go` are untouched, so my review stands.

`go build ./...` and `go vet ./...` are clean on the merged head, and CI is green on 15 through 18, samples, CodeQL and the lint jobs, with windows still running.

Merge-order note, unchanged from last time: #452 still touches the `useConcurrently` lines in `diff/tables.go` and `diff/views.go` that you simplify here, so whichever lands second needs the same mechanical resolution. #452 has also since removed the dead `oldName != newName` guard from both constraint rename detectors, matching what you do for `detectDomainRenames`, so the three end up consistent either way round.

---

PR #455, opened 2026-08-28
Compare a function call without its schema

winebarrel:
Picks up 362ba9d from the unmerged `claude/easy-fixes-fnjy5p`, plus a follow-up commit for what a review of it turned up.

## The fix

`pg_get_constraintdef`, `pg_get_expr`, `pg_get_indexdef`, `pg_get_viewdef` and `pg_get_triggerdef` print a call unqualified once the function's schema is on the search_path, while the desired side keeps what the file wrote. `stripFuncSchema` drops the schema from a `FuncCall` name inside `normalizeCheckExpr`, which reaches every site that walk covers, and `parseTriggerDef` does the same for the `EXECUTE FUNCTION` name.

Reproduced on `main` before applying anything: a CHECK, two indexes, a column default, a policy, a trigger and a view all re-emit on every plan, and a generated column fails the run with `cannot change GENERATED expression`.

## What the follow-up changes

**Coverage.** The cherry-pick alone measures 94.762% against main's 94.786%, a drop. `stripFuncSchema` reported 50% and `parseTriggerDef` 93.8%: the plan fixture reaches both, but it runs in the root package, where `diff` is a dependency rather than the package under test, so neither new line was counted. `TestStripFuncSchema` goes through each comparison entry point in turn, since a constraint, an index, a bare expression, a default and a view body each parse their own way in. Both functions reach 100%.

**A wrong claim.** The comment said a name longer than two parts is not one the parser accepts. It is: PostgreSQL takes `catalog.schema.f` when the catalog names the current database, and pg_query parses it to three parts, so the two-part test left it drifting. Keeping the last part alone handles both.

**A fixture that masked itself.** It claimed one site per statement kind so a failure names the one that drifted, but the generated column fails the diff outright and takes the run with it, so on an unfixed tree it was the only thing reported. It moves to its own fixture.

**Three sites that were missing.** A domain CHECK, a policy `WITH CHECK`, and a materialized view. All three drift without the fix. The materialized view is the worst case in the set: DROP plus CREATE discards the data it holds, on every apply.

**Prose.** `parseTriggerDef`'s comment had become one sentence of sixty words with two embedded `since` clauses; it is three sentences now. The change had no CHANGELOG entry; the one added names the matview and the generated column, since neither reads as ordinary drift.

## Known limitation, unchanged

The strip is symmetric, so two same-named functions in different schemas compare equal and moving a call between schemas produces no diff. This is the tradeoff a view body's table reference already carries. A schema inside a `nextval('public.counter'::regclass)` literal is not in a call name and still drifts; TODO.md is rewritten to cover that alone.

Measured the way CI does on PostgreSQL 16, 94.786% -> 94.789%. `make lint`, `make deadcode`, `make test` and `make test-scenario` pass.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* Stop a schema-qualified function call from drifting. `pg_get_constraintdef`, `pg_get_expr`, `pg_get_indexdef`, `pg_get_viewdef` and `pg_get_triggerdef` all print a call unqualified once the function's schema is on the search_path, while the desired side keeps what the file wrote, so `CHECK (public.lower_v(v) <> 'x')` and every other call written that way re-emitted its statement on every plan. A `CHECK` was dropped and re-added, which revalidates the whole table, an index was rebuilt, and a materialized view was dropped and recreated, discarding the data it held. A generated expression did not drift but failed the run outright, since it cannot be altered in place. The strip is symmetric and carries the tradeoff a view body's table reference already has: two same-named functions in different schemas compare equal. A schema inside a `nextval('public.counter'::regclass)` literal is not a call name and still drifts; TODO.md covers it.
+
 * Name the same pair every run when `-m` maps two schemas onto one destination. The check walked a Go map, so with three sources on one destination the rejection named a different pair each time the same command was run.
 
 * Stop one suppressed constraint rename from taking another with it. A renamed constraint whose definition also changes goes through `DROP` and `ADD` rather than `RENAME`, and the `RENAME` was suppressed by matching the rendered statement as a substring. With `a` renamed to `b` alongside `xa` renamed to `by` on one table, the needle for the first matched the statement for the second, so `xa` was left in the database under its old name and the same plan came back on the next run. Foreign keys were suppressed the same way and had the same hole.
--- a/TODO.md
+++ b/TODO.md
@@ -548,49 +548,34 @@ is the same prerequisite as the INHERITS plan / apply entry above.
 
 Origin: review of [#340](https://github.com/winebarrel/pistachio/pull/340).
 
-## Perpetual drift on a schema-qualified function call in a CHECK constraint
+## Perpetual drift on a schema-qualified sequence in a column DEFAULT
 
 Priority: low.
 
-A CHECK constraint that calls a function schema-qualified, written as
-`CHECK (public.lower_v(v) <> 'x')`, is re-emitted as `DROP CONSTRAINT` +
-`ADD CONSTRAINT` on every plan. Applying it succeeds and changes nothing, so
-`plan --check` stays at exit code 2 forever. `pg_get_constraintdef` returns
-the call unqualified when the function's schema is on the search_path, while
-the desired side keeps what the user wrote.
+A column `DEFAULT nextval('public.counter'::regclass)` is re-emitted as
+`ALTER COLUMN ... SET DEFAULT` on every plan. Applying it succeeds and changes
+nothing, so `plan --check` stays at exit code 2 forever. `pg_get_expr` returns
+the sequence unqualified when its schema is on the search_path, while the
+desired side keeps what the user wrote.
 
-`normalizeCheckExpr` (`diff/tables.go`) strips text-like casts and
-canonicalises `= ANY(ARRAY[...])` -> `IN (...)`, but never touches the
-`FuncCall` name, so the two forms deparse differently. Index expressions
-and index predicates run through `normalizeCheckExpr` as well
-(`normalizeIndexStmt`), so the same drift is expected there. A view body
-that calls a function qualified drifts the same way: `stripQualifications`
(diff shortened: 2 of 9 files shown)


winebarrel:
LGTM.

Reviewed the code against `main` and ran the new unit tests locally (no PostgreSQL here, so the plan fixtures rest on CI, which is green on 15-18 plus `samples`).

What I checked:

* **No leak into emitted DDL.** Every caller of `normalizeCheckExpr` (`equalConstraintDef`, `normalizeIndexStmt`, `equalDefault`, `equalSelectExpr`, `equalViewDef`, `equalPolicy*`) parses a fresh tree from a string and only deparses it for the comparison, and `parseTriggerDef`'s result is used the same way. `renameTriggerDef` re-parses from the original definition, so the stripped `Funcname` never reaches a generated statement. The DDL still carries what the file wrote.
* **Walk coverage.** Confirmed the strip reaches a call in a set-returning `FROM`, a `LATERAL`, a CTE, an aggregate argument and a window `ORDER BY`, not just the sites the fixture lists.
* **The three-part-name correction is right.** `postgres.public.f(v)` parses to three parts and the last-part-only strip handles it; the negative cases (`lower_v` vs `upper_v`, a different trigger function) still compare unequal.
* **Splitting the generated-column fixture** is the right call: it fails the diff rather than drifting, so in one file it would have hidden every other site on an unfixed tree.

Two notes, neither blocking:

1. Nit: `parseTriggerDef` inlines the same last-part-wins strip that `stripFuncSchema` performs. A shared helper over `[]*pg_query.Node` would keep the rule and its reasoning in one place, and the trigger comment already has to point at `stripFuncSchema` to explain itself.
2. Observation for TODO.md, not this PR: an `EXCLUDE` constraint is the one expression site `normalizeCheckExpr` still does not reach. `equalConstraintDef` normalizes `RawExpr` only, and an exclusion's elements and its `WHERE` clause live in `Exclusions` / `WhereClause`, so `EXCLUDE USING gist (public.lower_v(v) WITH =)` still drifts. This is pre-existing and wider than the schema question - a bare `::text` difference in an exclusion predicate drifts too - and `dump` writes the catalog form, so the round trip is unaffected. Low priority by the same standard as the `nextval` literal.

---

winebarrel:
Both points addressed in f380780.

**Nit 1.** `lastNamePart` now holds the last-part-wins rule and the reason for it, and both callers are one line. You were right that the trigger comment pointing at `stripFuncSchema` to explain itself was the tell.

**Nit 2.** Confirmed against a database before writing it down, and both halves hold.

```
ALTER TABLE public.t DROP CONSTRAINT t_excl;
ALTER TABLE public.t ADD CONSTRAINT t_excl EXCLUDE (public.lower_v(v) WITH =);
```

and, with no function in sight, `((v || 'x') WITH =)` against the catalog's `((v || 'x'::text) WITH =)`:

```
ALTER TABLE public.t DROP CONSTRAINT t_excl;
ALTER TABLE public.t ADD CONSTRAINT t_excl EXCLUDE ((v || 'x') WITH =);
```

So it is the whole normalization that is missing there, not the schema strip, exactly as you said. An unqualified exclusion with no cast plans clean, which is why nothing caught it.

Left out of the fix per your call. TODO.md records it, including that the elements are `IndexElem` nodes `normalizeIndexStmt` already handles, so whoever picks it up has the shape.

Measured the way CI does on PostgreSQL 16, 94.786% -> 94.791%. `lastNamePart`, `stripFuncSchema` and `parseTriggerDef` are all at 100%.

---

winebarrel:
Re-reviewed f380780. Both points are addressed, and the fix still LGTM.

* `lastNamePart` holds the rule and its reason in one place, and the guard is `len(name) < 2` rather than the caller's `> 1`, so a nil or single-part name returns unchanged and both call sites read as one line. The trigger comment no longer has to borrow `stripFuncSchema`'s explanation. Behavior is identical: `go build`, `go vet ./diff/` and the `diff` unit tests pass here on the new head. (CI is still running on this head; the previous head was green on 15-18.)
* The TODO.md entry matches what I saw: `Exclusions` and `WhereClause` bypass `normalizeCheckExpr` entirely, so the cast half drifts alongside the schema half, and pointing at `normalizeIndexStmt` for the `IndexElem` shape is the right note to leave for whoever picks it up. Recording it rather than fixing it here is the correct scope call.

One trivial prose nit, take it or leave it: `lastNamePart`'s comment ends with "never reaches the walk below", but `normalizeCheckExpr` and `normalizeExprNode` sit above it in the file. "the walk above" reads right.

---

winebarrel:
Taken, in 9af8dfa. `normalizeCheckExpr` and `normalizeExprNode` are at 566 and 570, `lastNamePart` at 606, so "below" was pointing the reader away from the thing it names. It reads "above" now.

Nothing else outstanding from my side.

---

winebarrel:
Re-reviewed 9af8dfa. It is one word in a comment ("below" -> "above") and nothing else, so the code is unchanged from f380780, which I already went through. LGTM.

---

PR #458, opened 2026-08-28
Manage a table's storage parameters, and stop an index's from drifting

winebarrel:
## A table's storage parameters were read by neither side

`WITH (fillfactor = 70, autovacuum_enabled = off)` was dropped from the desired schema without even the `ignored unsupported statement` warning an unsupported statement gets, `dump` did not write the clause, and no plan ever proposed one. pg_dump writes those parameters, and a pg_dump file is the natural starting point for a desired schema, so per-table autovacuum tuning went silently unmanaged the moment it was adopted.

Both sides read them now, from `pg_class.reloptions` and from the `CREATE TABLE`. A change goes out as `ALTER TABLE ... SET (...)`, with one `RESET` for the parameters the desired schema no longer names. Neither statement rewrites the table.

**TOAST.** A `toast.` parameter sits on the TOAST relation, which is where pg_dump writes it into the same clause. It is read from `reltoastrelid` and diffed alongside. PostgreSQL creates that relation only for a table with a toastable column and accepts the statement either way, discarding it when there is none, so such a table re-plans the parameter on every run. Documented in the README rather than special-cased.

**Partitioning.** A partitioned table takes no parameter at all, since it holds no storage, and a partition does not copy the parent's, so each is diffed on its own.

**Order.** Both sides are keyed in name order, so neither the order a file lists them in nor the order the catalog keeps them in reaches the comparison or the dump.

## An index's parameters were compared but not normalized

`pg_get_indexdef` quotes a value that does not read as an identifier, `fillfactor='80'`, while a file writes the number bare, so the two arrived as a String and an Integer node and never matched. An index carrying one was dropped and recreated on every plan, which is a full rebuild each run. Reproduced on `main`: apply the index, and the next two plans both emit `DROP INDEX` plus `CREATE INDEX` against a catalog that already holds `{fillfactor=80}`.

The catalog also keeps the parameters in creation order, which a file has no reason to follow, so the list is sorted before the comparison too.

Neither of these showed up in the dump round trip, since `dump` writes what `pg_get_indexdef` reports.

## Tests

14 plan fixtures, 3 apply, 2 dump, 1 parser, and unit tests in `model`, `diff` and `catalog`. The interactions are pinned as well: `--bulk-alter` leaves the `SET` on its own, `--skip-partition-child` drops a partition's parameters with the partition, and a rename carries them through and sets a changed one under the new name.

The `diff`, `parser` and `catalog` code needed package-local tests, not only fixtures. Measured the way CI does, the fixtures alone left the branch at 94.3% against main's 94.8%, because a fixture runs in the root package where those three are dependencies rather than the package under test.

## Left alone

Views and materialized views carry reloptions of their own, `security_barrier` and the autovacuum settings, and lose them the same way. A constraint's index takes them too, and `pg_get_constraintdef` does not print them, so a hand-written `UNIQUE (v) WITH (fillfactor = 70)` drops and re-adds the constraint on every run, rebuilding the index under an ACCESS EXCLUSIVE lock. Both are separate objects and separate statements. The constraint one is recorded in TODO.md with the shape of the fix.

`ALTER TABLE ... SET (...)` in a desired file is still warned about and dropped, like every other unsupported `ALTER TABLE` action. pg_dump writes the parameters inline in the `CREATE TABLE`, and so does `dump`, so nothing that round-trips depends on it.

Measured the way CI does on PostgreSQL 15, 94.8% -> 94.9%. `make lint`, `make deadcode`, `make test` and `make test-scenario` pass, and the 17 real-world sample schemas still round-trip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [1.36.0] - 2026-08-29
+
+* Manage a table's storage parameters, the `WITH (...)` clause. Neither side read them, so `dump` dropped the clause and no plan proposed one: the per-table autovacuum settings a pg_dump file carries went unmanaged without a warning. A change goes out as `ALTER TABLE ... SET (...)` with a `RESET` for the parameters the desired schema no longer names. A `toast.` parameter is read from the TOAST relation and diffed alongside. `WITH (oids = false)`, which a pg_dump before 12 writes, is not a parameter and is skipped.
+
+* Stop an index's storage parameters from drifting. `pg_get_indexdef` quotes a value that does not read as an identifier, `fillfactor='80'`, while a file writes the number bare, so the two never matched and the index was dropped and recreated on every run. The order the catalog keeps them in is not compared either.
+
 ## [1.35.0] - 2026-08-29
 
 * Stop a schema-qualified function call from drifting. `pg_get_constraintdef`, `pg_get_expr`, `pg_get_indexdef`, `pg_get_viewdef` and `pg_get_triggerdef` all print a call unqualified once the function's schema is on the search_path, while the desired side keeps what the file wrote, so `CHECK (public.lower_v(v) <> 'x')` and every other call written that way re-emitted its statement on every plan. A `CHECK` was dropped and re-added, which revalidates the whole table, an index was rebuilt, and a materialized view was dropped and recreated, discarding the data it held. A generated expression did not drift but failed the run outright, since it cannot be altered in place. The strip is symmetric and carries the tradeoff a view body's table reference already has: two same-named functions in different schemas compare equal. A schema inside a `nextval('public.counter'::regclass)` literal is not a call name and still drifts; TODO.md covers it.
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ pista apply ./schema/*.sql         # apply it
 - Composite types (`CREATE TYPE ... AS (...)`, `ALTER TYPE ... ADD/DROP/ALTER ATTRIBUTE`, `RENAME ATTRIBUTE`). Attributes are matched by name, so reordering them produces no diff (PostgreSQL cannot reorder attributes). `ALTER ATTRIBUTE ... TYPE` fails at apply while a table column uses the type; PostgreSQL does not allow it and `CASCADE` does not help.
 - Sequences (`CREATE SEQUENCE`, `ALTER SEQUENCE`, `DROP SEQUENCE`). Only standalone sequences are managed; sequences owned by a serial or identity column are handled as part of that column, not as separate objects.
 - Tables (including unlogged and partitioned tables)
+- Storage parameters, the `WITH (...)` clause on a table and on an index. A `toast.` parameter belongs to the TOAST relation. PostgreSQL creates that relation only for a table with a toastable column, and discards the setting when there is none, so such a table re-plans it on every run. A partitioned table holds no parameter, and a partition does not inherit the parent's.
 - Views
 - Materialized views
 - Columns (serial/bigserial/smallserial, identity, generated, TOAST storage and compression)
--- a/TODO.md
+++ b/TODO.md
@@ -548,6 +548,32 @@ is the same prerequisite as the INHERITS plan / apply entry above.
 
 Origin: review of [#340](https://github.com/winebarrel/pistachio/pull/340).
 
+## A constraint's index storage parameters are not read
+
+Priority: low.
+
+`ALTER TABLE t ADD CONSTRAINT t_v_key UNIQUE (v) WITH (fillfactor = 70)` puts
+the parameters on the index the constraint owns. `pg_get_constraintdef` does
+not print them, so the current side reads the constraint as `UNIQUE (v)` and
+the two never match: the plan is `DROP CONSTRAINT` + `ADD CONSTRAINT` on every
+run, which rebuilds the index under an ACCESS EXCLUSIVE lock. `dump` drops the
+clause for the same reason, so a table's parameters survive a dump and a
+constraint's do not.
+
+The parameters are on `pg_class` under `pg_constraint.conindid`, next to the
+ones `ListIndexes` already reads, and appending them to the definition is what
+would make both sides agree. `normalizeStorageParams` then folds the quoting
+the way it does for a plain index.
+
+`dump` writes the catalog form, so a dump fed back plans clean and only a
+hand-written constraint reaches this.
+
+Workaround: create the index separately, `CREATE UNIQUE INDEX ... WITH
+(fillfactor = 70)`, where the parameters are read.
+
+Origin: the storage parameter work; found while auditing what else carries a
+`WITH` clause.
+
 ## An EXCLUDE constraint is not normalized at all
 
 Priority: low.
--- a/TODO.md
+++ b/TODO.md
@@ -548,6 +548,27 @@ is the same prerequisite as the INHERITS plan / apply entry above.
 
 Origin: review of [#340](https://github.com/winebarrel/pistachio/pull/340).
 
+## View and materialized view storage parameters are not read
+
+Priority: low.
+
+A view carries `security_barrier` and `check_option`, a materialized view the
+autovacuum settings, and neither side reads them. `dump` drops the clause, and
+a desired schema that writes it plans nothing, the way a table's parameters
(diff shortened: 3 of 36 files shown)


winebarrel:
Reviewed the branch and ran it against a local PostgreSQL 16. `go vet` and `go test -p 1 ./...` pass.

The shape of the change is right. Reading both sides from `pg_class.reloptions` and the `CREATE TABLE`, keying in name order, and splitting `SET` from `RESET` all match how the rest of the diff works, and the index normalization is correct for the reason it claims: `ruleutils.c` leaves a value bare when `quote_identifier(value) == value`, so `deduplicate_items=off` arrives unquoted and `fillfactor='80'` quoted, and folding the Integer is exactly what makes the two spellings meet.

I confirmed several claims in the description hold: the `toast.` re-plan on a table with no TOAST relation, the constraint churn the new TODO entry describes, and the clean dump round trip for a table plus an index. A spelling difference such as `on` in the file against `true` in the catalog costs one `SET` and then converges, so that is not a loop.

Three things.

### 1. `WITH (oids = false)` now fails at apply

Every pg_dump from PostgreSQL 8.x through 11 writes this on every table, and such a file is the starting point the description has in mind. PostgreSQL still accepts the clause in `CREATE TABLE` and stores nothing for it:

```
postgres=# CREATE TABLE public.t7 (id int) WITH (oids = false);
CREATE TABLE
postgres=# SELECT reloptions FROM pg_class WHERE relname = 't7';
 reloptions
------------
 (null)
```

The parser now records it, the catalog can never report it, so the plan is a permanent `SET` that PostgreSQL rejects:

```
$ pista plan ./legacy.sql
ALTER TABLE public.t6 SET (oids='false');

postgres=# ALTER TABLE public.t6 SET (oids='false');
ERROR:  unrecognized parameter "oids"
```

On `main` the clause was dropped and the plan was clean, so this turns a working input into an apply failure on every table of an older dump. `oids = true` is rejected by `CREATE TABLE` itself, so skipping the name in `parseStorageParams` is enough and costs one condition:

```go
// WITH (oids = false) is what every pre-12 pg_dump writes. PostgreSQL still
// accepts it in CREATE TABLE and stores nothing for it, and ALTER TABLE ...
// SET rejects it outright, so it is not a parameter.
if name == "oids" {
    continue
}
```

With that the plan is clean again.

### 2. The `SET` splits a `--bulk-alter` group

`diffStorageParams` is appended in the middle of `result.Stmts`, so it lands between the column adds and the column drops, which `mergeAlterTable` groups only while they are consecutive. Keeping the `SET` out of the merged statement is right, but it also breaks the merge around it:

```
-- storage parameter unchanged
ALTER TABLE public.t4
  ADD COLUMN name text,
  DROP COLUMN old_col;

-- fillfactor 70 -> 90, same column changes
ALTER TABLE public.t4 ADD COLUMN name text;
ALTER TABLE public.t4 SET (fillfactor='90');
ALTER TABLE public.t4 DROP COLUMN old_col;
```

Three statements and three lock acquisitions where `--bulk-alter` asks for one. `bulk_alter_table_storage_params.yml` does not catch it because it only adds columns. Appending `diffStorageParams` after `colDropStmts` instead, at the end of `diffTable`, fixes it: nothing in `SET` / `RESET` depends on the column statements, and every fixture still passes, including `rename_table_storage_params.yml`. If the current placement is deliberate, say so in the comment, since I could not tell from the code why the `SET` sits where it does.

### 3. The view gap is only in the description

The description names views and materialized views, but only the constraint case reached TODO.md. `security_barrier` is dropped by `dump` today with nothing recording it:

```
postgres=# SELECT reloptions FROM pg_class WHERE relname = 'v5';
 {security_barrier=true}

$ pista dump
CREATE OR REPLACE VIEW public.v5 AS
SELECT id, v FROM t5;
```

The round trip plans clean, so it is not a bug by the design policy, but a dump silently losing `security_barrier` is worth the same TODO entry the constraint got, more so than the constraint one given what the option does.

None of this touches the table and index work itself, which reads correctly to me.

---

winebarrel:
Thanks, all three reproduce. Fixed in c9e821b.

### 1. `WITH (oids = false)`

Confirmed on 15: `CREATE TABLE` accepts the clause and `pg_class.reloptions` stays null, so the plan was a permanent `ALTER TABLE ... SET (oids='false')` and PostgreSQL answers `unrecognized parameter "oids"`. A real regression against `main`, and on every table of a pre-12 dump. `parseStorageParams` skips the name, with your comment. `oids = true` is rejected by the `CREATE` itself, so nothing else needs reading. `testdata/plan/table_storage_params_oids.yml` covers the plan; the skip also needed `testdata/parser/table_storage_params_oids.yml`, since a plan fixture runs in the root package where `parser` is a dependency rather than the package under test.

### 2. The `SET` splitting a `--bulk-alter` group

Reproduced exactly as you have it, three statements against one. Not deliberate, so no comment defending it: `diffStorageParams` now runs after `colDropStmts`, at the end of `diffTable`, and after `diffComments` in the partition-child branch.

Moving it turned up a second reason for the placement. A `toast.` parameter needs the TOAST relation, which PostgreSQL creates when the first toastable column is added, so the `SET` has to stay after `ADD COLUMN`:

```
postgres=# CREATE TABLE tc (id int);
postgres=# SELECT reltoastrelid FROM pg_class WHERE relname='tc';
 0
postgres=# ALTER TABLE tc ADD COLUMN body text;
postgres=# SELECT reltoastrelid <> 0 FROM pg_class WHERE relname='tc';
 t
```

The end of `diffTable` satisfies both. `DROP COLUMN` keeps the relation, so the drops in between change nothing. The comment says both reasons, since either one alone would let the next reader move it back.

`bulk_alter_table_storage_params.yml` drops a column now, which is what it should have done to begin with, and `apply/table_storage_params_new_toast_column.yml` pins the ordering through an apply and a re-plan.

### 3. The view gap

You are right that it is the more valuable entry of the two. TODO.md has it, above the constraint one, with the `ALTER VIEW ... SET` shape and a note that `SortedStorageParams` and `Table.StorageParamsSQL` are reusable as they are.

`make lint`, `make deadcode`, `make test`, `make test-scenario` and the sample round trip all pass. Coverage measured the way CI does on 15: 94.8% -> 94.9%.

winebarrel:
Re-reviewed c9e821b against a local PostgreSQL 16. `go vet`, `go test -p 1 ./...` and `make test-scenario` all pass, and all three are fixed.

**1.** `WITH (oids = false)` plans clean again, and `WITH (oids = false, fillfactor = 70)` still plans the `fillfactor` alone, so the skip is on the name rather than the clause. `toast.oids` needs no reading either: `CREATE TABLE` answers `unrecognized parameter "oids"` for it.

**2.** The group is whole again, and the `-- skipped:` line still lands after the `SET` when the drop is disallowed:

```
ALTER TABLE public.t4
  ADD COLUMN name text,
  DROP COLUMN old_col;
ALTER TABLE public.t4 SET (fillfactor='90');
```

Your TOAST reason holds, and so does the part about the drops in between:

```
after create:        reltoastrelid <> 0 = false
after add text col:  true
after drop text col: true
```

One correction to the framing, not the code: the old placement already sat after `colStmts`, so `ADD COLUMN` came before the `SET` there too. I ran the toast scenario against 2c58808 and it planned in the right order. So it is a constraint the new placement keeps rather than a second bug the move found, which makes `apply/table_storage_params_new_toast_column.yml` a guard against a future move rather than a regression test. Worth having either way, and the comment naming both reasons is the right call.

**3.** The TODO entry reads accurately. I confirmed the dump loses `security_barrier` and that a desired view writing the clause plans nothing, so `dump` feeding back clean is right as well.

`bulk_alter_table_storage_params.yml` covering a drop is the fixture that would have caught this, so the coverage gap is closed and not just the behavior.

LGTM.

---

winebarrel:
Agreed on the correction, and it is mine to own: the old placement was already after `colStmts`, so nothing about the TOAST ordering was broken on 2c58808. My earlier comment framed it as a second bug the move turned up, which overstates it. It is a constraint the end of `diffTable` happens to satisfy, and `apply/table_storage_params_new_toast_column.yml` guards a future move rather than fixing anything. The code comment and the commit message say it that way; only the comment above did not.

Thanks for checking `toast.oids` and the `-- skipped:` ordering. Neither had occurred to me.

Added since your review: `CHANGELOG.md` has a 1.36.0 section, two entries, one per fix. The table entry mentions the `oids` skip, since a pre-12 dump is exactly the input the entry is about.

PR #459, opened 2026-08-29
Check that a dump reloads into the schema it came from

winebarrel:
A third test suite, `make test-fidelity`, and the schemas to go with it.

## Why the two we have cannot catch this

`test/samples` feeds a `pista dump` back to `pista plan` and requires `No changes`. That catches the catalog reader and the parser disagreeing, and nothing else: when both sides overlook the same construct they agree perfectly. The Go fixtures compare against an expected output someone wrote, so they cover what someone thought to write down.

A table's storage parameters were invisible to both. `dump` did not write the clause, `plan` did not look for one, and no fixture mentioned `fillfactor`. It took reading psqldef's export output to notice, which is not a repeatable way to find bugs.

## What this does

For each schema: load it, record `pg_dump -s`, load `pista dump` into an empty database, record `pg_dump -s` again, require the two to be identical. PostgreSQL supplies the expected output, so nothing `dump` drops can hide, whether or not anyone knew to look for it.

```
=== restore fidelity ===
  column_storage           PASS
  columns                  PASS
  ...
  15 passed, 0 failed
```

Adding coverage is one `.sql` file under `schemas/` and nothing else. There is no expected output to write, which is what makes it general: it does not ask the author to name in advance what the schema is checking.

A schema holds only what pistachio manages. Adding a construct it does not manage produces a diff, which is a decision to make rather than noise to silence.

## Validation

Against 16d8e62, the commit before #458 read the storage parameters:

```
  storage_params           DIFF
    @@ -3,8 +3,7 @@
     CREATE TABLE public.events (
         id integer NOT NULL,
         body text
    -)
    -WITH (autovacuum_enabled=off, ... fillfactor='70', toast.autovacuum_enabled=off);
    +);
```

That is the gap the manual audit found, reproduced by the suite in one run.

A schema carrying `ALTER TABLE ... REPLICA IDENTITY FULL` fails on today's tree, which is the unmanaged case behaving as intended:

```
  replica_identity         DIFF
    -ALTER TABLE ONLY public.t REPLICA IDENTITY FULL;
```

## What it found on its own

A dump of a partitioned table with an index on the parent does not reload. `dump` writes each table followed by its own indexes, so the parent's index is created before the partitions, and PostgreSQL then creates the partition's copy under the name the dump goes on to declare:

```
psql: ERROR:  relation "logs_2025_at_idx" already exists
```

pg_dump avoids it by writing every table first and every index after, with an `ALTER INDEX ... ATTACH PARTITION` per child. Feeding the dump back to `pista plan` is clean, so this is within the contract and only reaches someone loading a dump with psql. TODO.md records it with the two possible shapes, and `partition.sql` leaves the parent index out with a note pointing there. Putting it back is the regression test once it is fixed.

Worth noting for the same reason: pg_dump does print a constraint's index storage parameters, `ADD CONSTRAINT t_v_key UNIQUE (v) WITH (fillfactor='70')`, even though `pg_get_constraintdef` does not. So the TODO entry from #458 is reachable by this suite too, and a schema for it is the regression test there.

## CI

Added to the existing version matrix job, so it runs on 15 through 18. pg_dump refuses a server newer than itself and the runner ships an older client than 18, so the job installs the newest one first; it dumps every version in the matrix.

`make lint`, `make test`, `make test-scenario` and `make test-fidelity` pass locally on 15.


winebarrel:
The approach is right: letting `pg_dump` supply the expected output is the one check the other two suites structurally cannot make, and the storage-parameter miss is the proof. The schemas cover a good spread, and leaving the parent index out of `partition.sql` with a TODO entry rather than working around it is the right call.

I ran the suite locally against a PostgreSQL 16 server: 15 passed, 0 failed. Adding `CREATE INDEX logs_at_idx ON public.logs USING btree (at);` back to `partition.sql` reproduces the documented failure exactly (`ERROR: relation "logs_2025_at_idx" already exists`), so the TODO entry is accurate and the schema comment points where it should.

One thing worth fixing before merge.

`test-fidelity` never builds `pista`. `run.sh` defaults `PISTA` to `./pista` but, unlike `test/scenario/run.sh`, it does not build it, and the Makefile target depends only on `clean-schema`. On a fresh checkout:

```
$ make test-fidelity
  column_storage           FAIL (dump)
    test/fidelity/run.sh: line 81: ./pista: No such file or directory
  ...
  0 passed, 15 failed
```

In CI it passes only because `make test-scenario` runs earlier in the same job and leaves the binary behind, so the suite silently depends on step order. Locally the stale-binary case is worse than the missing one: a `./pista` left over from an earlier build passes while testing old code, which is the failure mode this suite exists to prevent. `test-fidelity: build clean-schema` covers it, since `make build` already writes `./pista`; a `go build` at the top of `run.sh` matches what the scenario suite does.

Three minor notes, none blocking:

- If `pg_dump` is older than the server the version error prints once per schema, 15 near-identical failures. One preflight check comparing `pg_dump --version` against the server would say it once.
- `reset_db` drops only `public` and `schema_only` dumps only `-n public`, so "adding coverage is one `.sql` file" holds for single-schema cases. A cross-schema case, which is where `--search-path` and qualification decisions live, needs `run.sh` changes too. Fine as scoped, worth knowing before someone tries.
- This suite reaches the server through `PGHOST`/`PGUSER`/`PGPORT`, with `pista` picking up `PGPORT` through pgx's environment fallback on its default connection string. `TEST_PISTA_CONN_STR` therefore does not reach it at all, unlike `test/scenario`. The AGENTS.md line covers `PGPORT`; saying plainly that the connection string is not honored here would save someone a confused run.

LGTM apart from the build dependency.

---

winebarrel:
All four addressed in the follow-up commit.

**The build.** Confirmed by removing `./pista`: fifteen `FAIL (dump)` lines over a missing binary. `run.sh` now builds first, matching `test/scenario/run.sh` and `test/samples/run.sh`, which both do `go build -o pista ./cmd/pista` themselves rather than leaning on the make target. Your point about the stale binary is the one that decided it: passing while testing old code is exactly what this suite is supposed to make impossible, so it had to build rather than merely depend on `build`.

**The version error.** One preflight comparing `pg_dump --version` against `server_version_num`:

```
$ PATH=/tmp/fakebin:$PATH bash test/fidelity/run.sh
Building pista...
pg_dump 14 cannot dump a PostgreSQL 15 server
$ echo $?
1
```

**The two scoping notes.** Both are in the script header now, and AGENTS.md repeats them: the server is reached through `PGHOST` / `PGUSER` / `PGPORT` and `TEST_PISTA_CONN_STR` does not apply, and only `public` is reset and dumped, so a cross-schema case needs `run.sh` changes and not just a new file. You are right that this is where `--search-path` and qualification decisions live, so it is worth someone knowing before they try.

`make test-fidelity` passes on 15 with the binary removed first, and the ASCII check is clean.

winebarrel:
Re-reviewed d2f5531. The build step, the preflight and both doc notes are what I asked for, and I confirmed the fresh-checkout case locally against a PostgreSQL 16 server: `rm -f pista && make PGPORT=5416 test-fidelity` now builds first and reports 15 passed, 0 failed. The version parse lands on the right number for the packaging string in play here (`pg_dump (PostgreSQL) 16.13 (Ubuntu 16.13-...)` gives 16), and `server_version_num / 10000` is right.

One thing survives, from the same root as the original point. `run.sh` sets `set -uo pipefail` with no `-e`, so a failed `go build` does not stop the run. With a `pista` left over from an earlier build, the suite goes on to test the stale binary:

```
$ printf '\nfunc broken() { this is not go }\n' >> dump.go
$ bash test/fidelity/run.sh
Building pista...
# github.com/winebarrel/pistachio
./dump.go:586:22: syntax error: unexpected name is at end of statement
=== restore fidelity ===
  column_storage           PASS
  ...
  15 passed, 0 failed
$ echo $?
0
```

Green while the tree does not compile, which is the failure mode the commit message names. `test/scenario/run.sh` is safe from it because `set -euo pipefail` covers its build; here the rest of the script relies on `-e` being off, so guarding the one line is enough:

```sh
go build -o pista ./cmd/pista || exit 1
```

One optional nit while you are in there: when the server is unreachable the preflight reports it oddly. psql prints its connection error, then the arithmetic gets an empty operand and the script dies on `server_major: unbound variable`. It does exit 1, so nothing passes wrongly, and the real cause is on the line above; only the message is confusing.

With the build guard, LGTM.

---

winebarrel:
Both fixed in d4851fb.

**The build guard.** Reproduced with your recipe: green, exit 0, on a tree that does not compile. `go build -o pista ./cmd/pista || exit 1` now, with a comment saying why the guard is there rather than `set -e` on the whole script, since the checks below depend on `-e` being off. Verified the same broken tree exits 1.

Two commits in a row from the same root, both about a stale binary passing. The first left the binary unbuilt, the second left a failed build unnoticed; the check for the second is the one that would have caught the first.

**The preflight message.** It read the version inside an arithmetic expansion, so an unreachable server died on `server_major: unbound variable` after psql had already printed the real cause. It reads the value into a variable first and exits on the psql failure, so psql's message is the last thing printed:

```
$ PGPORT=9999 bash test/fidelity/run.sh
Building pista...
psql: error: connection to server at "localhost" (::1), port 9999 failed: Connection refused
	Is the server running on that host and accepting TCP/IP connections?
$ echo $?
1
```

`make test-fidelity` is 15 passed, 0 failed on 15.

PR #460, opened 2026-08-29
Split the documentation into a site, and cut the README down to an entry point

winebarrel:
Plan A from the discussion: the Markdown moves under `docs/`, MkDocs builds it, and README becomes an entry point.

## Why

README.md was doing four jobs in 1046 lines: the pitch, installation, a tutorial, and the complete CLI reference. getting-started.md then explained renaming, filtering, schema mapping, drops, transactions and arbitrary SQL a second time, shorter. There was no path from "what is this" to "how do I do X" that did not go through the whole file.

## What the site is

Markdown under `docs/`, built by MkDocs Material, published to GitHub Pages by `.github/workflows/docs.yml`. The Markdown stays the source and the HTML is never committed. `mkdocs build --strict` fails on a broken internal link, which is what caught the six this split created. A pull request builds without publishing; only a push to `main` deploys.

```
Home                          docs/index.md                     new
Getting Started               docs/getting-started.md           six steps, nothing else
Guides
  Renaming objects            docs/guides/renaming.md
  Working with multiple schemas
  Filtering what is managed
  Controlling drops
  Transactions and locks
  Running arbitrary SQL
  Splitting the schema across files
  Using pistachio in CI       docs/guides/ci.md                 new
Reference
  Commands                    docs/reference/commands.md
  Directives                  moved as is
  Supported objects
  Configuration
About
  Design and scope            docs/about/design.md              new
  Known limitations           symlink to TODO.md
  Performance                 moved as is
  Changelog                   symlink to CHANGELOG.md
Contributing
  Development
  Sample database tests       moved as is
```

Each guide is one task, and the duplication between README and getting-started is resolved into a single page each. The reference keeps the exhaustive material.

## The three new pages

**Home.** What pistachio is, the workflow diagram, the demo image, installation, one example.

**Using pistachio in CI.** This did not exist anywhere. `plan --check` has returned exit code 2 on drift for a long time and nothing showed what to do with it: failing a pull request on drift, posting the plan as a comment, applying on merge with `--with-tx`, and `dump --split` to keep the diff reviewable. The install snippet uses `gh release download` with a glob, since the release assets carry the version in the filename; I ran it to check it extracts a working binary.

**Design and scope.** What is deliberately not managed and why. Until now this lived only in AGENTS.md, where a user would not look. Publishing the known limitations alongside it is deliberate: 34 entries with reproduction and a fix sketch say more about the state of the project than a page claiming there are none.

## Notes

`CHANGELOG.md` and `TODO.md` stay at the repository root, where the release process and the habit of reading them expect them, and are symlinked into `docs/about/`. MkDocs follows symlinked files.

README.md is 102 lines: what it is, install, one example, links to the site.

## Before this can publish

Pages has to be enabled on the repository with **GitHub Actions** as the source. The workflow deploys but cannot turn the feature on. The site URL in `mkdocs.yml` and the README links assume `https://winebarrel.github.io/pistachio/`; a custom domain is a `CNAME` file and a DNS record on top, and can come later without touching the content.

`make lint`, `make test` and `mkdocs build --strict` pass.


winebarrel:
Reviewed the split. The structure is right, the move is faithful, and the site builds. Overall LGTM. Five cosmetic points, none blocking.

## What I checked

- **Help text is accurate.** Extracted the four fenced help blocks from `docs/reference/commands.md` and diffed them against `pista --help`, `pista plan --help`, `pista apply --help` and `pista dump --help` from a build of this branch at `COLUMNS=80`. All four match line for line.
- **The site builds.** `pip install -r requirements-docs.txt && mkdocs build --strict` is clean, symlinked `CHANGELOG.md` / `TODO.md` included.
- **Links resolve.** Every relative Markdown link under `docs/` and in `README.md` points at a file that exists; the README's deep links (`/getting-started/`, `/reference/commands/`, `/about/design/`, `/#try-it-with-docker`) match the pages and anchors the build produces.
- **Nothing was lost.** Of the non-heading lines in the old `README.md` plus `getting-started.md`, 19 do not appear verbatim in the new tree; each one is a link rewritten to its new target or the shorter half of a duplicated pair. The `-I`/`-E`/`--enable`/`--disable` examples and the `-- pista:execute` check semantics all survive in the longer version.
- **Numbers in `about/design.md` are right.** `SAMPLES` in `sample-db.mk` holds 47 entries and `test/fidelity/schemas/` holds 15 files, which is what "47 real-world schemas" and "a smaller set" describe.
- **No stale pointers.** `AGENTS.md` and `sample-db.mk` were updated for the moved files, and nothing else in the tree still references `getting-started.md`, `directives.md`, `performance.md` or `sample-db-test.md` at the old paths.

## Nits

1. **GitHub callouts render as literal text.** Material for MkDocs does not understand `> [!NOTE]`; in the built HTML those lines come out as a plain blockquote whose first line reads `[!NOTE]`. Eight occurrences: `guides/drops.md` (1), `guides/filtering.md` (2), `guides/renaming.md` (1, `[!TIP]`), `reference/commands.md` (4, one of them lowercase `[!note]`), `reference/objects.md` (2, `[!IMPORTANT]`). `admonition` and `pymdownx.details` are already enabled, so `!!! note` / `!!! tip` / `!!! warning` renders these properly. These pages are site-only now, so there is no GitHub rendering left to keep them for.

2. **The limitations page shows as "TODO" in the nav.** `about/limitations.md` is a symlink to `TODO.md`, so MkDocs takes the title from that file's H1. Give it an explicit title in `mkdocs.yml` (`- Known limitations: about/limitations.md`) to get the name the PR description promises. `about/changelog.md` happens to be fine.

3. **`docs/index.md` lost the `## Example` heading.** The `Create a schema file:` block and the plan/apply snippets now sit under `### Download binary` in both the nav and the page TOC. The README kept its `## Example` heading; the home page reads as if installation continues into the example.

4. **`docs/contributing.md` has two H1s** (`# Contributing` then `# Development`). The second should be `##`.

5. **Optional:** `.github/workflows/docs.yml` has no `concurrency` group. Two pushes to `main` in quick succession can have their `deploy-pages` jobs race; `concurrency: { group: pages, cancel-in-progress: false }` is the usual guard. Actions are pinned by SHA and the permission split between `build` and `deploy` is right otherwise.

Not approving, per your note.

---

winebarrel:
All four fixed in 17b19fc. The fifth, the limitations page showing as TODO, went in as 247c3e6 shortly before your review landed, along with the same problem one level up: `contributing.md` sat inside a section already called Contributing and is labelled Development now.

**1. Callouts.** Ten of them across five pages, one more than your count: `reference/commands.md` had four rather than three. All are `!!!` admonitions now. `[!IMPORTANT]` maps to `info` since Material has no `important`; the rest keep their names. Verified the built HTML carries `<div class="admonition note">` where the blockquote used to be.

Worth recording, since it nearly went wrong: the conversion walked `docs/**/*.md`, which includes the two symlinks. Writing through a symlink reaches the root file, so a callout in `CHANGELOG.md` or `TODO.md` would have been silently rewritten to a syntax GitHub does not render. Neither file has one, and `git status` after the run confirms neither changed, but the next person editing every page under `docs/` should know the symlinks are live.

**2. `## Example`.** Restored. The split dropped it because the extraction range started one line late.

**3. Two H1s in contributing.md.** Fixed.

**4. Concurrency.** Added as `group: pages` with `cancel-in-progress: false`. Cancelling the older deployment is the wrong default here: it publishes out of order.

Thanks for diffing the help blocks against a real build. That is the part of this change I could not check by reading.

winebarrel:
Re-reviewed at `17b19fc`. All five points are fixed, and I checked each one against a fresh `mkdocs build --strict` rather than the diff alone. LGTM.

## Verified

- **Callouts render.** No `[!NOTE]`-style text survives anywhere in the built site, and all eight now come out as real admonitions: 1 in `guides/drops`, 2 in `guides/filtering`, 1 in `guides/renaming`, 4 in `reference/commands`, 2 in `reference/objects`. Mapping `[!IMPORTANT]` to `!!! info` is a fair call — Material has no `important` type, and the two blocks it covers are explanatory rather than hazardous. Indenting the multi-paragraph ones by four spaces kept every paragraph inside the box.
- **Nav titles.** `Known limitations` and `Sample database tests` now read as intended, and `Home / Getting Started / Guides / Reference / About / Contributing` come out in the order the description lays out.
- **`## Example` is back** on the home page, with its own TOC entry between `Download binary` and `Where to go next`.
- **`docs/contributing.md`** has a single H1.
- **`concurrency: {group: pages, cancel-in-progress: false}`** is right for this: `cancel-in-progress: true` would have let a cancelled older run leave the newer artifact unpublished on some interleavings.
- **Nothing regressed.** `mkdocs build --strict` is clean, the four help blocks in `reference/commands.md` still match `pista --help` / `plan` / `apply` / `dump --help` line for line at `COLUMNS=80`, and every relative link still resolves.

## Two small things left, both non-blocking

1. **`docs/reference/objects.md` has four H1s** — `# Supported objects`, `# Routines`, `# Triggers`, `# Constraint and index naming`. All four were `##` siblings in the old README, so the merge into one page promoted each of them. It is the same shape `docs/contributing.md` just lost: the page TOC lists four top-level entries and the last three render as siblings of the page title. Demoting the last three to `##` matches every other page.

2. **`docs/contributing.md`'s tab title and its H1 disagree.** The nav override names the page `Development`, so the browser tab reads `Development - pistachio` while the page opens with `# Contributing`. Making the H1 `# Development` and dropping the `## Development` below it settles it. The same mismatch on the limitations page (`Known limitations` in the tab, `TODO` as the H1) is inherent to the symlink and not worth changing `TODO.md` for.

---

winebarrel:
Both fixed in 6af94cb.

`reference/objects.md` has one H1 now, with Routines, Triggers and Constraint and index naming as `##` under it.

`contributing.md` is headed Development, which matches the tab. That made the nav override redundant, so the entry names the file alone again and the title comes from the heading like every other page. Agreed on leaving the limitations page as it is: the tab and the heading disagreeing is the price of the symlink, and editing `TODO.md` to suit the site would be the wrong direction.

One correction to the record, not to the code: the callout count is ten, not eight. Your own enumeration adds up to ten (1 + 2 + 1 + 4 + 2), and ten is what the conversion reported.

`mkdocs build --strict` is clean.

winebarrel:
Re-reviewed at `6af94cb`. Both remaining points are fixed. LGTM.

- `reference/objects.md` is one H1 with three `##` under it now, and the demotion kept the anchors: `#routines`, `#triggers` and `#constraint-and-index-naming` are unchanged, so the in-page link in Triggers and the `../reference/objects.md#routines` link from the arbitrary-SQL guide still land. I checked every `#anchor` link in `docs/` against the ids the build emits: all resolve.
- `contributing.md` names itself `# Development`, and dropping the nav override was the cleaner half of that — the tab title and the H1 now come from one place.
- Still clean after the change: `mkdocs build --strict`, the four help blocks against `pista --help` / `plan` / `apply` / `dump --help`, all eight admonitions rendering (`note` x7, `tip` x1, `info` x2 — no literal `[!` left in any page), and the nav reading `Design and scope / Known limitations / Performance / Changelog` and `Development / Sample database tests`.

One optional leftover, purely cosmetic: `contributing-samples.md` still carries a nav override (`Sample database tests`) that differs from its H1 (`Sample Database Tests`) only in case. Dropping the override, as you just did for the other page, or matching the case would make the pair consistent. Not worth a commit on its own.

---

winebarrel:
Took the leftover as well, in 1d6923d. The heading is sentence case, which is what every other page in the tree uses, and the override is gone. `Known limitations` is now the only nav entry that overrides a title, and only because the symlink gives no choice.

Thanks for checking the anchors after the demotion. That was the part of it I would have missed.

winebarrel:
Re-reviewed at `1d6923d`. LGTM, nothing left from my side.

The last nav override is gone and the sample-tests page titles itself, so every page in the nav except the two symlinks now takes its name from its own H1. Nothing else referenced the old `Sample Database Tests` spelling.

Re-checked after the change: `mkdocs build --strict` clean, the four help blocks still match `pista --help` / `plan` / `apply` / `dump --help` line for line, every relative link and every `#anchor` in `docs/` resolves against the ids the build emits, all eight admonitions render with no literal `[!` in any page, and the nav reads as intended end to end.

The only remaining title-vs-H1 differences are the two deliberate ones: `Getting Started` shortened in the nav, and `Known limitations` over `TODO.md`'s own heading.

---

PR #470, opened 2026-08-29
Cut the Windows CI job's PostgreSQL connection overhead

winebarrel:
The Windows job takes twice as long as the Linux ones. Timing the last run on main (33242445350's predecessor, 33242189990):

| | Windows | Linux (pg18) |
| --- | --- | --- |
| job | 6:00 | 3:06 |
| `make` step | 4:38 | 1:11 |
| root package | 230s | 44s |
| TestApply / TestPlan / TestDump | 101 / 78 / 17s | 20 / 15 / 4s |

Plan, apply and dump each open their own connection, so the fixtures open more than a thousand of them, and the gap tracks the connection count exactly: plan runs one per case and costs 122ms more than on Linux, apply runs three and costs 358ms.

Two changes, both confined to the Windows job:

- A backend on Windows is a new process rather than a fork, and Defender scans every one. Exclude the server's files and executable. This only buys time, so it warns and carries on if the image stops allowing it.
- Drop the SSL negotiation round-trip from each connection through `sslmode=disable`. The server is on loopback with SSL off, so there is nothing to negotiate. Against a local container that round-trip alone is worth 12-19% of TestPlan (prefer 11.1/12.3/13.1s, disable 9.0/10.8/11.7s).

If this does not bring the job under about 4 minutes, the next step is splitting it into a matrix over TestApply / TestPlan / the rest, which trades runners for wall clock.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,11 @@ jobs:
       # The runner's PostgreSQL listens on the default port; the Makefile
       # otherwise defaults to the port compose.yaml publishes locally.
       PGPORT: "5432"
+      # Every plan, apply and dump opens its own connection, and the fixtures
+      # add up to more than a thousand of them. sslmode=disable takes the SSL
+      # negotiation round-trip out of each one; the server is on loopback and
+      # has SSL off anyway, so there is nothing to negotiate.
+      TEST_PISTA_CONN_STR: postgres://postgres@localhost:5432/postgres?sslmode=disable
     steps:
       # Keep LF line endings; test fixtures are compared byte for byte.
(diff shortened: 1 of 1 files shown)


winebarrel:
Measured on this run (33243014936) against main (33242189990):

| | main | this PR |
| --- | --- | --- |
| job | 6:00 | 3:57 |
| `make` step | 4:38 | 2:39 |
| root package | 230s | 118s |
| TestApply | 101s | 51s |
| TestPlan | 78s | 42s |
| TestDump | 17s | 8s |

The Windows job now lands within a minute of the Linux ones (3:03-3:31), so the matrix split is not needed.

PR #474, opened 2026-08-30
Stop BETWEEN, LIKE and NOT IN from drifting

winebarrel:
PostgreSQL rewrites several written forms before the expression reaches the
catalog, so pg_get_constraintdef, pg_get_expr, pg_get_indexdef, pg_get_viewdef
and pg_get_triggerdef never hand them back the way the file wrote them:

  BETWEEN, NOT BETWEEN, BETWEEN SYMMETRIC   -> the comparisons they stand for
  LIKE, ILIKE, NOT LIKE, NOT ILIKE          -> ~~, ~~*, !~~, !~~*
  SIMILAR TO, NOT SIMILAR TO                -> ~ / !~ similar_to_escape(...)
  NOT IN (...)                              -> <> ALL (ARRAY[...])
  (a, b)                                    -> ROW(a, b)

None of them matched, so a CHECK was dropped and re-added on every run, which
revalidates the whole table, a partial index was rebuilt and a view recreated.
A generated column failed the run instead, since its expression cannot be
altered in place, and no other table in the schema was planned either.
IN (...) already folded against the = ANY (ARRAY[...]) it is stored as; NOT IN
did not.

The folds move the written side onto the stored form, the direction PostgreSQL
itself defines, and they sit in the shared expression normalization, so a
domain CHECK, a column DEFAULT, a policy USING / WITH CHECK and a trigger WHEN
clause get them along with the table CHECK, the index and the view. The emitted
DDL still carries what the file wrote: the normalization is only compared,
never printed.

An expanded BETWEEN needs the AND/OR flattening the grammar performs while
parsing, since `(x BETWEEN 1 AND 10) AND y` is a single A_Expr at the point
makeAndExpr would have flattened it while the catalog definition re-parses
flat. AND and OR are associative, so the flat list means what the nesting did.

The scalar-array fold now matches the operator too. `= ANY` and `<> ALL` are
the two shapes transformAExprIn produces; every other operator reaches the
catalog as written, and folding one would have made `x > ANY (...)` and
`x > ALL (...)` compare equal.

The fixtures cover each expression site: a table CHECK, a partial index
predicate, an index key that is an expression, a view body and a view SELECT
target, a domain CHECK, a column DEFAULT, a policy predicate, a trigger WHEN
clause and a generated column. desugar_test.go carries the unit half, since
`make test` writes one profile per package and the fixtures compile into the
root package. Every catalog-side string in it is what pg_get_constraintdef
returned for the desired side on PostgreSQL 16, and each positive case fails
with the folds disabled. The negative cases are there too: LIKE against ILIKE
and NOT LIKE, SIMILAR TO against LIKE, IN against NOT IN, `> ANY` against
`> ALL`, a changed bound, a negated BETWEEN, and SYMMETRIC against the plain
form.

The grammar's other rewrites already planned clean and were re-checked against
a live server: OVERLAPS, COLLATE, IS JSON, ISNULL / NOTNULL, NULLIF,
SUBSTRING / TRIM / POSITION / OVERLAY / EXTRACT, CAST, LIKE ANY (ARRAY[...])
and nested and negated BETWEEN combinations. A typed literal is still
re-printed in its type's output form, so `CHECK (a > timestamp '2000-01-01')`
drifts; TODO.md records it. An EXCLUDE constraint normalizes nothing at all, so
the folds miss it like every other one, and its existing TODO entry now names
the case.

Claude-Session: https://claude.ai/code/session_01Ar29pa7DH2eRj8pfFcFMDN

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -628,8 +628,10 @@ Every normalization the walk performs is therefore missing there. A
 schema-qualified call drifts, `EXCLUDE USING btree (public.lower_v(v) WITH =)`
 being re-emitted as `DROP CONSTRAINT` + `ADD CONSTRAINT` on every plan, and so
 does a cast the catalog adds and the file does not: `((v || 'x') WITH =)` is
-stored as `((v || 'x'::text) WITH =)`. This is wider than the schema question
-and predates the strip.
+stored as `((v || 'x'::text) WITH =)`. The folds in `diff/desugar.go` are
+missing too, so the `BETWEEN` in
+`EXCLUDE USING gist (r WITH &&) WHERE (a BETWEEN 1 AND 10)` drifts here alone.
+This is wider than the schema question and predates the strip.
 
 Passing `Exclusions` and `WhereClause` through the same walk is the obvious
 shape, and the elements are `IndexElem` nodes, which `normalizeIndexStmt`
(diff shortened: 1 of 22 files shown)


winebarrel:
Reviewed the change. LGTM.

What I checked:

- **BETWEEN expansion is faithful to `transformAExprBetween`.** All four shapes (plain, NOT, SYMMETRIC, NOT SYMMETRIC) match what parse analysis produces, including the operator flips and the AND/OR swap in the negated forms. The operand is cloned per comparison, so a later mutation of one copy cannot leak into another.
- **The walk contract holds.** `pgast.Walk` is post-order and does not revisit a replaced node's children, which is exactly what `desugarBetween` relies on: the operand and bounds are already normalized (text-cast strip, `stripFuncSchema`) before they are cloned into the expansion, and the fresh comparison nodes themselves need no further normalization. `no_diff_between_over_function_call.yml` and the `char_length` unit case pin this down.
- **`flattenBoolExpr` is necessary and safe.** The grammar's `makeAndExpr`/`makeOrExpr` flatten only on the left, so the catalog's re-parse of `active AND ((level >= 1) AND (level <= 10))` keeps right-side nesting that the fold has to collapse symmetrically. AND/OR associativity makes the flat list equivalent, NOT is excluded, and the normalization is compare-only, so it never changes emitted DDL. The fixture covers both operand positions.
- **`foldArrayComparison` is a strict improvement.** Matching the operator (`= ANY`, `<> ALL` only) fixes the old fold, which rewrote any `ANY (ARRAY[...])` regardless of operator; `alter_check_any_to_all.yml` and the qualified-`OPERATOR(...)` unit case guard the boundaries.
- **Negative coverage is there.** Changed bounds, negation, SYMMETRIC vs plain, LIKE vs ILIKE/NOT LIKE, SIMILAR TO vs LIKE, IN vs NOT IN, and `> ANY` vs `> ALL` all still produce a diff.
- **Docs and TODO.** The typed-literal drift and the EXCLUDE gap are recorded in TODO.md, which reaches the docs through the `limitations.md` symlink.
- **Tests.** `go vet` and the `diff`/`pgast` unit tests pass locally; Codecov shows the full pg15–18 + windows matrix ran with 98.75% patch coverage (the one uncovered line is a defensive guard in `desugarBetween`).

One non-blocking note: `equalConstraintDef`'s doc comment still says "All five are symmetric" was updated to "All of them" — good catch keeping that in sync; nothing else in the comments has drifted as far as I can tell.

---

PR #477, opened 2026-08-31
Add --exclusive / --exclusive-wait to make apply runs mutually exclusive

winebarrel:
Two apply runs on one database had nothing keeping them apart. apply reads
the catalog, computes the diff and issues the DDL on one connection with no
re-check in between, so a second apply, or a second CI pipeline passing
plan --check at the same time, could compute its diff from a state the other
was still changing and issue DDL based on it.

--exclusive makes apply runs mutually exclusive. The exclusion is a
session-level advisory lock, keyed on a fixed class ID and a hash of the
database name, taken right after connecting and before the catalog read, so
the whole diff-then-execute sequence is covered and applies to different
databases on one cluster stay independent. A held lock is an immediate
error. --exclusive-wait waits for the holder instead, up to the given
duration (0 waits without limit), enforced with a Go-side context deadline
so no server-side setting leaks into the session the DDL then runs on. The
two flags conflict.

Session level is what makes the lock uniform: transaction boundaries do not
affect it, so it behaves the same with and without --with-tx and across
CREATE INDEX CONCURRENTLY, and it disappears with the connection, including
on a crash. It holds no table lock and does not block other clients, and an
apply without the flag, or any other tool, is not excluded.

Documented as a standalone guide page, Preventing concurrent applies, since
the mechanism is orthogonal to the transaction flags.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.38.0] - 2026-08-31
+
+* Add `--exclusive` / `--exclusive-wait` to `apply`. Opted-in apply runs on one database become mutually exclusive: `--exclusive` fails at once while another exclusive apply is running, `--exclusive-wait` waits for it up to the given duration (`0` waits without limit). The exclusion is a session-level advisory lock scoped to the database, taken before the catalog read and released when the connection closes; `--with-tx` and CONCURRENTLY index DDL do not affect it, and DDL from any other source is not blocked.
+
 ## [1.37.0] - 2026-08-31
 
 * Stop `BETWEEN`, `LIKE` and `NOT IN` from drifting. The grammar rewrites `BETWEEN` into the comparisons it stands for, `LIKE` / `ILIKE` / `SIMILAR TO` into their operators, `NOT IN` into `<> ALL (ARRAY[...])` and `(a, b)` into `ROW(a, b)` before the expression reaches the catalog, so a desired schema that wrote the keyword never matched what came back: a `CHECK` was dropped and re-added on every run, which revalidates the whole table, a partial index was rebuilt and a view recreated. A generated column failed the run instead, since a generated expression cannot be altered in place. The folds sit in the shared expression normalization, so a domain `CHECK`, a column `DEFAULT`, a policy `USING` / `WITH CHECK` and a trigger `WHEN` clause get them too. `IN` already folded against the `= ANY (ARRAY[...])` it is stored as; that fold now matches the operator as well, so `x > ANY (...)` and `x > ALL (...)` stay apart. A typed literal is still re-printed in its type's output form and drifts; TODO.md covers it.
--- a/apply.go
+++ b/apply.go
@@ -28,7 +28,8 @@ type ApplyOptions struct {
 	// ExclusiveWait enables the same mutual exclusion as Exclusive and waits
 	// for the other apply instead of failing. A pointer because 0 is a valid
 	// value (wait without limit) and must be distinguishable from "not set".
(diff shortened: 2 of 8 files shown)


winebarrel:
Reviewed the branch and verified it locally. LGTM.

What I checked:

- **Lock placement**: `acquireExclusive` runs right after `connect` and before `diffAll`, and `Apply` uses that single connection for the whole diff-then-execute sequence, so the session-level lock covers everything and is released by the deferred `conn.Close` on every path, including errors.
- **Lock key**: the two-key form with a fixed class ID plus `hashtext(current_database())` correctly scopes the exclusion to one database per cluster. `0x50697374` fits in int4, and Postgres infers `$1` as int4 from the `(int, int)` overload, so the wire types work out.
- **Timeout vs. cancel disambiguation**: `waitCtx.Err() != nil && ctx.Err() == nil` is right. When `*wait == 0`, `waitCtx` is `ctx` itself, so a caller cancel can never be misreported as "did not finish within" — and `TestApplyExclusiveWaitCanceled` pins that down.
- **Deliberate choice of a Go-side deadline over `lock_timeout`** keeps the session clean for the DDL that follows. Good call, and the comment explains it.
- **Flag design**: the `xor:"exclusive"` group works (`--exclusive --exclusive-wait=5m` is rejected by kong), the `*time.Duration` pointer correctly distinguishes `0` (wait without limit) from unset, and a bogus duration fails at parse time. Negative values are rejected with a clear error.
- **Tests**: all 9 new tests pass against PostgreSQL 16, and the full root-package `TestApply*` suite still passes. Coverage is thorough: conflict, timeout, unlimited wait, finite wait that succeeds, uncontended, `--with-tx` coexistence, cancel, negative, and lock release after apply.
- **End-to-end**: ran two competing sessions through the CLI; `--exclusive` fails fast with the hint pointing at `--exclusive-wait`, and `--exclusive-wait=10s` prints the waiting notice, blocks until the holder releases, then applies.
- **Docs**: the guide page is accurate about session-level semantics, the nav entry is in place, and the `commands.md` churn is just help-column realignment from the two new flags.

One non-blocking nit: the negative-duration check runs inside `acquireExclusive`, so `--exclusive-wait=-3s` only fails after the connection is established. Validating before connecting would fail a hair earlier, but it is cosmetic and the current behavior is covered by a test.

The design also fits the project policy well: strictly opt-in, no implicit DDL or server-side settings, and uniform behavior across `--with-tx` / `--try-tx` / CONCURRENTLY because the lock is session-level.

---

winebarrel:
Re-reviewed 7dd754c. The fix resolves the one nit cleanly — LGTM.

`UnsignedDuration` with `UnmarshalText` is the right shape: kong picks it up through the standard `encoding.TextUnmarshaler` path, so the rejection now happens at parse time on every input surface, before any connection is made. Verified against the built CLI:

- `--exclusive-wait=-3s` → `must not be negative: -3s` at parse time
- `PISTA_EXCLUSIVE_WAIT=-3s` → same error, annotated with the envar source
- `exclusive-wait: -3s` in a `-C` config file → same parse-time rejection
- `--exclusive-wait=0` still parses (wait without limit), `bogus` still fails with the duration parse error, and the xor conflict with `--exclusive` is unaffected

Replacing the in-`Apply` integration test with the pure `TestUnsignedDuration` unit test is the right trade: the negative case no longer reaches `acquireExclusive` at all, so testing the type directly covers more (valid, zero, negative, malformed) at zero DB cost. `acquireExclusive` itself is simpler for it.

All 8 remaining exclusive tests plus `TestUnsignedDuration` pass against PostgreSQL 16, the full root-package `TestApply*` suite passes, and build/vet/gofmt are clean. Nothing further from me.

---

PR #479, opened 2026-09-01
Make a table's storage parameters opt-in

winebarrel:
A table's storage parameters, the `WITH (...)` clause, have been managed since 1.36.0. That turned out to be the wrong default: the autovacuum settings a table is tuned with are set on the database more often than they are written in a schema file, so a run against such a table planned a `RESET` for every parameter the file did not name.

They are now opt-in behind `--manage-storage-param` (`$PISTA_MANAGE_STORAGE_PARAM`), available on `plan`, `apply` and `dump`.

Without the flag the parameters are cleared on both sides of the diff, in `diffAll` and in `Dump`: no `SET` or `RESET` is planned, the clause a file writes is left off the `CREATE TABLE`, and `dump` does not write one. A dump fed back plans clean either way. The catalog reads `reloptions` in the same query as everything else, so skipping the read would save nothing and it still happens.

An index's parameters are part of its definition and are managed either way.

## Tests

The 17 existing storage-parameter fixtures carry `manage_storage_param: true`, so the managed path is still asserted. Five fixtures cover the default:

- `plan`: an existing table whose parameters are added, changed and dropped at once plans nothing
- `plan`: a new table's `CREATE` comes out without the clause
- `dump`: the table clause is dropped and an index's is kept
- `dump`: a partition renders through its own branch and drops the clause there too
- `apply`: the parameters the database carries survive, a new table is created without the clause, and the re-plan is clean

`test/samples/run.sh` and `test/fidelity/run.sh` export `PISTA_MANAGE_STORAGE_PARAM=1`, the way they already do for routines, so both keep checking the round trip.

Coverage is unchanged against main (root 97.7%, diff 97.1%, 94.9% overall), and the new code is fully covered. `make lint`, `make test`, `make test-scenario` and `test/fidelity/run.sh` all pass on PostgreSQL 16.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.38.1] - 2026-09-01
+
+* Make a table's storage parameters opt-in, behind `--manage-storage-param` (`$PISTA_MANAGE_STORAGE_PARAM`). Managing them by default reset the autovacuum settings a table was tuned with, which are set on the database more often than written in a schema file. Without the flag they are dropped from both sides of the diff: no `SET` or `RESET` is planned, the `WITH` clause a file writes is left off the `CREATE TABLE`, and `dump` does not write one. An index's parameters are part of its definition and are managed either way.
+
 ## [1.38.0] - 2026-08-31
 
 * Add `--exclusive` / `--exclusive-wait` to `apply`. Opted-in apply runs on one database become mutually exclusive: `--exclusive` fails at once while another exclusive apply is running, `--exclusive-wait` waits for it up to the given duration (`0` waits without limit). The exclusion is a session-level advisory lock scoped to the database, taken before the catalog read and released when the connection closes; `--with-tx` and CONCURRENTLY index DDL do not affect it, and DDL from any other source is not blocked.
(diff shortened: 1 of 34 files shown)


winebarrel:
Reviewed the diff and ran the suite locally on PostgreSQL 16 (`go vet`, `make test`, `make test-scenario` — all green).

- The gate is in the right two places: `diffAll` clears both sides before any table diff runs, and `Dump` clears the filtered tables before rendering, so plan, apply (including its re-plan drift check) and dump all agree without touching the catalog read or the parser.
- Clearing `Table.StorageParams` covers every rendering branch (regular table, partitioned parent, partition), and the partition dump fixture confirms it. Index parameters live in the index definition, so they are untouched, as intended.
- The flag reaches all three commands automatically through the `FilterOptions` embed, and the docs cover each of them plus the new objects.md section.
- The round trip holds in every combination: a dump without the flag writes no clause and plans clean, and a managed dump fed to an unmanaged plan is also clean since the desired-side clause is dropped too.
- Fixture coverage of the default path is thorough (no-op on add/change/drop at once, CREATE without the clause, dump for table vs. index and for a partition, apply with re-plan), and the fidelity/samples runners exporting `PISTA_MANAGE_STORAGE_PARAM=1` keeps the managed round trip checked.

LGTM.

---

PR #481, opened 2026-09-01
Cover view triggers, empty enums and table extras in the diff suite

winebarrel:
The plan fixtures for these paths exist, but coverage is counted per
package, so DiffViews never ran viewTriggerStmts past the nil guard,
DiffEnums never added a value without an anchor, and newTableExtras
never rendered storage, RLS or triggers under its own tests.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01NL16AW9aeMc51p16kr1aSV

### DIFF
--- a/catalog/catalog_test.go
+++ b/catalog/catalog_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/winebarrel/pistachio/catalog"
 	"github.com/winebarrel/pistachio/internal/testutil"
+	"github.com/winebarrel/pistachio/model"
 )
 
 func TestNewCatalog(t *testing.T) {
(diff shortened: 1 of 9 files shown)


winebarrel (commented):
LGTM.

Reviewed the diff and verified locally on the PR head (`83d95e4`): `go vet ./...` is clean, and the parser, pgast, diff, and root dump tests all pass. The catalog closed-connection suite needs a live PostgreSQL, but its expectation (every reader returns an error on a closed connection) is sound.

Notes from the review:

- The `defElemString`/`defElemBool` fold is behavior-preserving: the generated protobuf getters are nil-safe, so `arg.GetString_().GetSval()` returns the zero value on a non-String argument exactly as the removed guards did.
- The Go tests here are justified over YAML fixtures: the coverage gaps are per-package (the fixtures live in the root suite), and the `DumpResult.String()` collision fallback is unreachable through `Client.Dump`, which rejects `--omit-schema` with several schemas.
- The empty-enum test correctly pins the anchor chain: first `ADD VALUE` bare, second anchored `AFTER` the first.
- One tiny nit, not blocking: `TestReadSQLFile_Stdin_ReadError` discards the read end of the pipe without closing it, leaking one fd for the life of the test process.

---

PR #483, opened 2026-09-02
Read a schema in a fixed number of queries

winebarrel:
A table's columns, constraints and policies were read one table at a
time, and a domain's constraints and a composite type's attributes one
type at a time, so a run cost a round trip per object. The gitlab sample
schema took 910 statements for its 1083 tables; it now takes 13.

Each reader now takes the objects it serves and filters on = ANY() over
their OIDs, returning the rows keyed by the OID they belong to. The
catalog is unchanged, and the sample schemas dump byte for byte as
before.

Every existing domain and composite type case declared a single object,
so nothing asserted that a row reaches the object it belongs to, and
handing each object the whole result set passed the suite. Both now
declare several objects, one of them without members.

The performance page notes that its numbers were measured before this,
since the catalog read is what they cover.

Claude-Session: https://claude.ai/code/session_01Y1tQFZ6dtHaafK4vyXaYvA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.39.0] - 2026-09-02
+
+* Read a schema in a fixed number of queries. A table's columns, constraints and policies were read one table at a time, and a domain's constraints and a composite type's attributes one type at a time, so a run cost a round trip per object. The gitlab sample schema took 910 statements for its 1083 tables; it now takes 13. The cost was in the round trips, so the gain grows with the latency to the database. The catalog is unchanged, so no plan or dump differs.
+
 ## [1.38.1] - 2026-09-01
 
 * Make a table's storage parameters opt-in, behind `--manage-storage-param` (`$PISTA_MANAGE_STORAGE_PARAM`). Managing them by default reset the autovacuum settings a table was tuned with, which are set on the database more often than written in a schema file. Without the flag they are dropped from both sides of the diff: no `SET` or `RESET` is planned, the `WITH` clause a file writes is left off the `CREATE TABLE`, and `dump` does not write one. An index's parameters are part of its definition and are managed either way.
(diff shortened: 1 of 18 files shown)


winebarrel:
LGTM.

Reviewed the diff and ran it locally against PostgreSQL 16: `go build`, `go vet`, `gofmt` and the full `go test -p 1 ./...` suite all pass.

What I checked:

- Every reader that took one object now filters on `= ANY(...::oid[])` and keys its result by the owning OID, and `ListTables`, `ListDomains` and `ListCompositeTypes` consume those maps. No per-object query loop is left in `catalog/`.
- The empty-input early return in each reader keeps a schema with no tables, domains or composite types from sending a query with an empty array.
- `tablesByOID` is only indexed with OIDs that came from the same `tables` slice the query filtered on, so the lookups in `ListConstraintsByTables` and `ListPoliciesByTables` cannot miss.
- The new tests cover the case the old suite could not catch (rows landing on the wrong object, an object with no members staying empty, same table name in two schemas), and the query-count test pins the fixed-statement-count property directly.

One optional nit, not blocking: the comment in `catalog/constraints.go` (around line 65) and two spots in `TODO.md` (lines 214 and 527) still refer to `ListColumnsByTable`, which this PR renames to `ListColumnsByTables`.

---

winebarrel:
Fixed the stale names in 4067970 (amended onto the single commit):

- `catalog/constraints.go:65` -> `ListColumnsByTables`
- `TODO.md:214` and `TODO.md:527` -> `catalog.ListColumnsByTables`

I checked the rest of the tree for the old names. The only remaining hits are three `CHANGELOG.md` entries for past releases (lines 302, 306, 310), which describe those releases as they shipped, so they keep the name that existed at the time.

`go vet`, `gofmt` and `go test -p 1 ./...` still pass against PostgreSQL 16.

---

PR #490, opened 2026-09-02
Widen the restore fidelity check, and fix the self-referencing FK sort

winebarrel:
## Summary

Fifteen new schema files for the restore fidelity check, covering constructs a real schema carries that the check did not reach. Twenty were tried; five failed. One was a bug and is fixed here, three are gaps now recorded in TODO.md, and the fourth TODO.md already covered.

The schemas double as fuzz seeds, since `fuzzseed.Schemas` reads that directory.

## The bug

A foreign key pointing at its own table's primary key, the `parent_id` of a tree, was added to the dependency graph as a self-edge and read back as a cycle. The composite type and view branches already skip a self-reference; the foreign key branch did not.

`orderStatements` runs on every plan and apply, and falls back to category order when the sort fails, so one such table anywhere in the schema dropped the order for the whole run:

```
BASE:  error: failed to execute SQL: CREATE OR REPLACE VIEW public.a_view ...
       ERROR: relation "public.b_view" does not exist (SQLSTATE 42P01)
FIXED: -- Apply finished in 4ms
```

`dump --sort-by-deps` treats a cycle as a hard error and wrote nothing at all.

`testdata/apply/self_fk_keeps_dependency_order.yml` fails without the fix; `testdata/dump/sort_by_deps_self_fk.yml` covers the dump path, and the toposort tests cover the qualified and unqualified reference.

## Gaps recorded rather than fixed

Each file notes what it leaves out, and TODO.md records the work:

- A `NOT VALID` check constraint is restored validated. The catalog reads `convalidated`, but a table's checks are written inside `CREATE TABLE`, where the clause cannot be spelled.
- An identity column's sequence options are not read. `Column.Identity` is a single byte.
- `WITH CHECK OPTION` on a view is not managed.

A sequence explicitly `OWNED BY` a column is dropped from the dump as well, which the existing "Sequence ownership transitions" entry already covers.

## Coverage

`toposort` goes from 97.9% to 99.2% of statements. The schema-qualified `nextval` literal, the form a pg_dump file carries, was reached by no test: both existing ones take the search_path fallback. The two blocks still uncovered are defensive guards on input neither the parser nor the catalog produces.

## Test plan

- `test-fidelity` 30/30 on PostgreSQL 16
- `make test` all packages, `make test-scenario` all scenarios, `make vet` clean
- both fuzz targets clean for 30s against the widened seed corpus
- `make lint` not run: the golangci-lint in this container is built with go1.25 and cannot load a go.mod targeting 1.27

https://claude.ai/code/session_01UCRyMzo6dnwWQf2dzoaBpW

---
_Generated by [Claude Code](https://claude.ai/code/session_01UCRyMzo6dnwWQf2dzoaBpW)_


winebarrel:
Reviewed. LGTM.

## The fix

`toposort/schema.go` is correct and minimal. A self-edge is never a real dependency in a "from depends on to" graph, and Kahn's algorithm can only read one as a cycle, so skipping it is the whole fix. Both the qualified and the unqualified branch are guarded, which matters because the two reach the node by different paths.

I checked out the head and confirmed the new unit test is a real regression test: with the two `ref != k` guards reverted, `TestOrderFromSchema_SelfReferencingFK` fails with `schema dependency sort failed: cycle detected: sorted 0 of 2 nodes`, and it passes with them. `go test ./toposort/` is green on the head as it stands.

The blast radius described in the PR body matches the code: `orderStatements` falls back to `fallbackOrder` on any sort error, so the whole run loses its order, and `dump --sort-by-deps` turns the same error into a hard failure. A `parent_id` tree is common enough that this was worth finding.

## One optional simplification

The self-edge guard now lives at three call sites (composite type attributes, view deps, and FKs), and two branches still lack it: the partition parent (`if defined[*t.PartitionOf]`) and the column type dep (`resolveTypeDep(col.TypeName, ...)`, which can return `k` since tables are in `defined`). `CREATE TABLE public.a (c public.a)` and `PARTITION OF public.a` are both rejected by PostgreSQL but accepted by the parser, so a desired schema or a fuzz input can still reach them and lose the order for the run the same way.

Dropping `from == to` once in `graph.AddEdge` would cover every branch and let all three site-local checks go. Given "do not chase corner cases", leaving it as is with the guard only where a valid schema reaches it is also a defensible call - just noting the asymmetry.

## Schemas and TODO entries

The fifteen new fidelity schemas are complementary to the existing ones rather than overlapping, and they follow the house style (one schema per file, `public` only, a leading comment naming what the file deliberately leaves out). `run.sh` globs the directory and `fuzzseed.Schemas` globs the same one, so both pick them up with no registration - the widened fuzz corpus comes for free.

Each of the three gaps has a TODO.md entry that names the code that would change, and the sequence-`OWNED BY` case correctly points at the existing entry at line 508 instead of adding a fourth. The `check_constraints.sql` note is the sharpest of the three: the constraint is readable but unspellable where it is written, which is a real design point and not just a missing field.

Nothing here is post-PG15 syntax (`NULLS NOT DISTINCT`, `INCLUDE` on a PK, `brin` with `pages_per_range`, transition-table triggers all land in 15 or earlier), so the fidelity run across the 15-18 matrix should agree with the 30/30 you got on 16.

## Minor

- CHANGELOG.md: the repo has not used an `## [Unreleased]` heading before. Every entry so far went in under its version heading in the same change (`## [1.39.0] - 2026-09-02` came in with 4067970). Worth matching unless the move to `Unreleased` is deliberate.
- `make lint` not having run locally is covered: `golangci-lint-action` runs inside each `test (15..18)` matrix job, so CI will report it. The Go diff is two boolean conditions, so there is little for it to find.

No blocking issues. Not approving, per your request.

---

PR #493, opened 2026-09-02
Manage an identity column's sequence options

winebarrel:
The `( ... )` after `AS IDENTITY` was read by neither side. `START WITH`, `INCREMENT BY`, `MINVALUE`, `MAXVALUE`, `CACHE` and `CYCLE` were dropped from the desired schema without a word, so the column was created at the defaults, and `pista dump` did not write them, so a restore lost them. TODO.md recorded it and `test/fidelity/schemas/identity.sql` left them out.

## What changed

`model.Column` carries an `IdentitySequence` with every value resolved. The catalog reads `pg_sequence` through the `pg_depend` row that ties the sequence to the column (deptype `'i'`, which only an identity column has), and the parser fills in the defaults PostgreSQL derives from the column type and the increment, so the diff is a plain comparison. The option-list parsing `CREATE SEQUENCE` already had is shared rather than written twice, and no query was added to the catalog read.

A column renders only the options that differ from those defaults, so the common identity column keeps its short form and no existing dump changes. A change to an existing column goes out as one `ALTER TABLE ... ALTER COLUMN ... SET`, alongside `SET GENERATED` when the kind changed too, and an option the desired schema no longer names is reset. `START WITH` moves where the sequence restarts and leaves the current value alone, the same as `ALTER SEQUENCE`.

A type change carries a bound that sat at the old type's default over to the new type's default and leaves an explicit bound alone, so the current side is read through that before the comparison. Without it an `integer` identity widened to `bigint` planned the `MAXVALUE` the retype had already set.

```sql
-- desired
CREATE TABLE public.items (
    id integer GENERATED ALWAYS AS IDENTITY (START WITH 200 INCREMENT BY 5 MAXVALUE 900 CACHE 20 CYCLE)
);

-- plan, against a column at START WITH 100 INCREMENT BY 5
ALTER TABLE public.items ALTER COLUMN id SET START WITH 200 SET MAXVALUE 900 SET CACHE 20 SET CYCLE;
```

## Tests

Eighteen fixtures across parser, dump, plan and apply: create, change, reset to the defaults, kind and options together, add identity to an existing column, add a new identity column, retype, drop identity from a column with options, an explicit `NO MINVALUE` / `NO MAXVALUE` / `NO CYCLE`, and `--bulk-alter`. Package-local tests in model, diff, catalog and parser, since fixtures run in the root package and do not count toward those.

`test/fidelity/schemas/identity.sql` now carries the options, so `pg_dump` before and after the restore compares them.

Coverage does not drop; parser and catalog gain slightly. Every statement of the new code is covered.

## Not covered

The name of the sequence behind an identity column is still unmanaged, so one renamed by hand restores under the default name. TODO.md has a new entry for it.


winebarrel:
Reviewed at f17fc6e. I built the branch and ran `go test -p 1 ./...`, `test/scenario/run.sh` and `test/fidelity/run.sh` against a local PostgreSQL 16; all green. The shape reads well: one struct with every value resolved on both sides, the option list parsing shared with `CREATE SEQUENCE`, no extra catalog query, and the short form kept for the common identity column.

One bug, in `RetypedTo`.

`model/identity.go` carries over a bound that sits at `DefaultIdentitySequence(oldType, increment)`, whose min is 1 for an ascending sequence and whose max is -1 for a descending one. PostgreSQL's rule is a different one: on a type change it resets a bound when the bound equals the **old type's** extreme, in either direction. A bound that is a type extreme against the direction is therefore moved by the server and not by the diff.

Repro on PG16:

```sql
-- current
CREATE TABLE public.items (id integer GENERATED ALWAYS AS IDENTITY (MINVALUE -2147483648));
-- desired
CREATE TABLE public.items (id bigint GENERATED ALWAYS AS IDENTITY (MINVALUE -2147483648));
```

```
$ pista plan desired.sql
ALTER TABLE public.items ALTER COLUMN id SET DATA TYPE bigint;
```

`SET DATA TYPE bigint` moves `seqmin` to -9223372036854775808, because -2147483648 was integer's minimum, but `RetypedTo` compares it against the ascending default of 1, leaves it, and the comparison sees no change. The apply leaves drift and the next `plan` prints `ALTER TABLE public.items ALTER COLUMN id SET MINVALUE -2147483648;`. A descending column with an explicit `MAXVALUE 2147483647` retyped to bigint goes the same way.

Reading both sides through `seqTypeBounds` covers those and covers everything the current version covers, since an ascending min of 1 and a descending max of -1 are not type extremes and the server leaves them alone too (checked against a live sequence):

```go
	oldMin, oldMax := seqTypeBounds(oldType)
	newMin, newMax := seqTypeBounds(newType)

	retyped := *s
	if retyped.Min == oldMin {
		retyped.Min = newMin
	}
	if retyped.Max == oldMax {
		retyped.Max = newMax
	}
```

With that the repro plans `SET DATA TYPE bigint` plus `SET MINVALUE -2147483648` and converges in one apply, and the full suite, the scenario tests and the fidelity check still pass. An `apply` fixture catches it, since the harness re-plans.

Two smaller notes, neither blocking:

* A change that puts the sequence's current value outside the new range is refused by the server rather than by the plan: `SET MAXVALUE 5` on a column whose sequence sits at 10 gives `ERROR: RESTART value (10) cannot be greater than MAXVALUE (5)`, and an ascending to descending flip gives the same. That is what `ALTER SEQUENCE` does, so not guessing a `RESTART` looks right to me; a line in `docs/reference/objects.md` would save the reader the surprise.
* `identitySeqClauses` returns early on `*current == *desired`, which the per-option comparisons already handle.

The rest read clean. `deparseTypeName` normalizes `int`, `int2` and `int8`, so the bounds resolve against the same spelling the catalog reports; a partitioned parent's identity round-trips, since the partition carries no sequence of its own and the nil guard keeps that from planning anything; and the `deptype 'i'` join skips a serial column's sequence.

---

winebarrel:
Fixed in d05afb7. RetypedTo now reads both sides through seqTypeBounds, exactly as suggested; verified against PostgreSQL 15 and 18 that the server moves a bound at the old type's extreme in either direction and leaves the rest, including an ascending min of 1 and a descending max of -1. Your repro now plans `SET DATA TYPE bigint` plus `SET MINVALUE -2147483648` and converges in one apply; it is pinned as a plan fixture and an apply fixture (the harness re-plans, so leftover drift fails), plus a unit case for the against-direction extreme.

Both non-blocking notes are addressed too: the whole-struct early return in identitySeqClauses is gone, and docs/reference/objects.md now says no RESTART is planned, so a change that puts the sequence's current value outside the new range fails at apply with the server's error.

winebarrel:
Re-reviewed at d05afb7. The fix reads right: `RetypedTo` now compares both bounds against `seqTypeBounds`, which is what the server does in either direction, and the comment says so.

Checked it end to end on PostgreSQL 16, plan then apply then plan again:

* the original repro, ascending `integer` with `MINVALUE -2147483648` widened to `bigint`, now plans `SET DATA TYPE bigint` plus `SET MINVALUE -2147483648` and converges in one apply
* the descending twin, `MAXVALUE 2147483647` widened to `bigint`, likewise
* `smallint` with `MINVALUE -32768` to `integer`
* narrowing: `bigint` to `integer`, and a descending `bigint` to `smallint`
* a bound at the new type's extreme but not the old one, `bigint (MAXVALUE 2147483647)` to `integer`, which correctly plans the type change alone
* type, kind and options changing together

All of them re-plan clean, and `dump` of a column whose bound sits at a type extreme round-trips with no changes. `go test -p 1 ./...`, `test/scenario/run.sh` and `test/fidelity/run.sh` pass.

The two fixtures cover the case, and the `apply` one re-plans, so a regression would show. The docs line on `RESTART` and the dropped early return in `identitySeqClauses` are both fine.

LGTM.

---

PR #494, opened 2026-09-02
Name the file, line and column in parse errors

winebarrel:
A typo reported only `failed to parse SQL: syntax error at or near "TABEL"`, with nothing to say which file or line, and the desired schema often spans several files. pg_query's error carries the cursor position, a 1-based character index into the parsed SQL, so the location was always there; it was just dropped.

## What changed

`ParseSQLFilesWithSchema` records where each file starts in the SQL it joins, and a failed parse renders the position in the style of a compiler:

```
pista: error: failed to parse SQL: syntax error at or near "TABEL"
 --> schema/items.sql:6:8
  |
6 | CREATE TABEL public.items (
  |        ^
```

The line number is local to the file the error is in. Directive validation returns the match offset its regexes already had, so an unknown or malformed `-- pista:` directive gets the same block. The caret pad keeps tabs so it lands under the column however the line is indented, and the column counts runes, matching how PostgreSQL counts the cursor. An error without a position is returned as it was, and no message text changed.

Duplicate-object errors still carry no position; giving them one means threading locations through every parse function.

## Tests

Twelve cases: a typo mid-file, an error in the second and in the middle of three files, an error at a file's first character, tab indentation, a multibyte line, an error at end of input, an unknown directive, a directive given arguments, an error without a position, a cursor past the input, and a two-digit line number. The parser fuzz target now runs the renderer on every failed parse, so arbitrary input cannot panic it.

Parser coverage goes from 93.7% to 93.9%; the new file is fully covered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* Name the file, line and column in parse errors. A syntax error or a bad `-- pista:` directive now prints the offending line with a caret under the column, in the style of a compiler, instead of the bare message. With several schema files the line number is local to the file the error is in.
+
 * Stop a self-referencing foreign key from disabling the dependency sort. A key pointing at its own table's primary key, the `parent_id` of a tree, was added to the graph as a self-edge and read back as a cycle. `plan` and `apply` fall back to category order when the sort fails, so one such table anywhere in the schema dropped the order for the whole run: two new views, one selecting from the other, went out in file order and the apply failed on the view that did not exist yet. `dump --sort-by-deps` treats a cycle as a hard error and wrote nothing at all. The composite type and view branches already skipped a self-reference; the foreign key branch now does the same.
 
 * Manage an identity column's sequence options, the `( ... )` after `AS IDENTITY`. `START WITH`, `INCREMENT BY`, `MINVALUE`, `MAXVALUE`, `CACHE` and `CYCLE` were read by neither side, so a file that wrote them was applied at the defaults and `dump` dropped them. A change goes out as `ALTER TABLE ... ALTER COLUMN ... SET`, and an option the desired schema no longer names is reset to its default. `dump` writes only the options that differ from the defaults.
--- a/diff_all.go
+++ b/diff_all.go
@@ -90,9 +90,10 @@ func (client *Client) diffAll(ctx context.Context, conn *pgx.Conn, options *diff
 		}
 	}
 
+	// Not wrapped: the parser's message already names the file.
 	desired, err := parser.ParseSQLFilesWithSchema(options.Files, client.Schemas[0])
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse SQL file: %w", err)
+		return nil, err
 	}
 
 	filterDesiredBySchemas(desired, client.Schemas, client.SchemaMap)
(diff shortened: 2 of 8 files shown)


winebarrel:
LGTM.

I checked out the branch and went over the offset math. The span list is built in ascending order and `offset += len(sql) + 1` matches the single `"\n"` that `strings.Join` inserts, so a file boundary always falls on a newline and a rendered line never spans two files. The span selection picks the last span whose start is at or before the offset, `lineNo` counts from `span.start` so it stays local to the file, and the `pg_query.Parse` call gets exactly the string the spans describe, so the cursor position lines up. `validateDirectives` also receives the joined SQL, so the byte offset a `locatedError` carries is in the same coordinate space. `errors.AsType` needs Go 1.26+, which `go.mod` already targets.

Behavior checks I ran locally: `go build ./...`, `go vet ./parser/...` and `go test ./parser/...` all pass, `annotateError` and `runeOffsetToByte` are at 100%, and the message text is unchanged so the `Contains`-based `bulk_alter_directive_args_error.yml` fixture still matches. Running the renderer on every failed parse in the fuzz target is a good way to keep the position arithmetic honest.

A few non-blocking notes.

The example in the description is not quite what a user sees. `diff_all.go:95` wraps the parse error again, so the first line comes out as `pista: error: failed to parse SQL file: failed to parse SQL: syntax error at or near "TABEL"`. The doubling predates this PR, but the location block makes the outer wrap redundant, so dropping it would be a cheap follow-up.

Tabs get special handling in the caret pad, but full-width characters do not: with `DEFAULT 'あいう' x` the pad writes one space per rune and the caret lands three columns left of the token. Column numbering is right either way since PostgreSQL counts characters too, and this is a corner case, so leaving it is fine.

When the schema comes from stdin the path is the literal `-`, so the location reads ` --> -:1:8`. `<stdin>` would read better.

`annotateError` treats any `*pgparser.Error` in the chain as positioned in the joined SQL, and `extractEnumValueDirectives` calls `pg_query.Scan` on a statement substring and wraps its error. That substring has already parsed once, so the error is not reachable in practice; it is only a trap for a future helper that parses a fragment.

---

winebarrel:
Both actionable notes are addressed in a6ae9fc. The wrap in diff_all.go is gone, so the first line now reads `failed to parse SQL: syntax error at or near "TABEL"` once; no test depended on the outer prefix. stdin renders as `<stdin>` in the location line, pinned by a test that pipes bad SQL through a swapped os.Stdin.

The full-width caret misalignment stays as you suggested, and thanks for the note on fragment parses; the fuzz hook should catch a future helper that leaks a fragment-positioned error into the joined-SQL path.

winebarrel:
Re-reviewed a6ae9fc. LGTM.

Dropping the wrap in `diff_all.go` is the right call: `plan` and `apply` both return `diffAll`'s error unchanged and kong prints it as-is, so the output now matches the example in the description exactly, and a missing file reads as a plain `failed to read SQL file: ...` instead of being buried behind a parse prefix. I grepped the whole tree for the removed string and nothing depends on it, so no fixture or scenario script had to move.

Mapping `-` to `<stdin>` in the span rather than in `annotateError` keeps the renderer free of path policy, and it stays consistent with `readSQLFile`, which already treats `-` as stdin. The new test mirrors the `os.Stdin` swap in `TestReadSQLFile_Stdin`, and nothing in the package calls `t.Parallel`, so the global swap is safe.

`go build ./...`, `go vet ./...` and `go test ./parser/...` pass locally, and `location.go` is still fully covered. On CI, build, both fuzz targets, samples, ascii, keywords and CodeQL are green on this head; the PostgreSQL matrix and the Windows job were still running when I looked, though nothing in this commit touches a path they exercise.

Leaving the full-width caret alignment and the substring-`Scan` note as they are is fine by me. Both are corner cases, and the schemas people actually write do not hit them.

---

PR #497, opened 2026-09-02
Name the object and the position in a duplicate-name error

winebarrel:
`duplicate column: id` said neither which table the column was on nor where the file repeats it. The parse-error location added in 1.40.0 covered syntax errors and directives only; duplicates were left out because giving them one meant threading locations through the parse functions.

`setUnique` now takes the object the name belongs to and the offset to point at, and returns a `locatedError`, so a duplicate gets the block a syntax error already gets:

```
duplicate column: id on public.users
 --> schema/users.sql:4:5
  |
4 |     id bigint
  |     ^
```

A column or constraint carries the location pg_query records on its own node, so the caret lands on the definition that repeats the name; every other kind points at the statement. pg_query counts the blank lines and comments before a statement as part of it, so `stmtStart` skips them by the rule the directive scan already uses.

The checks in `validateNamespaces` are unchanged. They report a clash between two objects rather than within one, and already name the scope.

## Tests

Location tests cover a duplicate column, a column-level constraint, a table-level constraint, a statement after a comment, a statement with no trailing semicolon, a second file, and a policy. The existing duplicate tests move from `Contains` to `Equal` so every kind's scope is fixed.

`go test ./parser/...` and `make lint` pass; parser coverage is 94.1% against 94.0% on main, with `setUnique` and `stmtStart` at 100%. The suites that need PostgreSQL were not run here.

---

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Name the object and the position in a duplicate-name error. `duplicate column: id` said neither which table the column was on nor where the file repeats it. It now reads `duplicate column: id on public.users` and prints the line with a caret under it, as a syntax error does. A column or constraint is pointed at where it is defined, everything else at the statement that repeats the name. The checks that report a clash between two objects, a table and a view of one name for example, are unchanged.
+
 ## [1.40.0] - 2026-09-04
 
 * Name the file, line and column in parse errors. A syntax error or a bad `-- pista:` directive now prints the offending line with a caret under the column, in the style of a compiler, instead of the bare message. With several schema files the line number is local to the file the error is in.
--- a/parser/location.go
+++ b/parser/location.go
@@ -7,8 +7,22 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	pg_query "github.com/pganalyze/pg_query_go/v6"
 	pgparser "github.com/pganalyze/pg_query_go/v6/parser"
 )
+
+// stmtStart returns the byte offset of a statement's first token. pg_query
+// counts the blank lines and comments before a statement as part of it, so
+// StmtLocation on its own can point at whitespace rather than at the keyword.
+func stmtStart(sql string, rawStmt *pg_query.RawStmt) int32 {
+	start := rawStmt.StmtLocation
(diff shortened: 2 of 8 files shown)


winebarrel:
LGTM.

I checked the branch out, ran `go test ./parser/...` and `go vet ./...` (both clean), and CI is green on e72baad's sibling commit a13f2a2 across the 15-18 matrix, windows, samples and both fuzz targets.

* The caret lands on the second occurrence, not the first, for every kind I tried, and `duplicate column: id on public.a` points at the repeat even when an earlier column has a quoted name (no rune/byte confusion).
* No `pg_query.ColumnDef` or `pg_query.Constraint` is synthesized anywhere in the parser, so `cd.Location` / `con.Location` always comes from the source text. Even if one were -1, `annotateError` returns the bare error rather than misrendering, so the degradation is graceful.
* Tightening the existing duplicate tests from `Contains` to `Equal` is worth it on its own: it pins the scope suffix for every kind.

Non-blocking notes:

1. `stmtStart` is added to `parser/location.go` by #499 as well, in exactly the same form, and all three open PRs add a `## [Unreleased]` heading at the same place in `CHANGELOG.md`. Whichever of the two lands second needs a conflict resolution that keeps one copy of `stmtStart`.

2. The description says `stmtStart` skips leading comments "by the rule the directive scan already uses", but it is stricter than that scan: it handles `StmtLen == 0`, which `extractStmtDirectives` / `extractFlagDirectives` / `extractExecuteDirectives` do not. On main, pg_query reports `StmtLen == 0` for a trailing statement with no semicolon, so those three compute an empty region and the directive is silently dropped. I confirmed it: `-- pista:renamed-from old_b` on the last statement gives `RenameFrom=nil` without a trailing semicolon and `public.old_b` with one. Pre-existing and outside this PR, but a shared `stmtRegion(sql, rawStmt)` helper would fix it and remove the clamp that is currently copied four times.

3. Two cosmetic corner cases, both inherited from `findLeadingCommentEnd` and both in the "do not chase corner cases" category, just noting them: with two statements on one line the caret points at the space before `CREATE` (column 32 rather than 33 in my check), and a `/* ... */` block comment before a statement is not skipped, so the caret lands on the `/*` line instead of the `CREATE` line.

---

PR #498, opened 2026-09-02
Read the desired schema before connecting

winebarrel:
`plan` and `apply` opened the connection first and only then read their files, so a missing file, a syntax error or a duplicate name was reported once the database had answered. A run against an unreachable database reported the connection rather than the file that was wrong whatever the database said, and an exclusive apply took its advisory lock before finding out it could not read its own input.

Nothing in the parse depends on the server: it needs the file paths and the first target schema, both of which are options. `loadDesiredInput` gathers what a run reads from the file system, the parsed schema and the resolved pre-SQL, and `Plan` and `Apply` call it before `connect` and hand the result to `diffAll`.

A successful run is unchanged, and so is every message. What changes is which error a broken run reports first, and that no connection is opened for one that cannot read its own files.

## Tests

`desired_input_test.go` points at a port nothing listens on and checks that `plan` and `apply` report a syntax error, a missing schema file and a missing pre-SQL file without a database, and that a schema that parses still reports the connection. All five fail without the change.

`go test` on the packages that need no database, `make lint` and `deadcode` pass. `loadDesiredInput` is at 100% from those tests alone. The suites that need PostgreSQL were not run here, so the claim that `--exclusive` no longer takes its lock for an unreadable file is not covered by a test.

---

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Read the desired schema before connecting. `plan` and `apply` opened the connection first, so a missing file, a syntax error or a duplicate name was reported only once the database answered, and a run against an unreachable database reported the connection rather than the file. Both now read and parse the schema files, and resolve `--pre-sql-file` and `--concurrently-pre-sql-file`, before connecting. A successful run is unchanged; one that cannot read its own files opens no connection, and takes no lock under `--exclusive`.
+
 ## [1.40.0] - 2026-09-04
 
 * Name the file, line and column in parse errors. A syntax error or a bad `-- pista:` directive now prints the offending line with a caret under the column, in the style of a compiler, instead of the bare message. With several schema files the line number is local to the file the error is in.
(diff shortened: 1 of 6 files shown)


winebarrel:
LGTM.

I checked the branch out, built it, ran `go vet ./...` and the new `desired_input_test.go` cases (all pass with no database, which is the point), and CI is green.

* `diffAll` has exactly two callers, `Plan` and `Apply`, and both set `Desired`, so the unconditional `options.Desired.schema` dereference is safe today. Worth remembering if a third caller appears, since `Desired` is now a required pointer field on an options struct where everything else is a zero-value-safe scalar.
* Both call sites run `client.validateSchemas()` before `loadDesiredInput`, so `client.Schemas[0]` cannot index an empty slice.
* Nothing in the parse depends on the connection, and no `options.Files` reference is left in `diff_all.go`, so the move is complete.
* The `exclusive-apply.md` wording change is accurate: the lock is still taken right after connecting, and the files are now read before that.

One non-blocking note: `loadDesiredInput(files, preSQL, preSQLFile, concurrentlyPreSQL, concurrentlyPreSQLFile)` takes four adjacent `string` parameters, so transposing two of them compiles and passes most tests. `PlanOptions` and `ApplyOptions` declare those fields separately (the help text differs), so there is no shared struct to hand over as it stands, and both call sites already put each argument on its own line, which keeps it readable. Fine as is; a small `preSQLOptions` value would remove the footgun if the signature grows again.

---

PR #499, opened 2026-09-02
Name the file, line and column in the ignored-statement warning

winebarrel:
A statement pistachio drops warned with its own text and nothing else, so finding it meant searching the desired schema files for that text, and one `GRANT` of several could not be told from another. A syntax error has pointed at its statement since 1.40.0; the warning now does the same:

```
pistachio: schema/items.sql:12:1: ignored unsupported statement: DROP TABLE public.items
```

The span lookup and the line and column arithmetic `annotateError` already did move into `locate`, which the warning shares. The position is the statement's first token: pg_query counts the blank lines and comments before a statement as part of it, so `stmtStart` skips them by the rule the directive scan already uses.

`parseSQLWithSchema` takes the file spans, so a caller parsing a string rather than files passes none and gets the message it got before. The warning now scans the file up to the statement; on a 4.7 MB schema with 2000 dropped statements, far past what a warning is meant for, that adds under a second to the parse.

## Tests

Four cases: a warning in one file, a warning in the second of two files, an ALTER TABLE action no handler reads, and a parse of a string, which names no position.

`go test ./parser/...` and `make lint` pass; parser coverage is 94.0%, the same as main, with `locate`, `sourcePos.String` and `stmtStart` at 100%. The suites that need PostgreSQL were not run here.

---

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Name the file, line and column in the ignored-statement warning. It carried the statement text alone, which does not say where the statement is when the schema spans several files, or when the same `GRANT` appears in more than one of them. It now reads `pistachio: schema/items.sql:12:1: ignored unsupported statement: DROP TABLE public.items`. A parse that came from no file names no position.
+
 ## [1.40.0] - 2026-09-04
 
 * Name the file, line and column in parse errors. A syntax error or a bad `-- pista:` directive now prints the offending line with a caret under the column, in the style of a compiler, instead of the bare message. With several schema files the line number is local to the file the error is in.
--- a/docs/reference/objects.md
+++ b/docs/reference/objects.md
@@ -19,7 +19,7 @@
 - Array, JSON, UUID, and other built-in types
 - Quoted identifiers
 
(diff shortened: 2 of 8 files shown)


winebarrel:
LGTM.

I checked the branch out, ran `go test ./parser/...` and `go vet ./...` (both clean), and CI is green, fuzz targets included.

* The `locate` / `sourcePos` extraction is behavior-preserving for `annotateError`: the same offset clamp, the same span selection, the same line and column arithmetic, and the caret pad still walks `sql[lineStart:offset]`, so the existing location tests pin the rendering unchanged.
* `warnIgnoredAlterTableCmds` copies `StmtLocation` and `StmtLen` onto the rebuilt statement, so the synthetic `ALTER TABLE` resolves to the real position rather than to offset 0. The test covers that.
* Deferring `locate` until after the empty-snippet check keeps the cost off statements that warn nothing.

Non-blocking notes:

1. `stmtStart` is added to `parser/location.go` by #497 as well, in exactly the same form, and all three open PRs add a `## [Unreleased]` heading at the same place in `CHANGELOG.md`. Whichever lands second needs a conflict resolution keeping one copy of `stmtStart`.

2. `locate` counts newlines from the start of the file for every warning, so N warnings in one file cost O(N x file size). Your 4.7 MB / 2000 warnings measurement says that is fine at any realistic scale, and a warning path is the right place to spend it, so this is a note rather than a request.

3. The changelog says "a library caller sees the old message", but `ParseSQLFilesWithSchema` and `ParseRoutineDef` are the only exported entry points; there is no exported string parse, so the nil-spans path is reachable only from the tests and the fuzz target today. "A parse that came from no file names no position" would describe it without implying a caller that does not exist.

4. Same two `findLeadingCommentEnd` corner cases I noted on #497 apply here (a `/* ... */` block comment before the statement is not skipped, and two statements on one line point at the separating space). Cosmetic and shared, not this PR's to fix.

---

winebarrel:
Note 3 is right: `ParseSQLFilesWithSchema` and `ParseRoutineDef` are the only exported entry points, so the nil-spans path has no caller outside the tests and the fuzz target. The changelog now says "A parse that came from no file names no position." (067d137).

Notes 1, 2 and 4 I am leaving as they are. The `stmtStart` copy and the changelog heading are conflicts to resolve at merge; the newline scan and the two `findLeadingCommentEnd` corner cases are shared with the existing code and not this change's to fix.

---

PR #500, opened 2026-09-02
Apply a directive on a file's last statement without a semicolon

winebarrel:
pg_query reports `StmtLen` 0 for a final statement that ends without a semicolon. Three of the scans that carve out a statement's text clamped only the upper bound, so such a statement got an empty region and every directive in it was dropped in silence: a rename planned a drop and a create, `-- pista:ignore` left the object managed, and `-- pista:execute` left the statement to the ignored-statement warning instead of running it at apply. The inline column, enum value and composite attribute renames went the same way.

On main:

```
renamed-from   no semicolon: RenameFrom=<nil>      with semicolon: RenameFrom=public.old_b
ignore         no semicolon: Ignore=false          with semicolon: Ignore=true
concurrently   no semicolon: Concurrently=false    with semicolon: Concurrently=true
execute        no semicolon: ExecuteStmts=0        with semicolon: ExecuteStmts=1
column rename  no semicolon: RenameFrom=<nil>      with semicolon: RenameFrom=old_name
enum value     no semicolon: ValueRenameFrom=map[] with semicolon: map[a:old_a]
```

`stmtRegion` is the one place that decides where a statement's text ends, and `stmtStart`, the ignored-statement snippet and all six directive scans read it. The clamp it holds is the one the snippet and `stmtStart` already had.

`findLeadingCommentEnd` counted a newline for a final line that has none and clamped the result back to the length. The empty region was what reached that clamp; without it the offset can no longer overrun, so the scan now walks the string it is given and stops at its end. Checked against the old implementation over the fixed cases and 200k random inputs: same answer everywhere.

Reported in review on #497.

## Tests

`TestDirectivesOnTrailingStatementWithoutSemicolon` runs each directive kind against the three endings a file can have, no semicolon, no semicolon and a newline, and a semicolon, and requires the three to agree: `renamed-from`, `ignore`, `concurrently`, `bulk-alter`, `execute`, and the inline column, enum value and composite attribute renames. All eight fail on the two without-semicolon endings before this change. The `execute` case also asserts that no ignored-statement warning is emitted, since the statement is run at apply.

`go test ./parser/...`, `make lint` and `deadcode` pass. Parser coverage is 94.3% against 94.1% on main; `stmtRegion` and `stmtStart` are at 100%, `findLeadingCommentEnd` stays at 100% without its clamp, and the uncovered statements in `directive.go` go from 8 to 5. The five left are pre-existing: two deparse/scan failures that a statement which already parsed cannot reach, a `FormatExecuteStmtWithNote` branch the plan tests cover with a database, and two guards on malformed input. The suites that need PostgreSQL were not run here.

---


winebarrel:
LGTM. This is the fix for what I raised on #497, and it takes the shape that removes the duplication rather than patching each scan.

I checked the branch out and verified the reasoning independently rather than trusting the description:

* **The `findLeadingCommentEnd` rewrite is equivalent.** I ran the new implementation against the old one over the edge cases (empty string, comment-only with and without a trailing newline, CRLF, tabs, invalid UTF-8, no-newline final line) and 200k random strings built from a comments-and-whitespace alphabet. Not one difference. So the clamp really was reachable only through the empty region, and dropping it is safe.
* **`StmtLen == 0` really does mean "final statement, no semicolon".** A final statement that ends with a semicolon carries a length, and a trailing comment after it stays outside its region. In a multi-file parse, a non-final file that ends without a semicolon is a syntax error (now with a location, from #499), so a joined input cannot produce a zero-length statement in the middle. The widening is confined to the case it is meant for.
* **The widened region cannot pick up a directive that is not the statement's own.** Only comments and whitespace can follow a semicolon-less final statement, and the leading-block scans never look past `leadingEnd`. For the two scans that read the whole region, I checked that a `-- pista:renamed-from` written after the closing paren of a `CREATE TABLE` or `CREATE TYPE ... AS ENUM` leaves every column and enum value untouched, so nothing is misattributed.
* **The consolidation is complete.** `StmtLen` now appears in one region computation, `stmtRegion`, plus the copy onto the rebuilt `ALTER TABLE` statement in `warnIgnoredAlterTableCmds`. All five inline copies of the clamp are gone.
* `go test ./parser/...` and `go vet ./...` are clean here, and CI is green across the 15-18 matrix, windows, samples and both fuzz targets. I could not run `make lint` locally: the golangci-lint in this environment is built with go1.25 and the module targets 1.27.

Two non-blocking notes:

1. The tests stop at the parser. Per the repository guidelines, the user-visible half of this fix, a rename that now plans as `ALTER TABLE ... RENAME TO` instead of a drop and a create, is exactly a plan fixture: one YAML case whose desired schema ends without a semicolon would guard the behavior end to end. The table test is the right home for the eight-by-three matrix, so this would be an addition rather than a replacement.

2. `TestDirectivesOnTrailingStatementWithoutSemicolon` swaps `warnWriter` by hand. `SetWarnWriter` in `export_test.go` is in `package parser` too, so `defer SetWarnWriter(&warnings)()` would do the same in one line.

---

winebarrel:
Both notes addressed in 5822c40.

1. Added `testdata/plan/rename_table_no_trailing_semicolon.yml`: the same case as `rename_table.yml` with the desired schema ending without a semicolon, expecting `ALTER TABLE public.users RENAME TO accounts;`. I cannot run the plan suite here (no PostgreSQL), so I checked the half that runs without one: the fixture's desired text ends at `)` and parses with `RenameFrom=public.users` and the same columns and constraint as `rename_table.yml`. CI runs the rest.

2. `TestDirectivesOnTrailingStatementWithoutSemicolon` now uses `defer SetWarnWriter(&warnings)()`.

---

winebarrel:
Re-reviewed at 5822c40. Both notes are addressed and I found nothing new. Still LGTM.

On the plan fixture, which is the part worth checking rather than reading:

* It is a real guard, not a tautology. I parsed the fixture's desired schema directly against both heads: on main (1a1a858) it yields `RenameFrom=nil`, so the case would plan a drop and a create, and on this branch it yields `public.users`. The fixture fails without the fix.
* It is `rename_table.yml` minus the semicolon, so the expected plan is already pinned by the neighbouring case, and the diff between the two files is exactly the thing under test. That is the right minimal delta.
* The harness picks it up: `plan_test.go` globs `testdata/plan/*.yml`, and the three keys match `planTestCase`'s `init` / `desired` / `plan`.
* The block scalar leaves the desired schema ending in `)\n`, so the fixture exercises the "no semicolon, trailing newline" variant, which is how a file usually ends. The variant with no trailing newline at all is not expressible this way and stays covered by the unit test, so nothing is missing.

`defer SetWarnWriter(&warnings)()` reads better and restores the same way.

`go test ./parser/...` and `go vet ./...` are clean here. On this head, test (15), (17), (18), samples, keywords, ascii, build, CodeQL and both fuzz targets are green; test (16) and windows were still running when I wrote this. I could not run the plan suite locally, since this environment has no PostgreSQL and no docker daemon, so CI is what covers the new fixture end to end.

---

PR #501, opened 2026-09-03
Read a directive that follows a block comment

winebarrel:
`findLeadingCommentEnd` walked the text before a statement line by line and stopped at the first line that was neither blank nor a `--` comment, so a `/* ... */` block ended the scan. A directive written after one was never seen.

On main:

```
directive alone                        RenameFrom=public.old_b
block comment before directive         RenameFrom=<nil>
ignore after block comment             Ignore=false
trailing block comment on prev line    RenameFrom=<nil>
```

```sql
/* renamed in release 3 */
-- pista:renamed-from public.users
CREATE TABLE public.accounts (...);   -- planned a drop and a create
```

A block comment left at the end of the line above the statement does the same, since pg_query counts it as part of the next statement.

The scan now walks the text rather than its lines, skipping whitespace, `--` comments and `/* ... */` comments, the last of these nested the way PostgreSQL nests them. `stmtStart` therefore lands on the statement's first token whatever precedes it: the caret in a warning or a duplicate-name error no longer points at the `/*` line, at the indent, or at the space after the previous statement on the same line. That is the cosmetic half of the note on #497 and #499.

Skipping a block comment means the directive scans can see inside one, so `leadingDirectiveText` blanks a block comment's body before they read it. Commenting a schema out with `/* ... */` therefore comments out the directives in it, and blanking rather than removing keeps the offsets `extractExecuteDirectives` slices a check SQL out of.

`validateDirectives` scans the whole file and still reports an unknown `-- pista:` name inside a block comment. That is unchanged, and it keeps a typo from passing unnoticed.

## Tests

`TestFindLeadingCommentEnd` and `TestBlankBlockComments` cover the two helpers directly, including the inputs the parser cannot produce: an unterminated comment and a comment that ends the string without a newline. `TestDirectivesAroundBlockComments` covers the behavior end to end: a directive after a block comment applies, one inside a block comment does not, a schema commented out with `/* ... */` keeps its directives commented out, and a check SQL after a block comment still comes through. `TestParseSQLFiles_IgnoredStatementWarningSkipsToTheStatement` pins the three caret positions. A plan fixture covers the user-visible half, a rename under a block comment planning `ALTER TABLE ... RENAME TO`.

`go test ./parser/...`, `make lint` and `deadcode` pass. Parser coverage is 94.4% against 94.3% on main, with every new function at 100% and the uncovered statements in `directive.go` unchanged at 5. The suites that need PostgreSQL were not run here; the plan fixture's parse half was checked without one.

---

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,14 @@
 
 ## [Unreleased]
 
+* Read a directive that follows a `/* ... */` comment. The scan for the comments before a statement stopped at the first block comment, so a directive written after one was never seen: a `-- pista:renamed-from` under a block comment planned a drop and a create instead of a rename, and a block comment left at the end of the line above did the same. Both comment forms are now skipped, nested block comments included, and a directive inside a block comment stays commented out. The position an error or a warning points at skips them as well, so it lands on the statement rather than on the `/*` line or on the indent.
+
 * Apply a directive on a file's last statement when it ends without a semicolon. pg_query reports no length for such a statement, so the scans that carve out a statement's text got an empty region: `-- pista:renamed-from` planned a drop and a create instead of a rename, `-- pista:ignore` left the object managed, and `-- pista:execute` left the statement to the ignored-statement warning instead of running it at apply. `-- pista:concurrently`, `-- pista:bulk-alter` and the inline column, enum value and composite attribute renames went the same way. The scans now share one helper, which reads such a statement to the end of the input.
 
 * Name the file, line and column in the ignored-statement warning. It carried the statement text alone, which does not say where the statement is when the schema spans several files, or when the same `GRANT` appears in more than one of them. It now reads `pistachio: schema/items.sql:12:1: ignored unsupported statement: DROP TABLE public.items`. A parse that came from no file names no position.
+
 * Read the desired schema before connecting. `plan` and `apply` opened the connection first, so a missing file, a syntax error or a duplicate name was reported only once the database answered, and a run against an unreachable database reported the connection rather than the file. Both now read and parse the schema files, and resolve `--pre-sql-file` and `--concurrently-pre-sql-file`, before connecting. A successful run is unchanged; one that cannot read its own files opens no connection, and takes no lock under `--exclusive`.
+
 * Name the object and the position in a duplicate-name error. `duplicate column: id` said neither which table the column was on nor where the file repeats it. It now reads `duplicate column: id on public.users` and prints the line with a caret under it, as a syntax error does. A column or constraint is pointed at where it is defined, everything else at the statement that repeats the name. The checks that report a clash between two objects, a table and a view of one name for example, are unchanged.
 
 ## [1.40.0] - 2026-09-04
--- a/docs/reference/directives.md
+++ b/docs/reference/directives.md
@@ -1,6 +1,6 @@
 # Directives
 
-pistachio reads directives from SQL comments in schema files. A directive is a line comment of the form `-- pista:<name>` placed on its own line immediately before the target statement. Unknown directive names are rejected at parse time. A directive placed before a statement it does not apply to is ignored.
+pistachio reads directives from SQL comments in schema files. A directive is a line comment of the form `-- pista:<name>` placed on its own line before the target statement. Blank lines and further comments of either form may come between the two, so a `/* ... */` note above the statement does not detach the directive from it. A directive written inside a `/* ... */` comment is commented out and does not apply, but it is still checked, so a typo or a stray argument in one fails the parse rather than passing unnoticed. Unknown directive names are rejected at parse time. A directive placed before a statement it does not apply to is ignored.
 
 | Directive | Arguments | Applies to | Purpose |
 |---|---|---|---|
--- a/parser/directive.go
+++ b/parser/directive.go
@@ -91,9 +91,7 @@ func extractExecuteDirectives(rawSQL string, stmts []*pg_query.RawStmt) ([]*Exec
 
 	for _, stmt := range stmts {
 		loc := stmt.StmtLocation
-		region := stmtRegion(rawSQL, stmt)
-		leadingEnd := findLeadingCommentEnd(region)
-		leading := region[:leadingEnd]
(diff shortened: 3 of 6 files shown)


winebarrel:
LGTM. This covers both halves of what I raised: the functional one, a directive after a block comment silently dropped, and the cosmetic one from #497 and #499, the caret landing on the `/*` line or on the indent.

The scanner is the risky part of the change, so I tested it against the authority rather than reading it. For 20 leading-comment prefixes I parsed `prefix + "SELECT 1;"` with pg_query and compared `stmtStart` against `len(prefix)`: a successful parse proves PostgreSQL's own lexer treats the whole prefix as comment, so the two must agree on where it ends. Every case does, nesting (`/* /* nested */ */`, `/* /* /* deep */ */ */`), the shapes that trip hand-rolled scanners (`/*/ */`, `/**//**/`, `/* * / */`), a `--` holding a `/*` and a block holding a `--`, and non-ASCII inside a comment. That is the property that matters here: end a comment one byte off from PostgreSQL and a directive scan starts reading statement text.

The rest I checked:

* The blanking is safe by construction, which is the thing that would worry me about a byte-level scan. `leadingDirectiveText` blanks only `region[:findLeadingCommentEnd(region)]`, and that slice holds nothing but whitespace and comments, so `blankBlockComments` can never meet a string literal or a dollar-quoted body where a bare `/*` would be misread.
* Blanking preserves length, which the test asserts directly, so the offsets `extractExecuteDirectives` slices a check SQL out of still line up. The "check SQL after a block comment" case covers that end to end.
* `blockCommentEnd` returns -1 for an unterminated comment and both callers fall back to the end of the input. Unreachable through the parser, since pg_query rejects such a file, and the helper test covers it anyway.
* The plan fixture is a real guard, not a restatement: its desired half yields `RenameFrom=nil` on main (aca1e36) and `public.users` on this branch.
* `go test ./parser/...` and `go vet ./...` are clean here, and CI is green across the 15-18 matrix, windows, samples and both fuzz targets.

Three non-blocking notes:

1. `extractInlineDirectives` is still line-based, so the rule this PR establishes, that a `/* ... */` comments out the directives inside it, stops at the statement boundary. Most shapes inside a `CREATE TABLE` body are harmless because the pending rename is absorbed by the `*/` line, but one leaks:

   ```sql
   CREATE TABLE public.b (
       id integer,
       /*
       -- pista:renamed-from old_name */
       other text
   );
   ```

   `other` comes out with a `RenameFrom` of `old_name */`. Pre-existing and unchanged by this PR, and it is a corner case rather than something to chase; worth knowing only because it is the one place the new rule does not reach.

2. `docs/reference/directives.md` still describes a directive as "placed on its own line immediately before the target statement" and says nothing about either rule this PR settles: a directive may be separated from its statement by comments of either form, and a directive inside a block comment does not apply but is still name-checked. The second is surprising enough to earn a sentence next to "Unknown directive names are rejected at parse time", since `/* -- pista:ignore something */` fails the parse instead of being a no-op.

3. Beyond the new entry, the CHANGELOG diff inserts blank lines between bullets in released sections (1.10.0, 1.8.0, 1.6.0, 1.5.0, 1.4.0, 1.2.0). Harmless, but it is a whole-file reformat riding along in a bug-fix PR.

---

winebarrel:
Notes 2 and 3 addressed in 609bc28.

2. `docs/reference/directives.md` now says what may come between a directive and its statement (blank lines and comments of either form) and that a directive inside a `/* ... */` comment does not apply while its name is still checked, so a typo in one fails the parse.

3. The CHANGELOG diff is back to the four lines this entry needs. The blank lines in the released sections are gone.

Leaving note 1. `extractInlineDirectives` is line-based and pre-existing, and the one shape that leaks, `-- pista:renamed-from old_name */` inside a block comment in a table body, fails loudly at plan time with a rename source that does not exist rather than renaming the wrong thing.

---

winebarrel:
Re-reviewed at 609bc28. Both actionable notes are addressed and I found nothing new. Still LGTM.

`git diff 18312f3 609bc28 -- parser/ testdata/` is empty, so the scanner, the blanking and the plan fixture are byte for byte what I verified last round: the differential check against PostgreSQL's own lexer over 20 comment shapes, the blanking-is-safe-by-construction argument, and the fixture yielding `RenameFrom=nil` on main. Only the two documents changed, and I checked those:

* **CHANGELOG.** The released sections are back to their original spacing, so the diff against main is now the new entry alone. The two blank lines left inside `## [Unreleased]` make that section internally consistent, which is the section the PR is editing, so that reads right to me.
* **docs/reference/directives.md.** Dropping "immediately" and naming both rules is accurate against the behavior: I re-checked that a directive separated from its statement by a `/* ... */` applies, that one inside a block comment does not, and that `/* -- pista:foo bar */` still fails the parse. ASCII-only, and the `ascii` job agrees.

One optional refinement, take it or leave it: "its name is still checked" is narrower than what happens. Argument validation fires inside a block comment too, so `/* -- pista:ignore something */` fails with `-- pista:ignore does not accept arguments` rather than an unknown-name error. "It is still checked" would cover both without adding a sentence.

The third note from last round, the line-based `extractInlineDirectives`, was never a request on this PR and remains as it was.

`go test ./parser/...` and `go vet ./...` are clean here. On this head the jobs that do not need the database, ascii, keywords, build, CodeQL and both Analyze jobs, are green, and the test matrix, samples, windows and the fuzz targets were still re-running when I wrote this; they were fully green on 18312f3, whose code this head does not change.

---

winebarrel:
Took the refinement, in 18b4cd8: "but it is still checked, so a typo or a stray argument in one fails the parse rather than passing unnoticed."

Confirmed the behavior first. A directive on its own line inside a block comment is checked both ways: `/*\n-- pista:foo\n*/` gives `unknown directive: -- pista:foo`, and `/*\n-- pista:ignore something\n*/` gives `-- pista:ignore does not accept arguments`. The single-line `/* -- pista:ignore something */` from the note is not checked at all, since the patterns want the `--` at the start of a line, and that form is not a directive by the rule the same paragraph opens with.

Only the one sentence changed; `git diff 609bc28 18b4cd8` touches nothing else.

---

PR #503, opened 2026-09-04
Poll for the apply exclusion instead of blocking on it

winebarrel:
`--exclusive-wait` waited by running `pg_advisory_lock`, which blocks in the server until the holder is gone. That statement holds a snapshot for as long as it waits, and the apply it waits for may run `CREATE INDEX CONCURRENTLY`, whose build waits for every backend holding an older snapshot, whatever table that backend is touching.

The two waits close a cycle: the build waits for the waiter's snapshot, and the waiter waits for the exclusion the build's session holds. Advisory locks sit in the same lock manager as table locks, so the deadlock detector sees it and breaks it, and the session it kills is the waiter, which was doing nothing but wait.

```
apply A  holds the exclusion, runs CREATE INDEX CONCURRENTLY
         --[ ShareLock on B's virtualxid ]--> apply B
apply B  blocked in pg_advisory_lock, holding a snapshot
         --[ advisory lock ]-----------------> apply A
```

On main, with a holder that builds an index concurrently:

```
pistachio: failed to acquire the apply exclusion: ERROR: deadlock detected (SQLSTATE 40P01)
```

a second after the wait started, with the other apply still making progress and minutes of the wait budget left.

The wait now retries `pg_try_advisory_lock` on a one-second ticker. An attempt is a single fast statement, so between attempts the session is idle and holds no snapshot, and there is nothing for a concurrent index build to wait on: the cycle cannot form.

The deadline, the `0` unlimited wait, the message a wait that runs out reports and the one a canceled context reports are unchanged. What changes is that a lock that frees is picked up within a second rather than at once, and that the queue is no longer first-come-first-served, neither of which matters for a wait measured in minutes.

## Tests

`TestApplyExclusiveWaitDuringConcurrentIndexBuild` reproduces the deadlock end to end. It rests on how a given PostgreSQL decides which backends an index build has to wait for, so `TestApplyExclusiveWaitHoldsNoSnapshot` asserts the property the fix rests on instead: while it waits, the session reaches `pg_stat_activity` `idle` with a null `backend_xmin`. Both fail on the blocking wait, the second with `the wait held a snapshot the whole time`, on 15 through 18.

`TestApplyExclusiveWaitConnectionLost` terminates the waiting backend and checks the failure is reported as itself rather than as the wait running out, which also covers the retry loop's error path. `TestApplyExclusiveWaitTimeout` now also pins that a deadline shorter than the poll interval ends on the deadline rather than at the next attempt.

`make lint`, `make test` and `make test-scenario` pass, and the exclusive tests pass on 15, 16, 17 and 18. Package coverage is 98.3%, the same as on main: `tryExclusive` is at 100% and `acquireExclusive` at 96.8%, its one uncovered statement the error return of the first attempt, which was uncovered on main as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Wait for the apply exclusion by retrying rather than by blocking in `pg_advisory_lock`. A blocked statement holds a snapshot while it waits, and the apply it waits for may run `CREATE INDEX CONCURRENTLY`, whose build waits for every backend that holds one. The two waits close a cycle, and since advisory locks share a lock manager with table locks the deadlock detector saw it and killed the waiter: a wait died a second in with `failed to acquire the apply exclusion: ERROR: deadlock detected`, with the other apply still making progress. The wait now retries `pg_try_advisory_lock` once a second, so between attempts the session is idle and holds no snapshot and no cycle can form. The deadline, the unlimited wait and the messages are unchanged; a lock that frees is taken within a second rather than at once, and the queue is no longer first-come-first-served.
+
 ## [1.41.0] - 2026-09-03
 
 * Read a directive that follows a `/* ... */` comment. The scan for the comments before a statement stopped at the first block comment, so a directive written after one was never seen: a `-- pista:renamed-from` under a block comment planned a drop and a create instead of a rename, and a block comment left at the end of the line above did the same. Both comment forms are now skipped, nested block comments included, and a directive inside a block comment stays commented out. The position an error or a warning points at skips them as well, so it lands on the statement rather than on the `/*` line or on the indent.
--- a/docs/guides/exclusive-apply.md
+++ b/docs/guides/exclusive-apply.md
@@ -20,4 +20,6 @@ The exclusion is a session-level [advisory lock](https://www.postgresql.org/docs
 
 Transaction boundaries do not affect a session-level lock, so it works the same with and without `--with-tx` and across `CREATE INDEX CONCURRENTLY`. The lock key includes a hash of the database name; applies to different databases on one cluster do not exclude each other.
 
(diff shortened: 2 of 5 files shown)


winebarrel:
LGTM.

The diagnosis holds up: a backend blocked in `pg_advisory_lock` sits in an active statement with a snapshot, `CREATE INDEX CONCURRENTLY` waits for that snapshot, and the advisory lock puts the waiter's edge into the same wait-for graph, so the detector kills the waiter. Polling `pg_try_advisory_lock` leaves the session idle with no `backend_xmin` between attempts, so the cycle cannot form.

What I checked in `acquireExclusive`:

- The first non-blocking attempt and the `wait == nil` error path are unchanged.
- The `select` on `waitCtx.Done()` keeps the two messages apart: a canceled caller context reports `failed to acquire the apply exclusion`, a spent budget reports `did not finish within`. `0` still waits without limit since `waitCtx` is then just `ctx`.
- Running each attempt under `ctx` rather than `waitCtx` is the right call. It keeps a deadline that lands mid-statement from surfacing as a query error, and the deadline is still honored on the next loop iteration.
- A lost connection surfaces from `tryExclusive` as itself, which `TestApplyExclusiveWaitConnectionLost` pins.

The tests cover the property rather than only the symptom. `TestApplyExclusiveWaitHoldsNoSnapshot` sampling for one `idle` + null `backend_xmin` observation is robust to landing on an attempt, and `TestApplyExclusiveWaitDuringConcurrentIndexBuild` reproduces the original deadlock end to end. Including `items_id_idx` in the desired schema so the apply does not plan a drop of the index the holder built is a nice detail.

One non-blocking note: the new upper bound in `TestApplyExclusiveWaitTimeout` (`elapsed < ExclusivePollInterval`) includes connect time, so it leaves about 700ms of headroom on a 300ms wait. It passed on all four PostgreSQL versions and on the Windows job, so I would leave it as is and only widen the gap if it ever flakes.

Docs and changelog read clearly and stay ASCII. CI is green across the matrix.

---

PR #505, opened 2026-09-05
Restore a NOT VALID check constraint unvalidated

winebarrel:
Closes the TODO entry "A NOT VALID CHECK constraint is restored validated".

A table's check constraints were written inside `CREATE TABLE`, where `NOT VALID` cannot be spelled, so `dump` printed such a constraint as validated and the reload scanned the table to validate it. A plan that created the table had the same gap: the constraint was applied validated, and every later plan re-emitted DROP+ADD forever.

## Changes

- `Table.SQL` leaves a NOT VALID check out of `CREATE TABLE`; the new `Table.NotValidConSQL` renders it as its own `ALTER TABLE ONLY ... ADD CONSTRAINT ... NOT VALID`, the way a foreign key already is. Both `dump` and the new-table plan path emit it.
- `ONLY` is dropped on a partitioned table, which rejects it once a partition exists (verified on 15). A partition child renders nothing: its unvalidated entries are the clones the parent's statement creates. An INHERITS child keeps the inline form, since switching it to ALTER would collide with the constraint it inherits.
- `test/fidelity/schemas/check_constraints.sql` gets its NOT VALID line back, which is the regression test the TODO entry named.
- The gitlab sample round-trips without `--assume-validated` (verified against the live sample: 68 NOT VALID constraints, `-- No changes`), so the flag comes off its SAMPLES record and `docs/contributing-samples.md` follows.

## Tests

- Fixtures: dump (two NOT VALID checks, output order pinned), dump with `--omit-schema`, plan for a new table, plan round trip of the dump form, apply for a new table and for a partitioned parent. Every apply fixture re-plans and requires no drift.
- Package tests in `model` and `diff` keep per-package coverage where it was (model 99.8%, diff 97.5%).
- `make test`, `make test-scenario`, `make test-fidelity` (30 passed), `make lint` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Restore a `NOT VALID` check constraint unvalidated. A table's checks are written inside `CREATE TABLE`, where the clause cannot be spelled, so `dump` printed such a constraint as validated and the reload scanned the table to validate it. Such a check is now left out of `CREATE TABLE` and added by its own `ALTER TABLE ... ADD CONSTRAINT ... NOT VALID`, the way a foreign key is. This also stops a plan that created the table from re-emitting the constraint on every later run.
+
 ## [1.41.1] - 2026-09-04
 
 * Wait for the apply exclusion by retrying rather than by blocking in `pg_advisory_lock`. A blocked statement holds a snapshot while it waits, and the apply it waits for may run `CREATE INDEX CONCURRENTLY`, whose build waits for every backend that holds one. The two waits close a cycle, and since advisory locks share a lock manager with table locks the deadlock detector saw it and killed the waiter: a wait died a second in with `failed to acquire the apply exclusion: ERROR: deadlock detected`, with the other apply still making progress. The wait now retries `pg_try_advisory_lock` once a second, so between attempts the session is idle and holds no snapshot and no cycle can form. The deadline, the unlimited wait and the messages are unchanged; a lock that frees is taken within a second rather than at once, and the queue is no longer first-come-first-served.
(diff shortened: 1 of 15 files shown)


winebarrel:
LGTM.

What I verified:

- `Table.SQL` / `inlineConstraintDefs` / `NotValidConSQL` split is straightforward and mirrors the existing FK path. `TableToSQL` (dump) and `newTableExtras` (new-table plan) are the only two emitters, and both pick it up.
- `ONLY` handling matches PostgreSQL: `ALTER TABLE ONLY` on a partitioned table is rejected once a partition exists, and the partition-child early return is right because `diffTable` already skips constraint diffs on partition children, so the clones never drift.
- `make test` passes for the change (the only failures on my machine are four auth-related tests that hardcode a passwordless connection string, unrelated to this PR). `test/fidelity/run.sh` reports 30 passed, 0 failed, including the restored `NOT VALID` line in `check_constraints.sql`.
- `--bulk-alter` on a new table with two `NOT VALID` checks and a comment: the two `ALTER TABLE ONLY ... NOT VALID` statements are left unmerged, consistent with how FK adds are treated by `parseMergeableAlterTable`. Apply then re-plan gives `-- No changes`.
- Fed the dump of a schema with a `NOT VALID` check on a partitioned parent back to `plan`: clean.

One non-blocking observation, for the record rather than for this PR: a `NOT VALID` check declared directly on an `INHERITS` child (not inherited from the parent) still dumps in the inline form and re-plans as `VALIDATE CONSTRAINT`. This is identical on `main`, so it is not a regression, and the PR description already explains why the INHERITS branch is left alone. It may be worth a one-line TODO entry so the fidelity check's silence on it is a recorded decision.

---

winebarrel:
Thanks for the review. Reproduced the INHERITS-child case on 15: the child's NOT VALID check dumps inline and the dump fed back plans `VALIDATE CONSTRAINT` for it, as you observed. Recorded it as a TODO entry in c7cf751, including why the ALTER form cannot be used there yet: the child's constraint map also holds the unvalidated clones a NOT VALID check on the parent pushes down, and telling those apart from the child's own checks needs `pg_constraint.conislocal`, which the catalog does not read today.

PR #506, opened 2026-09-05
Track UNLOGGED on a standalone sequence

winebarrel:
Closes the TODO entry "UNLOGGED sequence support".

An `UNLOGGED SEQUENCE` was read by neither side: `dump` wrote it as a plain `CREATE SEQUENCE`, and a `LOGGED` <-> `UNLOGGED` change in the desired schema produced no diff.

## Changes

Follows the table pattern, as the TODO entry laid out:

- `model.Sequence` gains an `Unlogged` field; `SQL()` renders `CREATE UNLOGGED SEQUENCE`.
- The catalog reads `pg_class.relpersistence = 'u'`.
- The parser reads the persistence off `CreateSeqStmt`.
- The diff emits `ALTER SEQUENCE ... SET LOGGED/UNLOGGED` for a transition. The SET form is its own statement, since the option list of `ALTER SEQUENCE` does not take it; a plan that also changes options emits both, persistence first.

Temporary sequences stay out of scope: session-local, never dumped.

## Tests

- Fixtures: dump, plan (create, both transitions, transition combined with an option change, no diff), apply (create and transition; both re-plan and require no drift).
- Package tests in model, diff, parser, and catalog (live read of an unlogged and a plain sequence).
- `test/fidelity/schemas/sequences.sql` gains an unlogged sequence, the regression test for the dump side.
- Coverage: model 99.8%, parser 94.4%, catalog 93.1% unchanged; diff 97.5% -> 97.6% (the `MINVALUE` / `START WITH` clauses of `diffSequence` were uncovered in the package before and are now exercised).
- `make test`, `make test-scenario`, `make test-fidelity` (30 passed), `make lint` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Track `UNLOGGED` on a standalone sequence. It was read by neither side, so `dump` wrote an unlogged sequence as a plain `CREATE SEQUENCE` and a `LOGGED` <-> `UNLOGGED` change produced no diff. `dump` now writes `CREATE UNLOGGED SEQUENCE` and a change goes out as `ALTER SEQUENCE ... SET LOGGED/UNLOGGED`. Temporary sequences stay out of scope.
+
 * Restore a `NOT VALID` check constraint unvalidated. A table's checks are written inside `CREATE TABLE`, where the clause cannot be spelled, so `dump` printed such a constraint as validated and the reload scanned the table to validate it. Such a check is now left out of `CREATE TABLE` and added by its own `ALTER TABLE ... ADD CONSTRAINT ... NOT VALID`, the way a foreign key is. This also stops a plan that created the table from re-emitting the constraint on every later run.
 
 ## [1.41.1] - 2026-09-04
--- a/TODO.md
+++ b/TODO.md
@@ -271,19 +271,6 @@ and is not pursued.
 
 Origin: discussion during [#124](https://github.com/winebarrel/pistachio/pull/124). No current plan to implement.
 
-## UNLOGGED sequence support
-
-Standalone sequence management does not track the `UNLOGGED` attribute. An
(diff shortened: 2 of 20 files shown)


winebarrel:
LGTM.

What I verified:

- The change follows the table pattern end to end: `model.Sequence.Unlogged`, `relpersistence = 'u'` in the catalog, `Relpersistence` off `CreateSeqStmt` in the parser, and `SET LOGGED` / `SET UNLOGGED` as its own statement in `diffSequence`. Emitting the persistence change before the option clause is the right order, and rename detection is unaffected since `detectSequenceRenames` copies the current sequence and only swaps the name.
- `go vet` and the `model`, `parser`, `diff`, `catalog` packages plus the `TestPlan` / `TestApply` / `TestDump` fixture suites pass. `test/fidelity/run.sh` reports 30 passed, 0 failed, with `sequences` covering the new `CREATE UNLOGGED SEQUENCE` line.
- Live probe: an unlogged sequence with `CACHE 10` and a comment dumps as `CREATE UNLOGGED SEQUENCE ... CACHE 10;` and the dump re-plans as `-- No changes`. A desired schema that flips it to logged, changes `CACHE`, changes the comment, and at the same time renames a second sequence with `-- pista:renamed-from` while making it unlogged plans as rename, `SET LOGGED`, `CACHE 20`, `COMMENT`, `SET UNLOGGED`, and after apply `pg_class.relpersistence` is `p` / `u` as expected with a clean re-plan.
- `ALTER SEQUENCE ... SET LOGGED/UNLOGGED` exists from PostgreSQL 15, which is the floor of the CI matrix, so no version gate is needed.

One optional nit: `docs/reference/objects.md` says "Tables (including unlogged and partitioned tables)" but the Sequences bullet does not mention unlogged. A few words there would keep the two in step, but it is not needed for this PR.

---

winebarrel:
Thanks for the review. Updated the Sequences bullet in docs/reference/objects.md to mention unlogged sequences, matching the Tables bullet (daa0f94).

PR #507, opened 2026-09-05
Normalize an EXCLUDE constraint's elements and predicate

winebarrel:
Closes the TODO entry "An EXCLUDE constraint is not normalized at all".

`equalConstraintDef` normalized `RawExpr`, where a CHECK body sits. An exclusion constraint puts its elements in `Exclusions` and its predicate in `WhereClause`, so the two sides were compared as the parser left them, and a hand-written exclusion was dropped and re-added on every plan for a difference the CHECK path already folds. A dump fed back was already clean, since dump writes the catalog form.

## Changes

- `normalizeExclusion` runs each element expression and the predicate through the same walk (`normalizeCheckExpr`).
- The element canonicalization (explicit ASC, default NULLS order) is shared with the index path as `normalizeIndexElem`.
- An operator's qualification is dropped from both sides, the way `stripFuncSchema` treats a function name: the catalog prints a visible operator bare.
- `alignExclusionCasts` strips the casts only the catalog wrote, pairing element for element and predicate for predicate the way a CHECK body is paired.

## Tests

- Each drift shape was reproduced on 15 before being pinned as a no-diff fixture: schema-qualified function, qualified operator, text cast, BETWEEN in the predicate, numeric cast in the predicate, explicit ASC, and EXCLUDE without USING (already clean, pinned).
- Negative cases: a predicate change still plans DROP + ADD (fixture), and operator, element-count, and desired-side-cast differences still compare unequal (package tests in `diff`).
- Coverage: diff stays at 97.6%. The only uncovered new statements are two nil guards; the exclusion AST always carries `[IndexElem, operator list]`, so they are defensive, in the shape rename.go already uses.
- `make test`, `make test-scenario`, `make test-fidelity` (30 passed), `make lint` all pass. Dump output is unchanged, so no fidelity schema was added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Normalize an EXCLUDE constraint's elements and predicate the way a CHECK body is normalized. They were compared as written, so a hand-written exclusion drifted on a schema-qualified function or operator, a cast the catalog adds, a BETWEEN in the predicate, or an explicit ASC, and was dropped and re-added on every plan. A dump fed back was already clean; only a desired schema written another way hit this.
+
 * Track `UNLOGGED` on a standalone sequence. It was read by neither side, so `dump` wrote an unlogged sequence as a plain `CREATE SEQUENCE` and a `LOGGED` <-> `UNLOGGED` change produced no diff. `dump` now writes `CREATE UNLOGGED SEQUENCE` and a change goes out as `ALTER SEQUENCE ... SET LOGGED/UNLOGGED`. Temporary sequences stay out of scope.
 
 * Restore a `NOT VALID` check constraint unvalidated. A table's checks are written inside `CREATE TABLE`, where the clause cannot be spelled, so `dump` printed such a constraint as validated and the reload scanned the table to validate it. Such a check is now left out of `CREATE TABLE` and added by its own `ALTER TABLE ... ADD CONSTRAINT ... NOT VALID`, the way a foreign key is. This also stops a plan that created the table from re-emitting the constraint on every later run.
--- a/TODO.md
+++ b/TODO.md
@@ -652,7 +652,8 @@ normalization of any kind beyond the walk.
 
 A function moved between two schemas is the cost of the symmetric strip:
 `a.f(v)` and `b.f(v)` compare equal, so the move produces no diff. This is the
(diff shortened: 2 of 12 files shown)


winebarrel:
LGTM.

What I verified:

- The shape matches the CHECK path exactly: normalize both sides, then strip only the casts the catalog added from the current side, element for element and predicate for predicate. Extracting `normalizeIndexElem` from `normalizeIndexStmt` is a pure move, and dropping an operator's qualification on both sides is consistent with how `stripFuncSchema` treats a function name. Since `detectConstraintRenames` and the domain diff go through `equalConstraintDef` too, a renamed or domain-side exclusion gets the same treatment for free.
- `go vet`, the `diff` / `parser` / `model` packages, the `TestPlan` / `TestApply` / `TestDump` fixture suites, `make lint` (0 issues), and `test/fidelity/run.sh` (30 passed, 0 failed) all pass.
- Live probe with `btree_gist`, a hand-written schema that the fixtures do not cover:
  - `EXCLUDE USING gist (room_id WITH =, tstzrange(s, e, '[)') WITH &&) WHERE (price BETWEEN 1 AND 100)`, which the catalog stores with a `'[)'::text` cast, the BETWEEN expanded, and `(1)::numeric` / `(100)::numeric` casts.
  - `EXCLUDE USING btree (lower(name) DESC NULLS FIRST WITH =)`, which the catalog prints as `lower(name) DESC`.
  - `EXCLUDE USING btree (name text_pattern_ops WITH =)`, an explicit non-default opclass.

  On `main` the first two are dropped and re-added on every plan. On this branch the plan is `-- No changes`, and the opclass case stays clean on both.

One non-blocking note for the record: the operator strip keeps only the last name part regardless of schema, so `OPERATOR(a.=)` and `OPERATOR(b.=)` would compare equal. That is the same trade-off `stripFuncSchema` already makes for functions, and two same-named operators in different schemas on one column is not a schema people write, so I would not change it.

---

winebarrel:
Thanks for the review, and for probing the DESC NULLS FIRST and opclass cases beyond the fixtures. Recorded the operator-strip tradeoff in the TODO passage that already documents it for function names (see the schema-qualified sequence entry), so it stays discoverable next to the same decision for functions. The two lines Codecov flags are the nil guards on the exclusion list; the AST always carries [IndexElem, operator list], so they are defensive only.

PR #508, opened 2026-09-05
Read the storage parameters of the index a constraint owns

winebarrel:
Closes the TODO entry "A constraint's index storage parameters are not read".

`ALTER TABLE ... ADD CONSTRAINT ... UNIQUE (v) WITH (fillfactor = 70)` puts the parameters on the index the constraint owns, and `pg_get_constraintdef` does not print them for a primary key or a unique constraint. The current side read `UNIQUE (v)`, the two sides never matched, and the plan dropped and re-added the constraint on every run, rebuilding the index under an ACCESS EXCLUSIVE lock. `dump` dropped the clause the same way.

## Changes

- The catalog reads `pg_class.reloptions` under `pg_constraint.conindid` and appends the `WITH` clause to the definition, values quoted the way `pg_get_indexdef` quotes an index's. The clause goes before the deferral suffix, which the grammar takes last; appended after it, a deferrable constraint's definition does not parse (found while writing tests, verified on 15).
- An exclusion constraint gets no append: `pg_get_constraintdef` prints its parameters itself (verified on 15). A foreign key's `conindid` names the referenced table's index, which is not part of the key, so the CASE keeps it out too.
- `equalConstraintDef` runs `normalizeStorageParams` over the constraint's options. The function also folds a bare-word value into the String its quoted spelling parses as: `deduplicate_items = off` arrives as a single-name TypeName and `'off'` as a String, a mismatch that also made a plain index written with the quoted spelling drift.

Not gated behind `--manage-storage-param` (confirmed with the repository owner): no SET or RESET is emitted, the parameters are part of the definition the way a plain index's already are, and gating the read would leave the drift in place when the flag is off. The flag keeps gating a table's own parameters. A parameter set on a constraint's index outside the schema file now shows up as a diff, the same as on a plain index.

## Tests

- Fixtures (each confirmed failing first): no-diff plans for PK / UNIQUE / EXCLUDE / deferrable UNIQUE / two params in reversed order / quoted bare word on a plain index; a parameter change and a parameter removal both plan DROP + ADD; a rename with matching parameters stays a rename; dump pins the two-param format; applies for the plain and the deferrable case, both re-plan drift-free.
- Package tests: catalog (live read asserting the appended definitions, the deferral insert position, and no double append on an exclusion) and diff (`equalConstraintDef` on quoting, order, value change, presence).
- Coverage: diff 97.6% unchanged, catalog 93.1% -> 93.3%; the only uncovered lines in the touched files are two pre-existing scan-error paths.
- `make test`, `make test-scenario`, `make test-fidelity` (30 passed; `storage_params.sql` gains a PK and a UNIQUE with parameters), `make lint` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Read the storage parameters of the index a constraint owns. `ALTER TABLE ... ADD CONSTRAINT ... UNIQUE (v) WITH (fillfactor = 70)` puts them on the index, which `pg_get_constraintdef` does not print, so the two sides never matched and the constraint was dropped and re-added on every plan, rebuilding the index under an ACCESS EXCLUSIVE lock. `dump` dropped the clause the same way and now writes it. A parameter set on such an index outside the schema file shows up as a diff, the same as on a plain index. A quoted parameter value and the bare spelling, `deduplicate_items = 'off'` and `off`, also compare equal now, on a plain index too.
+
 * Normalize an EXCLUDE constraint's elements and predicate the way a CHECK body is normalized. They were compared as written, so a hand-written exclusion drifted on a schema-qualified function or operator, a cast the catalog adds, a BETWEEN in the predicate, or an explicit ASC, and was dropped and re-added on every plan. A dump fed back was already clean; only a desired schema written another way hit this.
 
 * Track `UNLOGGED` on a standalone sequence. It was read by neither side, so `dump` wrote an unlogged sequence as a plain `CREATE SEQUENCE` and a `LOGGED` <-> `UNLOGGED` change produced no diff. `dump` now writes `CREATE UNLOGGED SEQUENCE` and a change goes out as `ALTER SEQUENCE ... SET LOGGED/UNLOGGED`. Temporary sequences stay out of scope.
--- a/TODO.md
+++ b/TODO.md
@@ -578,31 +578,6 @@ clean and only a hand-written view reaches this.
 
 Origin: review of [#458](https://github.com/winebarrel/pistachio/pull/458).
 
-## A constraint's index storage parameters are not read
-
-Priority: low.
-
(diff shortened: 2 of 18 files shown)


winebarrel:
LGTM.

What I verified:

- The catalog read is minimal and correct: a `LEFT JOIN` on `conindid`, and the `CASE` on `contype IN ('p','u')` keeps the exclusion (whose parameters `pg_get_constraintdef` already prints) and the foreign key (whose `conindid` is the referenced table's index) out. `ListIndexes` already excludes constraint-owned indexes, so the parameters are read in exactly one place.
- The deferral insert is sound. `pg_get_constraintdef` prints only `DEFERRABLE` and `DEFERRABLE INITIALLY DEFERRED` as trailing clauses on a PK or UNIQUE, so the two suffixes tried are the complete set, and `NOT VALID` is already trimmed before this runs.
- The `TypeName` fold in `normalizeStorageParams` matches the grammar: an unreserved keyword such as `off` reaches `def_arg` as a `TypeName`, while a reserved one such as `on` and a quoted value arrive as a `String`, which is why only the `off` spelling drifted before.
- `go vet`, the `diff` / `parser` / `model` / `catalog` packages, the `TestPlan` / `TestApply` / `TestDump` fixture suites, `make lint` (0 issues), and `test/fidelity/run.sh` (30 passed, `storage_params` included) all pass.
- Live probe with shapes the fixtures do not cover, all planning `-- No changes` both from the hand-written file and from its own dump:
  - `PRIMARY KEY (id) INCLUDE (w) WITH (fillfactor = 90)`, so the clause lands after `INCLUDE` as the grammar wants.
  - `UNIQUE NULLS NOT DISTINCT (v) WITH (fillfactor = 70, deduplicate_items = on) DEFERRABLE INITIALLY DEFERRED`, exercising the reserved-word value and the deferral insert together.
  - A PK with `fillfactor` on a partitioned parent with one partition; dump writes the clause on the parent only.
  - `--bulk-alter` adding both constraints to an existing table merges them into one `ALTER TABLE`, and the apply re-plans clean.

---

PR #510, opened 2026-09-05
Manage an INHERITS child as the columns and constraints it declares

winebarrel:
`pista dump` wrote an `INHERITS` child as its constraints alone, the parent's included, and dropped every column. A child declared as

```sql
CREATE TABLE child (extra text) INHERITS (parent);
```

came back as

```sql
CREATE TABLE public.child(
    CONSTRAINT parent_n_chk CHECK (n > 0))
INHERITS (public.parent);
```

That output does not reload: the constraint names a column the statement no longer declares, so `CREATE TABLE` fails with `column "extra" does not exist`. A child could not be created from a schema file at all.

Fed back as the desired schema it planned a `DROP COLUMN` for every column, the child's own included, and a `DROP CONSTRAINT` PostgreSQL rejects:

```
ALTER TABLE public.child_own DROP CONSTRAINT parent_n_check;
ALTER TABLE public.child_own DROP COLUMN id;
ALTER TABLE public.child_own DROP COLUMN n;
```

Under `--allow-drop`, applying a dump of your own database dropped the columns. The design policy calls a break in that round trip a bug.

## What changed

Both sides now carry the local half alone, which is what the child's `CREATE TABLE` writes and what `pg_dump` writes for it.

- The catalog filters columns on `pg_attribute.attislocal` and constraints on `pg_constraint.conislocal`, so the ones the child merely inherits belong to the parent and are read from neither side. A column the child redeclares is local and inherited at once, and stays.
- `Table.SQL` renders the child through the plain branch, columns and all, with `INHERITS` after the column list. A child that declares nothing writes the parentheses empty, the way `pg_dump` does.
- `NotValidConSQL` now covers it: a NOT VALID check on the child is added by its own `ALTER` rather than written inline, so a reload no longer validates it and no longer scans the table.

A partition child is unaffected. It still keeps its inherited columns, since its branch of the diff skips the column comparison but compares their comments.

Closes the INHERITS plan / apply and the INHERITS NOT VALID entries in `TODO.md`.

## Why it was not caught

`test/fidelity` had no INHERITS schema. `schemas/inherits.sql` is what catches it, and it fails with `FAIL (reload)` without the change.

Twenty fixtures in all. The two that matter most are the parent-side ones: a column or a check added to the parent is pushed down to every child by PostgreSQL, so the plan must name the parent alone, which is what the two filters exist for. Also covered: the child losing its own column, a column rename on the child (which rewrites the index over it, a block partition children skip), a NOT VALID check added to an existing child, an index and a foreign key on a child, three levels of inheritance, a child added to an existing parent, and parent and child dropped in one run, which has to drop the child first.

## Verification

`make test`, `make lint`, `make test-scenario` and `make test-fidelity` (31 passed) are green on PG15.

Against the ledgersmb sample, the only one that uses `INHERITS` (21 children over 158 tables), the dump fed back with `--allow-drop all` plans `-- No changes`. A real child now matches the shape `pg_dump` writes.

Coverage does not decrease: `catalog` 93.3% -> 93.5%, every other package unchanged. The new code is fully covered. What remains uncovered in the two catalog readers is the scan error paths and the PG18-only NOT NULL name strip, neither reachable naturally on 15.

## Left unmanaged

Both are symmetric on the two sides, so nothing drifts and a dump feeds back clean. `TODO.md` records them and `schemas/inherits.sql` notes what it leaves out.

- A comment on an inherited column. `pg_dump` writes it and that line is dropped on the way in.
- A child of more than one parent. Only `inhseqno = 1` is read, so `INHERITS (a, b)` is dumped as `INHERITS (a)`. Pre-existing; closing it needs `PartitionOf` to hold a list.

## Also

`CHANGELOG.md` gets an `## [Unreleased]` section, and `AGENTS.md` now says to write the entry there rather than at release time.


winebarrel:
Looks good overall. One documentation sentence needs correcting before merge; the code itself I would merge as is.

What I verified locally against PostgreSQL 16:

- `go vet`, the `diff` / `parser` / `model` / `catalog` packages, the `TestPlan` / `TestApply` / `TestDump` fixture suites, `make lint` (0 issues), and `test/fidelity/run.sh` (31 passed, `inherits` included) all pass.
- The `attislocal` / `conislocal` filters are scoped by `inheritsChildOIDs`, so a partition child still keeps its inherited columns and the comment diff on them is unchanged. `validate.go` already skips INHERITS children, so an index on an inherited column (`CREATE INDEX child_id_idx ON public.child (id)`) parses, plans, applies, and re-plans clean.
- A `NOT VALID` check added to the parent after the fact is dumped on the parent alone and the dump fed back plans `-- No changes`, so the clone-collision the old TODO entry described is gone.
- `pg_dump -s` of `schemas/inherits.sql`, with the `\restrict` / `\unrestrict` lines a current pg_dump emits stripped, fed to `pista plan --allow-drop all` gives `-- No changes`. So the child really does match the shape pg_dump writes now.

The one thing to fix. `docs/reference/objects.md` says "Writing an inherited column on the child plans nothing." That is not what happens. With `parent (id, n)` and a child created as `child (extra) INHERITS (parent)`, a desired schema that redeclares the column:

```sql
CREATE TABLE public.child (n integer NOT NULL, extra text) INHERITS (public.parent);
```

plans `ALTER TABLE public.child ADD COLUMN n integer NOT NULL;`, which PostgreSQL rejects with `column "n" of relation "child" already exists`, and the same statement comes back on every later plan. `ADD COLUMN` on a child does not merge with an inherited column; only `CREATE TABLE` and a parent-side `ADD COLUMN` do. The reverse holds too: a child that redeclared `n` in the database while the desired schema omits it plans `DROP COLUMN n` under `--allow-drop column`, which fails with `cannot drop inherited column "n"`. Both come from the same place: `attislocal` separates the two on the current side, but the desired side can only say "declared", so a flip of `attislocal` in either direction has no DDL the diff can emit. On `main` the first case planned `ALTER COLUMN n SET NOT NULL` (alongside the bogus drops), so it is a small change in a hand-written shape.

By the design policy this is low priority, since the dump round trip is clean and a redeclared column dumps as declared. I would not widen this PR for it. I would change the sentence to say what happens (a column redeclared on an existing child cannot be planned, and neither can dropping the redeclaration), and record the two shapes in `TODO.md` next to the inherited-column comment entry, which names the same prerequisite: the inherited column set materialised on the child, kept apart from the local one.

---

winebarrel:
Thanks, you are right and I reproduced both shapes on 15 before changing anything.

The redeclaration:

```
ALTER TABLE public.child ADD COLUMN n integer NOT NULL;
pista: error: ... ERROR: column "n" of relation "child" already exists (SQLSTATE 42701)
```

and the reverse:

```
ALTER TABLE public.child DROP COLUMN n;
pista: error: ... ERROR: cannot drop inherited column "n" (SQLSTATE 42P16)
```

The sentence was wrong because I wrote it from the current side alone. `attislocal` tells a declared column from an inherited one there, but the desired side only ever says "declared", so there is nothing for the diff to compare a flip against, and it emits the column statement.

Fixed in 26be890, documentation only. `docs/reference/objects.md` now says that a redeclaration merges at `CREATE TABLE` only, that adding or removing one on an existing child has no DDL behind it, and names both errors. `TODO.md` records the two shapes next to the inherited-column comment entry, marked `Priority: low` since a dump writes a redeclared column as declared and feeds back clean, and pointing at the same prerequisite that entry names. I left the note that erroring at plan time may fit better than emitting DDL that cannot work, rather than deciding it here.

I did not widen the PR, as you suggested. No fixture either: pinning the plan output would lock in DDL that fails at apply.

`make test`, `make lint` and `test/fidelity` (31 passed) still green.

winebarrel:
LGTM.

Checked 26be890. The change is documentation only, `TODO.md` and `docs/reference/objects.md`, and both now say what I observed: a redeclaration merges at `CREATE TABLE` only, adding or removing one on an existing child plans `ADD COLUMN` / `DROP COLUMN` that PostgreSQL rejects, and the two errors are named. The TODO entry sits next to the inherited-column comment entry and points at the same prerequisite, which is where I would have put it. Agreed on not pinning a fixture for DDL that fails at apply, and the note that a plan-time error may fit better than emitting it is a fair thing to leave open.

Both files are ASCII-only, so the `ascii` job will pass. No behaviour changed since my earlier run, so the test, lint, and fidelity results from that stand.

---

PR #512, opened 2026-09-05
Manage a view's WITH CHECK OPTION

winebarrel:
Neither side read it, so dump dropped the clause and an updatable view
lost the constraint on restore, and a desired schema that wrote it
planned nothing. The catalog reads check_option from pg_class.reloptions,
the parser reads the trailing WITH [LOCAL | CASCADED] CHECK OPTION and
the WITH (check_option = ...) spelling before AS, and dump writes the
clause after the query the way pg_dump does.

A change on a view whose definition stays goes out as ALTER VIEW ... SET
(check_option=...) or RESET (check_option). A definition change carries
the clause on its CREATE OR REPLACE VIEW, which replaces the option along
with the query, so no ALTER follows it.

The fidelity schema that left the clause out now carries it.

### DIFF
--- a/catalog/views.go
+++ b/catalog/views.go
@@ -68,7 +68,7 @@ func (c *Catalog) ListViews(ctx context.Context) ([]*model.View, error) {
 			c.relkind = 'm' AS materialized,
 			(
 				SELECT
-					split_part(o, '=', 2)
+					lower(split_part(o, '=', 2))
 				FROM
 					unnest(c.reloptions) AS o
 				WHERE
--- a/catalog/views.go
+++ b/catalog/views.go
@@ -66,6 +66,14 @@ func (c *Catalog) ListViews(ctx context.Context) ([]*model.View, error) {
 			c.relname,
 			pg_catalog.pg_get_viewdef(c.oid, true) AS definition,
 			c.relkind = 'm' AS materialized,
+			(
+				SELECT
+					split_part(o, '=', 2)
+				FROM
+					unnest(c.reloptions) AS o
+				WHERE
+					o LIKE 'check_option=%'
+			) AS check_option,
 			d.description
 		FROM
 			pg_catalog.pg_class c
@@ -95,16 +103,21 @@ func (c *Catalog) ListViews(ctx context.Context) ([]*model.View, error) {
 	var views []*model.View
 	for rows.Next() {
 		var v model.View
+		var checkOption *string
 		err := rows.Scan(
 			&v.OID,
 			&v.Schema,
 			&v.Name,
 			&v.Definition,
 			&v.Materialized,
+			&checkOption,
 			&v.Comment,
 		)
 		if err != nil {
 			return nil, fmt.Errorf("catalog: failed to scan view info: %w", err)
+		}
+		if checkOption != nil {
+			v.CheckOption = *checkOption
 		}
 		v.Indexes = orderedmap.New[string, *model.Index]()
 		v.Triggers = orderedmap.New[string, *model.Trigger]()
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -1302,12 +1302,41 @@ func parseViewStmt(vs *pg_query.ViewStmt, defaultSchema string) (*model.View, er
 	}
 
 	return &model.View{
-		Schema:     schema,
-		Name:       vs.View.Relname,
-		Definition: def,
-		Indexes:    orderedmap.New[string, *model.Index](),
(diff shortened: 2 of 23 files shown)


catalog/views.go
```diff
@@ -66,6 +66,14 @@ func (c *Catalog) ListViews(ctx context.Context) ([]*model.View, error) {
 			c.relname,
 			pg_catalog.pg_get_viewdef(c.oid, true) AS definition,
 			c.relkind = 'm' AS materialized,
+			(
+				SELECT
+					split_part(o, '=', 2)
```
winebarrel: PostgreSQL stores the `check_option` value as written, and its validator is case-insensitive, so a view created with `WITH (check_option = 'CASCADED')` has `{check_option=CASCADED}` in `reloptions`. The parser lowercases the option spelling, but this read keeps the raw value, so the two sides never match for such a view. I reproduced it on 16: `dump` writes `WITH CASCADED CHECK OPTION`, and feeding that back plans `ALTER VIEW public.q SET (check_option='cascaded');` on every run, which breaks the dump round trip.

`lower(split_part(o, '=', 2))` fixes it (verified locally: the same plan comes back as `No changes`, and `TestViews` still passes). The trailing `WITH ... CHECK OPTION` form is unaffected since `DefineView` always stores it lowercase.

---

parser/parser.go
```diff
@@ -1302,14 +1302,43 @@ func parseViewStmt(vs *pg_query.ViewStmt, defaultSchema string) (*model.View, er
...
+		CheckOption: viewCheckOption(vs),
+		Indexes:     orderedmap.New[string, *model.Index](),
+		Triggers:    orderedmap.New[string, *model.Trigger](),
 	}, nil
 }
 
+// viewCheckOption reads a view's check option from the trailing
+// WITH [LOCAL | CASCADED] CHECK OPTION, which pg_dump and pista dump write, or
+// from a check_option entry in the WITH (...) before AS. A bare WITH CHECK
+// OPTION is CASCADED. The trailing clause wins when both are written, as in
```
winebarrel: Nit: PostgreSQL does not let the trailing clause win here. `DefineView` appends the trailing clause to `options`, so writing both forms fails with `parameter "check_option" specified more than once` (checked on 16). The precedence this function picks is harmless since the server rejects the statement at apply, but the comment and the `both_forms` case in `TestParseSQL_ViewCheckOption` describe behavior PostgreSQL does not have. Consider rewording the comment (or dropping that case).

---

winebarrel (commented):
Reviewed at bb34e47. The design looks right: `ALTER VIEW SET/RESET (check_option)` when only the option changes, and the clause riding on `CREATE OR REPLACE VIEW` when the definition changes, which I confirmed replaces `reloptions` as a whole so no ALTER is needed after it. Rename, recreate, and the `WITH (check_option = ...)` spelling are all covered by fixtures.

Verified locally on PostgreSQL 16: `go vet`, `golangci-lint`, `make test`, and `make test-fidelity` (31 passed) are all clean.

One thing worth fixing before merge: the catalog reads the option value verbatim while the parser lowercases it, so a view created with an uppercase value (`WITH (check_option = 'CASCADED')`) drifts forever and breaks the dump round trip. A `lower()` in the query resolves it. Details and a nit on the parser comment inline.

---

catalog/views.go
winebarrel: Reproduced on 16 and fixed in fe28fdb with `lower(split_part(o, '=', 2))`. `TestViews` now reads a view created with `check_option = 'CASCADED'`, and `testdata/plan/no_diff_view_check_option_case.yml` pins the round trip: it planned the `ALTER VIEW` before the fix and plans nothing after.

---

parser/parser.go
winebarrel: Confirmed: `parameter "check_option" specified more than once` on 16. The comment now says PostgreSQL rejects the statement and the order does not matter, and the `both_forms` case is dropped (fe28fdb).

---

winebarrel (commented):
Re-reviewed at fe28fdb. Both points are addressed: the catalog lowercases the value, `TestViews` covers a view created with `check_option = 'CASCADED'`, and the new `no_diff_view_check_option_case.yml` fixture pins the round trip. The parser comment now matches PostgreSQL and the invalid `both_forms` case is gone.

Verified locally on 16: `go vet`, `golangci-lint`, and `make test` are clean, and the uppercase view that planned an `ALTER VIEW` on every run before now plans `No changes`. CI is green on this head.

LGTM.

---

PR #519, opened 2026-09-06
Manage a view's storage parameters

winebarrel:
Neither side read the WITH (...) that precedes AS, so dump dropped the
clause and a schema file that wrote one was applied without it.

A plain view holds security_barrier and security_invoker and nothing
else. Both say what the view means rather than how it is stored, so they
are managed without a flag, the way the check_option beside them in
reloptions is: a dump that dropped them restored a view that leaks rows
or reads the base table with the wrong privileges. A materialized view
holds what a table holds, tuning included, so it sits behind
--manage-storage-param with the table.

A change to a view whose definition stays goes out as
ALTER [MATERIALIZED] VIEW ... SET (...) / RESET (...). A definition
change carries the clause on its own statement, which replaces the
view's options as a whole, so a parameter left off it is reset and one
that stays has to be written again.


winebarrel:
LGTM.

I read the whole diff (parser, catalog, model, diff, diff_all/dump gating, docs, fixtures) and did not find anything blocking.

What I checked:

- The catalog and the parser both drop `check_option` before building the map, so the check option is never compared or written twice. The `view_check_option_and_storage_params` apply fixture pins that down.
- The `WITH (...)` clause is emitted only in the "definition unchanged" branch as `ALTER ... SET/RESET`, and every CREATE path (new view, `CREATE OR REPLACE VIEW`, DROP+CREATE for a matview, the renamed-and-recreated case) goes through `View.SQL()`, which carries the clause. Since `CREATE OR REPLACE VIEW` replaces reloptions as a whole, the "no separate RESET on a definition change" behavior is correct.
- A materialized view sits behind `--manage-storage-param` on both the plan and dump sides (`clearMatViewStorageParams` on current and desired), while a plain view is managed unconditionally. Every code path tolerates a nil `StorageParams`, so callers that build a `model.View` by hand are unaffected.
- The TOAST join on `reltoastrelid` yields NULL for a plain view and the `toast.` prefix for a matview, matching the table path.
- `go vet` and the DB-free `model`/`parser` tests pass locally; the PG 15/17, samples and fuzz jobs are already green in CI.

One optional nit, not blocking: none of the seeds in `internal/testutil/fuzzseed` carry a view `WITH (...)` clause, so adding one would let the fuzz targets start from the new parser path.

---

winebarrel:
The fuzz seed nit is already covered by this PR. `fuzzseed.Schemas` globs `test/fidelity/schemas/*.sql`, and `view_storage_params.sql`, added here, carries a plain view with `WITH (security_barrier, security_invoker)` and a matview with `WITH (fillfactor, autovacuum_enabled)`, so both targets already start from the new parser path. The hand-written seeds next to the glob exist only for the directive scanner, which pg_query drops before a schema file can reach it.

Pushed one commit for Codecov's two missing lines. Coverage is measured per package in CI, so the two append branches in `diffViewStorageParams` were reported uncovered even though the plan and apply fixtures reach them through the root package. `TestDiffViews_storageParams` and `TestDiffViews_matViewStorageParams` cover them in the diff package itself.

---

winebarrel:
Re-reviewed 3b5fe21. LGTM.

You are right about the fuzz seeds: `fuzzseed.Schemas` globs `test/fidelity/schemas/*.sql`, so `view_storage_params.sql` already puts both the plain view and the matview clause in front of the fuzz targets. My nit was wrong, sorry for the noise.

The new commit only adds `TestDiffViews_storageParams` and `TestDiffViews_matViewStorageParams` in `diff/views_test.go`. They exercise the SET and RESET branches of `diffViewStorageParams` inside the diff package, which is what Codecov's per-package measurement was missing. Both pass locally, and `golangci-lint run` reports 0 issues on this head. Nothing else changed since my first review, so the earlier conclusions stand.

---

PR #521, opened 2026-09-06
Manage a comment on an index

winebarrel:
COMMENT ON INDEX was read by neither side, so dump dropped the line
pg_dump writes and a schema file that carried one was warned about and
ignored.

The catalog joins pg_description on the index, the parser reads the
statement onto the index the tables and views declare, and the diff
emits COMMENT ON INDEX ... IS ... or IS NULL for a change. An index that
is dropped and recreated loses its comment, so the recreate writes it
again, on a materialized view's index too.

The statement names the index without the relation it sits on, so the
plan's topological sort gives every index the position of the relation
it sits on. Without that the comment sorted as an unidentified statement
and ran before the CREATE INDEX it comments on.

The index a PRIMARY KEY, UNIQUE or EXCLUDE constraint owns is the
constraint's and is on neither side, so a comment naming it is dropped
without a warning, as is one naming an index no schema file declares.
Comments on constraints, triggers and policies stay unmanaged; TODO.md
records what is left and the shape the work takes.

Verified with the plan, dump and apply fixtures, package tests for the
catalog read, the parser, the diff and the SQL rendering, and two
fidelity schemas that pg_dump compares. Coverage is unchanged per
package, parser up 0.1 point. make test, test-fidelity, test-scenario,
test-samples and lint all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Manage a comment on an index. `COMMENT ON INDEX` was read by neither side, so `dump` dropped the line `pg_dump` writes and a schema file that carried one was warned about and ignored. A change now goes out as `COMMENT ON INDEX ... IS ...` or `IS NULL`, and a recreate writes the comment again, since PostgreSQL drops it with the index. A comment on a constraint, a trigger or a policy, and one on the index a constraint owns, stays unmanaged.
+
 ## [1.45.0] - 2026-09-06
 
 * Manage the storage parameters of a view and a materialized view, the `WITH (...)` clause before `AS`. Neither side read them, so `dump` dropped the clause and a schema file that wrote one was applied without it. A plain view holds `security_barrier` and `security_invoker` and nothing else. Both decide what the view means rather than how it is stored, so they are managed without a flag: a dump that dropped them restored a view that leaks rows, or one that reads the base table with the wrong privileges. A materialized view holds what a table holds, tuning included, so it sits behind `--manage-storage-param` with the table. A change to a view whose definition stays goes out as `ALTER [MATERIALIZED] VIEW ... SET (...)` / `RESET (...)`; a definition change carries the clause on its own statement, which replaces the view's options as a whole. Two things that used to plan nothing now plan something: a view carrying `security_barrier` the schema file does not name is reset, and, under the flag, a materialized view's parameters are diffed like a table's.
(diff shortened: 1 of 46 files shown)


winebarrel (commented):
Reviewed the full diff and ran it locally against PostgreSQL 16 (`make lint` with golangci-lint 2.13.2, deadcode, `make test`, `make test-fidelity`, `make test-scenario`).

The implementation looks right to me:

- The `pg_description` join in `ListIndexes` is scoped by `classoid` and `objsubid = 0`, and the constraint-owned index stays excluded, so a comment on a PK/UNIQUE index cannot leak into the model.
- `findIndex` checks `idx.Schema == schema` while scanning, so a same-named index in another schema does not catch the comment (covered by `TestParseSQL_CommentOnIndexSameNameOtherSchema`).
- Re-emitting the comment on the recreate path in both `diffIndexes` and the matview create/recreate paths is the correct call, since `DROP INDEX` takes the comment with it.
- `addIndexPositions` plus `extractFirstIdentifier` preserving quoting means `COMMENT ON INDEX` sorts with its table even for quoted index names, and it lands after `CREATE INDEX [CONCURRENTLY]` because both sit at the same position and the sort is stable.
- Dump writes the comment after `IdxSQL()`, so the round trip is clean. Docs, CHANGELOG and the TODO narrowing all match the behavior.

One thing, which I understand you are already fixing:

- `testdata/dump/matview_index_comment.yml` expects `SELECT users.id, users.name FROM users`, which is the PG15 form of `pg_get_viewdef`. PG16+ emits `SELECT id, name FROM users`, which is why `test (16/17/18)` and `windows` are red while `test (15)` is green. Adding `dump_pg16`/`dump_pg17`/`dump_pg18` with a YAML anchor, as `enable_view.yml` does, makes it pass on 16 locally.

Optional nit, no action needed:

- `addIndexPositions(dropPosMap, ...)` has no effect today. Nothing tagged with `dropPosMap` resolves to a `schema.index` key (`DROP INDEX` returns `""` from `extractObjectName`, and the comment statements are all in `tableDiff.Stmts` / `viewDiff.CreateStmts`). It is harmless and keeps the two maps symmetric, so fine to leave.

LGTM once the fixture is version-aware.

---

winebarrel (commented):
Re-reviewed at 13caec5. The only change since the last round is `testdata/dump/matview_index_comment.yml`, which now carries `dump_pg16`/`dump_pg17`/`dump_pg18` through a YAML anchor in the same shape as `enable_view.yml`. That was the one blocker, and it is resolved: `TestDump/matview_index_comment` passes on PostgreSQL 16 locally, and every CI job is green on this head (test 15/16/17/18, windows, samples, fuzz, CodeQL, codecov).

No further comments. LGTM.

---

PR #522, opened 2026-09-06
Change a foreign key's deferral clause in place

winebarrel:
DEFERRABLE and INITIALLY DEFERRED were compared only as part of the
definition string pg_get_constraintdef returns, so a toggle dropped the key
and added it back. Adding a foreign key back rescans the table to validate
it, under a lock, for a change that rewrites one catalog row.

compareFKDef now answers both questions the diff asks of a pair of
definitions: whether they are equal, and whether the deferral clause is all
that separates them. The second routes the change to
ALTER TABLE ... ALTER CONSTRAINT. A key that is also being validated takes
VALIDATE CONSTRAINT next to it; a validated key the desired schema writes
NOT VALID is still added back, since nothing takes the flag away; and a
change to anything else about the key is a drop and an add as before, with
the deferral clause riding on the ADD.

PostgreSQL accepts ALTER CONSTRAINT on a foreign key alone, verified on 15
and 18, so a unique, primary key or exclusion constraint keeps the drop and
the add. A partition holds a copy of every key its parent declares and
rejects a statement on the copy; the catalog now marks such a copy through
conislocal and the diff leaves it to the parent's ALTER, which recurses to
the partitions.

While confirming that last case I found an existing bug next to it: a
foreign key's *definition* change on a partitioned table plans a drop for
the copy as well, and the apply dies with `constraint "fk_user" of relation
"orders_2025" does not exist`. It is out of scope here and TODO.md records
it, with the note that the Inherited flag this PR adds is what closes it.

Eleven plan fixtures, two apply fixtures, diff and catalog package tests.
make test passes on 15, 16, 17 and 18; test-fidelity, test-scenario,
test-samples and lint pass on 15. Coverage does not drop: catalog 93.6 ->
93.7, diff 97.6 -> 97.7, the rest unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+* Change a foreign key's deferral clause in place. `DEFERRABLE` and `INITIALLY DEFERRED` were compared only as part of the definition string, so a toggle dropped the key and added it back, rescanning the table to validate it again. When nothing else about the key differs it now goes out as `ALTER TABLE ... ALTER CONSTRAINT`, which rewrites a catalog row, and a key that is also being validated takes `VALIDATE CONSTRAINT` next to it. PostgreSQL accepts the statement on a foreign key alone, so a unique, primary key or exclusion constraint is still dropped and added. A partition's copy of its parent's key takes no statement, since PostgreSQL rejects one and the parent's reaches the copy.
+
 * Manage a comment on an index. `COMMENT ON INDEX` was read by neither side, so `dump` dropped the line `pg_dump` writes and a schema file that carried one was warned about and ignored. A change now goes out as `COMMENT ON INDEX ... IS ...` or `IS NULL`, and a recreate writes the comment again, since PostgreSQL drops it with the index. A comment on a constraint, a trigger or a policy, and one on the index a constraint owns, stays unmanaged.
 
 ## [1.45.0] - 2026-09-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-* Change a foreign key's deferral clause in place. `DEFERRABLE` and `INITIALLY DEFERRED` were compared only as part of the definition string, so a toggle dropped the key and added it back, rescanning the table to validate it again. When nothing else about the key differs it now goes out as `ALTER TABLE ... ALTER CONSTRAINT`, which rewrites a catalog row, and a key that is also being validated takes `VALIDATE CONSTRAINT` next to it. PostgreSQL accepts the statement on a foreign key alone, so a unique, primary key or exclusion constraint is still dropped and added. A partition's copy of its parent's key takes no statement, since PostgreSQL rejects one and the parent's reaches the copy.
+* Change a foreign key's deferral clause in place. `DEFERRABLE` and `INITIALLY DEFERRED` were compared only as part of the definition string, so a toggle dropped the key and added it back, rescanning the table to validate it again. When nothing else about the key differs it now goes out as `ALTER TABLE ... ALTER CONSTRAINT`, with `VALIDATE CONSTRAINT` next to it when the key is also being validated. PostgreSQL takes the statement on a foreign key alone, so a unique, primary key or exclusion constraint is still dropped and added, and a partition's copy of its parent's key takes no statement at all.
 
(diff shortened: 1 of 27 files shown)


winebarrel (commented):
Reviewed the full diff at 0f5d4e4 and ran it locally against PostgreSQL 16: golangci-lint, deadcode, `make test`, `make test-fidelity` (32 passed), `make test-scenario` (11 passed). CI is green on 15/16/17/18 and windows.

I also checked the PostgreSQL premises the change rests on, on 16 directly:

- `ALTER CONSTRAINT` on the partition's copy is rejected (`cannot alter constraint "fk_user" on relation "orders_2025"`, derived from the parent's), and the parent's statement recurses to the copy.
- `ALTER CONSTRAINT` on a unique constraint is rejected (`is not a foreign key constraint`), so keeping the drop and the add there is right.
- `NOT DEFERRABLE` clears both `condeferrable` and `condeferred`, so `alterFKDeferralSQL` rendering it that way is correct.

The code reads well. `compareFKDef` answering both questions from one pair of parses keeps the cost where it was, the `Inherited` flag is read from `conislocal` after the existing children-of-partition skip so it only lands on a leaf's copy, and the drop/add/rename loops treat `deferralOnly` exactly like `validateOnly` where the key stays. The fixtures cover the branches I could think of: added/removed/immediate, validate alongside, becomes NOT VALID, definition change, rename, quoted name, drop policy, bulk alter, assume-validated, partition copy vs. partition-own key, and unique still recreating.

One non-blocking question:

- In `foreignKeyChanges`, the partition-copy case is `currentFk.Inherited && sameValidation`. When the copy's validation also differs, it falls through to the default branch and plans `ALTER CONSTRAINT` on the copy, which PostgreSQL rejects, before the `VALIDATE`. Today that is unreachable on 15-17, since a `NOT VALID` foreign key on a partitioned table is refused there (`cannot add NOT VALID foreign key on partitioned table`, verified on 16), so it can only show up on 18 where that became possible. If the intent from the docs is "the copy takes no statement of its own", dropping the `sameValidation` guard would match it, since the parent's `ALTER` and `VALIDATE` both recurse. Fine to leave as a follow-up given the TODO already covers the copy's drop path.

LGTM.

---

winebarrel:
Took the non-blocking point rather than deferring it, since the docs already said the copy takes no statement of its own.

Confirmed the premises on 15 and 18 first:

- 15 refuses `ALTER TABLE ... ADD CONSTRAINT ... FOREIGN KEY ... NOT VALID` on a partitioned table (`cannot add NOT VALID foreign key on partitioned table`), so the case is unreachable there. 18 accepts it and the copy comes out `convalidated = f`, `conislocal = f`.
- On 18 the copy still rejects `ALTER CONSTRAINT` (`cannot alter constraint "fk" on relation "orders_2025"`).
- On 18 the parent's `ALTER CONSTRAINT` and `VALIDATE CONSTRAINT` both reach the copy: after the two, the copy reads `convalidated = t`, `condeferred = t`.

So the guard is now `case currentFk.Inherited:` alone. The copy takes no statement whatever changed, and the parent's two settle it.

Covered by `testdata/plan/fk_deferrable_partition_not_valid.yml` and `testdata/apply/fk_deferrable_partition_not_valid.yml`, both `min_pg: 18`, plus a diff-package subtest. The apply fixture re-plans afterwards and requires no drift, which is what proves the parent's statements reach the copy. Docs say which two statements reach it.

6965cc5. `make test` passes on 15, 16, 17 and 18; lint, deadcode and vet pass.

winebarrel (commented):
Re-reviewed at 6965cc5. The change since the last round is the one I asked about: the partition-copy case in `foreignKeyChanges` is now `currentFk.Inherited` alone, so a copy whose deferral clause and validated flag both differ takes no statement, and the docs say the same. `validateOnly` in the default branch is spelled out as `!currentFk.Validated && desiredFk.Validated`, which is equivalent to the old `!sameValidation` after the validated-to-NOT-VALID case above it, and reads clearer.

The two new `min_pg: 18` fixtures cover exactly that case: the parent's `ALTER CONSTRAINT` and `VALIDATE CONSTRAINT` are planned, the copy gets nothing, and the apply fixture shows the second plan is clean. They skip on 15-17 as intended, and `test (18)` is green on this head along with every other job.

Locally on 16: lint 0 issues, `go test ./diff/` passes, and all `fk_deferrable*` / `unique_deferrable*` plan and apply fixtures pass.

LGTM.

---

PR #523, opened 2026-09-06
Manage a foreign key on a partitioned table

winebarrel:
The TODO entry #522 opened was one of three things standing between a
foreign key on a partitioned table and an apply that works. All three are
verified on 15 and 18.

**ONLY is rejected.** `ALTER TABLE ONLY ... ADD CONSTRAINT ... FOREIGN KEY` on
a partitioned table fails with `cannot use ONLY for foreign key on partitioned
table` (SQLSTATE 42809), so adding a key to one has never applied, whatever
else was in the plan. The word is left off there now. A foreign key is not
inherited by an INHERITS child either, checked directly, so ONLY decides
nothing anywhere else; it stays on a plain table because that is what pg_dump
writes.

**The partition's copy took statements of its own.** A definition change
planned a drop and an add for the parent and for each copy. The parent's DROP
takes the copy with it, and the copy's own drop then died with
`constraint "fk_user" of relation "orders_2025" does not exist`. PostgreSQL
rejects a DROP on a copy in the first place. The copy now takes no statement:
the parent's DROP takes it, the parent's ADD puts a new one back, and its
ALTER CONSTRAINT and VALIDATE CONSTRAINT recurse. A key the partition declares
itself has `conislocal = true` and is untouched by this.

**dump wrote the copy.** Reloading that output died with
`constraint "fk_user" for relation "orders_2025" already exists`, since the
parent's statement had already created it. `dump` leaves the copy out now, as
pg_dump does.

test/fidelity/schemas/partition_fk.sql is the regression test for the last
one: it fails on the old code with `FAIL (reload)` and passes here. Plan,
apply and dump fixtures cover the other two, plus diff and model package
tests.

make test passes on 15, 16, 17 and 18. On 15: test-fidelity (33), test-scenario
(11), test-samples (54), lint, deadcode, vet; test-fidelity also on 18.
Coverage: diff 97.6 -> 97.7, catalog 93.7, the rest unchanged.


winebarrel (commented):
Reviewed the full diff at 447ef70 and ran it locally against PostgreSQL 16: vet, golangci-lint, deadcode, `make test`, `make test-fidelity` (33 passed, `partition_fk` included), `make test-scenario` (11 passed). CI is green on 15/16/17/18 and windows.

I checked each premise on 16 directly, and all three hold:

- `ALTER TABLE ONLY` on a partitioned table fails with `cannot use ONLY for foreign key on partitioned table`; without `ONLY` it succeeds and the partition gets a `conislocal = false` copy.
- Adding the copy again on the partition fails with `already exists`, which is the reload failure `dump` used to produce. `DROP CONSTRAINT` on the copy fails with `cannot drop inherited constraint`.
- An `INHERITS` child gets no copy of a foreign key, so leaving `ONLY` on a plain table decides nothing.

The code follows from that. Testing `Inherited` at the top of `foreignKeyChanges` and again in the drop loop is the right pair, since a key the desired schema drops never reaches `changes`, and the drop-loop skip sits before the policy check so the copy is not reported as a skipped drop either. `FkSQL` leaving the copy out while `newTableExtras` still writes every desired key is consistent, since only the catalog sets `Inherited`. The nested fixture covers the mid-level partition taking no `ONLY`, and the `diff` package tests pin the definition-change, dropped and drops-denied shapes.

One observation, not a blocker:

- Renaming the parent's key does not recurse. On 16, `ALTER TABLE p RENAME CONSTRAINT p_fk TO p_fk2` leaves the partition's copy named `p_fk`, while a partition attached afterwards gets a copy named `p_fk2`. The plan stays clean either way, since the copy is neither dumped nor diffed, so after a `pista:renamed-from` on a partitioned table the existing copies just keep the old name in the catalog. The comment in `foreignKeyChanges` about a rename being the one statement a copy still takes matches what PostgreSQL allows: renaming the copy itself is accepted. Worth a line in the docs at most.

LGTM.

---

winebarrel:
Checked the rename observation on 15 and it reproduces exactly as described: after `ALTER TABLE dtest.orders RENAME CONSTRAINT p_fk TO p_fk2`, the existing partition still reads `p_fk` while one attached afterwards reads `p_fk2`.

Added the line to the docs rather than leaving it out, since the divergence is visible in psql and nowhere in pistachio's output. No code change: the copy is neither written nor compared, so the plan is clean either way.

winebarrel (commented):
Re-reviewed at 7939eb5. The only change since the last round is the one sentence in `docs/reference/objects.md` on renames, and it states exactly what I saw on 16: the copy keeps its old name, a partition attached later takes the new one, and nothing drifts because the copy is neither written nor compared. No code changed, so the earlier verification stands, and the `ascii` job is already green on this head.

LGTM.

---

PR #533, opened 2026-09-07
Add pista fmt

winebarrel:
Lay out schema SQL files without changing the statements they hold.

`pista fmt` works on the token stream, so the keyword case, the identifier spelling and the line breaks come out as they were written. Only the whitespace between the tokens moves.

```sql
create table "public"."items"(id bigserial,"email" text not null,tags text [ ],price numeric ( 10 , 2 ));
```

```sql
create table public.items(
    id bigserial,
    email text not null,
    tags text[],
    price numeric (10, 2)
);
```

## Rules

- The definition list of `CREATE TABLE` and `CREATE TYPE` is written one element per line, with the comma at the end of the line, the closing parenthesis on a line of its own, and any clause after it on the next line.
- A line inside parentheses is indented four spaces past the line that opened them, and the closing parenthesis lines up with that line again.
- The clauses of `CREATE FUNCTION` and `CREATE PROCEDURE` are indented one level.
- Consecutive spaces collapse to one. The space before a comma or a semicolon, just inside a parenthesis, around an array subscript, and around a cast is closed up.
- Consecutive blank lines collapse to one, trailing whitespace goes, and a statement that shares a line with another moves to its own line.
- A quoted identifier that reads the same without its quotes loses them.

Everything else stays where it was written: a statement written on one line stays on one line, and one broken across several keeps its breaks. The body of a view is left alone, since PostgreSQL writes it back with an indentation of its own, and so is the body of a routine, which is a single quoted token.

## Guarantees

A formatted file carries the same tokens as the input. `Format` checks this before returning, and the command leaves a file alone when the check fails, so a bug in the layout cannot reach a schema file. One such bug was caught this way while writing the tests: a comma moved onto a line that ended in a comment was swallowed by it.

An identifier loses its quotes only when the statement, parsed again without them, deparses to what it deparsed to before. That settles the position for us: `"name"` is written bare as a column name and stays quoted as a function name, and `"select"` and `"left"` keep their quotes everywhere.

## dump

`dump` writes through the same formatter, so the two agree on the layout. No dump output changes today, which every dump fixture now asserts; `--no-format` turns the pass off.

## Tests

`make test`, `make test-scenario` (11), `make test-fidelity` (33), `make lint` and `make vet` pass. Coverage: `format` 95.5%, root package 98.3% (unchanged), `cmd/command` 95.2% (96.5% before; the difference is four error returns in the atomic write that cannot be forced without faking the filesystem).

A fuzz target holds the token invariant and idempotency; it ran 5.8M executions with no failure.


winebarrel (commented):
I read the whole diff, ran the `format` and `cmd/command` tests locally, and probed the formatter with a set of inputs beyond the fixtures (an `INHERITS` child with an empty column list, a `-- pista:` directive on an element line, leading commas, a blank line inside a definition list, two statements on one line, a routine with `RETURNS TABLE (...)`, a `CHECK` whose `IN (...)` list spans lines, a comment between the closing parenthesis and the semicolon). Every one came out sane and idempotent, and the empty-list and `INHERITS` output matches what `TableToSQL` writes.

The design holds up well: working on the token stream plus the `verify` pass means a layout bug cannot corrupt a file, and settling the unquote decision by re-parsing and comparing deparse output is a neat way to avoid modeling the grammar. The dump fixtures asserting `Format` is a no-op is the right guard for the dump/fmt agreement.

One thing I would fix before merging:

**`tokenRange` is quadratic in the input size.** It scans the token slice from index 0 for every statement, and it is called from `listRegions`, `unquote` and `stmtStarts` for each statement, so `Format` is O(statements x tokens). On a generated schema of one `CREATE TABLE` (6 defs) plus one `CREATE INDEX` per table:

| tables | statements | Format |
|---|---|---|
| 500 | 1,000 | 0.18s |
| 1,000 | 2,000 | 0.74s |
| 2,000 | 4,000 | 3.3s |
| 4,000 | 8,000 | 13.5s |

A CPU profile puts 88% of the time in `tokenRange`. Since `dump` now runs the whole output through `Format` in `String()`, a `pista dump` of a few thousand tables pays this too. Tokens are sorted by `start`, so a binary search fixes it without changing behavior:

```go
func tokenRange(toks []*token, s stmt) (int, int) {
	lo := sort.Search(len(toks), func(i int) bool { return toks[i].start >= s.start })
	hi := sort.Search(len(toks), func(i int) bool { return toks[i].start >= s.end })

	return lo, hi
}
```

With that change the same runs take 0.07s / 0.11s / 0.26s / 0.50s, and the `format` tests still pass.

Everything else reads fine to me. Two minor notes, no action needed:

- `writeFileAtomic` replaces a symlinked schema file with a regular file at the link's path. That is the usual atomic-write trade-off and not worth handling here.
- A whitespace-only file returns unchanged rather than being normalized, since `scan` yields no tokens. Harmless.

---

winebarrel:
Thanks. Fixed the `tokenRange` scan in be2c79e, with the binary search you suggested.

I reproduced the growth here before changing anything, on the same shape of schema (one `CREATE TABLE` of six definitions plus one `CREATE INDEX` per table):

| statements | before | after |
|---|---|---|
| 1,000 | 69ms | 23ms |
| 2,000 | 137ms | 38ms |
| 4,000 | 514ms | 76ms |
| 8,000 | 1.8s | 0.15s |

The absolute numbers are lower than yours but the curve is the same, and it is linear now.

On the two notes: I left both alone. Replacing a symlink with a regular file is what an atomic write does, and a whitespace-only file has no tokens to lay out, so returning it unchanged is the same answer the rules would give.

`make test`, `make lint`, `make test-scenario` and `make test-fidelity` pass. The Windows failures were the two permission tests I had added; they are skipped there as of 5c69876.

winebarrel (commented):
Re-reviewed the two new commits.

The `tokenRange` fix is what I suggested and it works: on the same generated schemas, 4,000 tables (8,000 statements) went from 13.5s to about 1s on my machine, and the `format` and `cmd/command` tests pass on the new head. The performance page update reads well, and it is good to see `fmt` and `dump --no-format` measured against the same schemas.

One thing to remove before merging:

**`format/zz_prof_test.go` is a leftover profiling scratch file.** It came in with the "Re-measure the performance page" commit and reads a schema from an absolute path under the author's local scratchpad, so it skips everywhere else. It does not break CI, but it is not part of the change and should not be in the repo.

Once that file is gone, LGTM.

---

winebarrel:
Removed in 9186f24. That was the one file; the commit before it only touched `docs/about/performance.md`.

Thanks for re-running the timings on your side.

winebarrel (commented):
The only change since the last round is the removal of `format/zz_prof_test.go`, and nothing else of that kind is left in the tree. `go vet` and the `format` and `cmd/command` tests pass on the new head.

LGTM.

---

PR #543, opened 2026-09-08
Test the pista entrypoint and drop it from the codecov ignore list

winebarrel:
main only read os.Args and exited through os.Exit, so cmd/pista could not
be tested in-process and codecov.yml left it out. Move the body into a run
function that takes the arguments, the output streams and the exit
function, and test the parsing, the pager hookup and the exit codes from
there. No change in behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Test the `pista` entrypoint. `main` now hands the arguments, the output streams and the exit function to a `run` function, so the argument parsing, the pager hookup and the exit codes are covered by `go test`, and `cmd/pista` is no longer left out of the coverage report. No change in behavior.
+
 ## [1.47.0] - 2026-09-07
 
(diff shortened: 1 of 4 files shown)


winebarrel:
LGTM.

What I checked:

- `run` keeps the same behavior as the old `kong.Parse` path: `kong.New` failure still panics, a parse error goes through `parser.FatalIfErrorf` and exits with kong's usage code (80), a run error exits with 1 via `Fatalf`, and `--check` diffs exit with 2. Stdout/stderr and the `io.Writer` binding are now injected instead of hard-coded, with the pager rebinding unchanged.
- The explicit `closePager()` before `kctx.Exit(2)` / `FatalIfErrorf` is preserved, so the pager still drains before the process exits, and the deferred close covers the panicking `exit` the tests pass in.
- Ran `go test -cover ./cmd/pista/` against a local PostgreSQL 16: all 8 tests pass, 94.4% coverage (only `main` itself and the `kong.New` panic are uncovered). `cmd/command` tests still pass.
- `TestRun_Pager` relies on `cat`, which the existing `cmd/command` pager tests already use on the Windows job, so dropping `cmd/pista` from the codecov ignore list should not cause a platform difference.

---

PR #544, opened 2026-09-08
Fix four round-trip breaks with one-line causes

winebarrel:
Four unrelated bugs that each break the `apply` then `plan` round trip. Every one has a one-line cause, so they are grouped here; the follow-up PR carries the ones that need a few lines each.

Every case below was reproduced against PostgreSQL 16 before the fix.

## `CREATE INDEX IF NOT EXISTS` is dropped and created on every plan

`parseIndexStmt` clears `Concurrent` before deparsing but not `IfNotExists`, so the clause stayed in the stored definition. `pg_get_indexdef` never writes it, so the two sides never compared equal.

```
DROP INDEX public.users_name_idx;
CREATE INDEX IF NOT EXISTS users_name_idx ON public.users USING btree (name);
```

The clause says how to run the statement, not what the index is, so it is dropped rather than carried into the emitted `CREATE INDEX`.

## An enum with no labels is invisible to the catalog

`CREATE TYPE mood AS ENUM ()` has no `pg_enum` row, and the labels were reached through an inner join, so the type read as absent. Every plan created it again, `dump` left it out, and the second apply failed with `type "mood" already exists`.

The join is now outer, and rendering an empty enum writes `AS ENUM (\n);`, which is what `pg_dump` writes for the same type. PostgreSQL accepts the form: `CREATE TYPE` takes an optional label list, the result is a usable type, and labels can be added later with `ALTER TYPE ... ADD VALUE`.

## A routine's `SET timezone` is replaced on every plan

GUC names are case-insensitive and `pg_get_functiondef` writes the spelling PostgreSQL keeps in `proconfig`, so a routine created with `SET timezone` reads back as `SET "TimeZone"` and never matched what the file wrote. `DateStyle` and `IntervalStyle` behaved the same way; `work_mem` and `search_path` did not, which is why it went unnoticed. Both sides parse through `parseRoutineConfig`, so folding the name there is symmetric.

The emitted `CREATE FUNCTION` writes the name in lower case, which is valid and need not match what `pg_dump` writes.

## Widening a `serial` column fails at apply

The change went out as `ALTER TABLE ... SET DATA TYPE bigserial`, which PostgreSQL rejects with `type "bigserial" does not exist`: a serial pseudo-type is accepted only in `CREATE TABLE` and `ADD COLUMN`.

Naming the base type alone is not the whole change. `bigserial` is a `bigint` column and a `bigint` sequence, and `ALTER TABLE` reaches only the column, so the sequence stayed as `serial` created it and kept handing out numbers up to 2147483647. By hand the change is two statements, and both go out now:

```
ALTER SEQUENCE public.users_id_seq AS bigint;
ALTER TABLE public.users ALTER COLUMN id SET DATA TYPE bigint;
```

The catalog reads the name of the sequence a serial column owns, which it already had at hand for the `is_serial` test. PostgreSQL moves the sequence bounds with the type when they are the old type's defaults, so nothing else is needed. The `ALTER SEQUENCE` sorts first because the table's default draws from the sequence; that order applies cleanly and leaves the column, the sequence and the next value all as `bigserial` would have created them.

A column that owns no sequence, and one being retyped to something a sequence cannot hold, take the single statement they always did.

## Tests

Ten fixtures and five unit tests. Each fixture was checked to fail with its fix reverted: removing the `ALTER SEQUENCE` fails the widening fixture, removing only the type guard emits `ALTER SEQUENCE ... AS text` and fails the `serial` to `text` fixture, and restoring the inner join fails the fidelity check.

- `plan/no_diff_index_if_not_exists.yml`, `apply/create_index_if_not_exists.yml`
- `plan/no_diff_empty_enum.yml`, `dump/empty_enum.yml`, `apply/empty_enum.yml`, `plan/add_value_to_empty_enum.yml`
- `plan/no_diff_routine_set_timezone.yml`
- `plan/alter_column_serial_to_bigserial.yml`, `apply/alter_column_serial_to_bigserial.yml`, `plan/alter_column_serial_to_text.yml`
- `test/fidelity/schemas/types.sql` carries an empty enum, so the restore check compares it against `pg_dump`
- unit tests for `alterTypeName`, `alterSerialSequenceSQL`, `alterColumnSQL` and `Enum.SQL`, since CI measures coverage per package and the root fixtures credit nothing to `diff` and `model`

Per-package coverage is 96.2%, the same as `main`. The lines still uncovered in the touched files are pre-existing defensive paths that valid SQL does not reach.

## Also

`LIMITATIONS.md` gains a bug this work turned up rather than introduced. A column is read as a serial when it owns a sequence and its default draws from it, without regard to type, and the default is then dropped from the dump. A serial column retyped to `text` loses its default, and a hand-attached sequence on a `numeric` column loses both the default and the sequence.

`make vet`, `make test`, `make test-scenario` and `make test-fidelity` pass against PostgreSQL 16. `make lint` could not run here: the installed golangci-lint is built with Go 1.25 and refuses a module targeting 1.27.

https://claude.ai/code/session_018iZFWSeiGCSNGi6zT4R81Q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 * Fold a routine's `SET` name to lower case. GUC names are case-insensitive and `pg_get_functiondef` writes the spelling PostgreSQL keeps, so a routine created with `SET timezone` read back as `SET "TimeZone"` and was replaced on every plan. `DateStyle` and `IntervalStyle` behaved the same way.
 
-* Widen a `serial` column to `bigserial`. The change went out as `SET DATA TYPE bigserial`, which fails with `type bigserial does not exist`, since PostgreSQL takes a serial pseudo-type only in `CREATE TABLE` and `ADD COLUMN`. The base type is named instead, and the sequence the column owns follows it with `ALTER SEQUENCE ... AS`: `bigserial` is a `bigint` column and a `bigint` sequence, so widening the column alone left the numbers stopping at 2147483647. A column that owns no sequence takes the one statement it always did.
+* Widen a `serial` column to `bigserial`. The change went out as `SET DATA TYPE bigserial`, which fails: PostgreSQL takes a serial pseudo-type only in `CREATE TABLE` and `ADD COLUMN`. The base type goes out instead, with `ALTER SEQUENCE ... AS` for the sequence the column owns. `bigserial` is a `bigint` column and a `bigint` sequence, and widening the column alone left the numbers stopping at 2147483647. A column that owns no sequence takes the one statement it always did.
 
 * Test the `pista` entrypoint. `main` now hands the arguments, the output streams and the exit function to a `run` function, so the argument parsing, the pager hookup and the exit codes are covered by `go test`, and `cmd/pista` is no longer left out of the coverage report. No change in behavior.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+* Read `CREATE INDEX IF NOT EXISTS` as the index it describes. The clause was kept in the stored definition, which `pg_get_indexdef` never writes, so an index a schema file spelled that way was dropped and created again on every plan.
+
+* Read an enum that has no labels. `CREATE TYPE ... AS ENUM ()` has no `pg_enum` row, and the catalog reached the labels through an inner join, so the type read as absent: every plan created it again and the second apply failed with `type already exists`. `dump` writes it now as well.
+
+* Fold a routine's `SET` name to lower case. GUC names are case-insensitive and `pg_get_functiondef` writes the spelling PostgreSQL keeps, so a routine created with `SET timezone` read back as `SET "TimeZone"` and was replaced on every plan. `DateStyle` and `IntervalStyle` behaved the same way.
+
+* Widen a `serial` column to `bigserial`. The change went out as `SET DATA TYPE bigserial`, which fails with `type bigserial does not exist`, since PostgreSQL takes a serial pseudo-type only in `CREATE TABLE` and `ADD COLUMN`. The base type is named instead; the column keeps the sequence it owns, and widening that sequence stays a separate change.
+
 * Test the `pista` entrypoint. `main` now hands the arguments, the output streams and the exit function to a `run` function, so the argument parsing, the pager hookup and the exit codes are covered by `go test`, and `cmd/pista` is no longer left out of the coverage report. No change in behavior.
 
 ## [1.47.0] - 2026-09-07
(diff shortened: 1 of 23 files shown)


winebarrel (commented):
LGTM.

I went through each of the four fixes against the code on the branch and did not find anything blocking.

- `IF NOT EXISTS` on `CREATE INDEX`: clearing the flag next to `Concurrent` before deparsing is the right place, since the stored definition is what gets compared against `pg_get_indexdef`.
- Empty enum: `LEFT JOIN` plus `array_remove(..., NULL)` gives `{}` for a type with no `pg_enum` row, and `Enum.SQL` renders it the way `pg_dump` does, so the fidelity check covers it.
- Routine `SET` names: folding in `parseRoutineConfig` is symmetric because both the file and `pg_get_functiondef` output go through it, and GUC names are case-insensitive so the lower-cased `CREATE FUNCTION` is valid.
- Serial widening: `pg_get_serial_sequence` returns a schema-qualified, properly quoted name, so it is safe to splice into `ALTER SEQUENCE`. The `sequenceTypes` guard keeps `serial -> text` on the single statement. `ALTER SEQUENCE ... AS` moves the default bounds with the type, so no extra `MAXVALUE` is needed. Adding `SerialSequence` to `model.Column` does not affect column comparison, since nothing compares the struct as a whole.

One non-blocking nit: `alterColumnSQL` appends the `ALTER SEQUENCE` right after the `SET DATA TYPE`, and `mergeAlterTable` runs on the per-table statement list before the dependency sort. Because the merge only joins consecutive `ALTER TABLE` statements, a `--bulk-alter` table that widens a serial column and changes another column in the same plan ends up with two `ALTER TABLE` statements instead of one. The output is still correct. If you want a single statement, appending the sequence statements after the column loop in `diffColumns` (the same way storage params are kept out of the group) would do it, and the sort still moves them ahead of the `ALTER TABLE`. Fine to leave for a follow-up.

CI is green on PG 15 and 16 at the time of writing; 17 and 18 were still running.

---

winebarrel (commented):
Re-reviewed 3374d4a. LGTM.

The `ALTER SEQUENCE` statements are now collected in `diffColumns` and appended after the column loop, so a `--bulk-alter` run of `ALTER TABLE` stays in one statement. The new `bulk_alter_serial_widening.yml` fixture asserts exactly that shape, with the sequence statement sorted ahead of the merged `ALTER TABLE`.

The `equalTypeName` guard added to `alterSerialSequenceSQL` is the right addition: the call no longer sits under the `retyped` branch of `alterColumnSQL`, so without it every serial column would emit the statement on every plan. The `type unchanged` unit case covers that.

`go vet` and the `diff` package tests pass locally on this head. CI for the new commit was still running when I looked.

---

PR #545, opened 2026-09-08
Fix seven round-trip breaks that need a few lines each

winebarrel:
Six bugs that each break the `apply` then `plan` round trip. #544 carried the ones with a one-line cause; these need a few lines each. Every case was reproduced against PostgreSQL 16 before the fix.

Four of them reach only a hand-written schema, since `pista dump` writes the form that avoids them. The partition drop and the inherited collation reach a dump too.

The collation change is marked BREAKING: a file an earlier `pista dump` wrote no longer plans until it is dumped again. See below.

## A primary key added by `ALTER TABLE` does not imply `NOT NULL`

PostgreSQL sets the flag whichever way the key arrives, and only the inline form was read that way. A file that adds the key in a later statement plans `ALTER COLUMN id DROP NOT NULL` on every run, and applying it fails with `column "id" is in a primary key`.

The inline path's own code moves into a helper both paths call, so nothing new is inferred: it is the rule that path already had, applied to the other spelling.

`pista dump` writes the key inline with `NOT NULL` on the column, and `pg_dump` writes `NOT NULL` too, so neither output reaches this. `PRIMARY KEY USING INDEX` still does; that form has a larger gap, recorded in LIMITATIONS.

## A `serial` column declared without `NOT NULL` is read as nullable

The type expands to a `NOT NULL` column and the catalog reports one, so the plan right after the table was created proposes `DROP NOT NULL`. Applying it takes the constraint away, and the next plan is clean, so nothing says it happened. A `serial` column that also carries a primary key is reached through the key and was never affected.

Of the 757 serial declarations in the sample schemas, 381 write `NOT NULL`, 54 carry a primary key instead, and 322 write neither.

## A `numeric` written with only a precision drifts

PostgreSQL stores `numeric(5)` as `numeric(5,0)` and `format_type` prints both digits, so the two sides never compare equal and the column is retyped on every plan.

The array marker is separated from the modifier so an array of the type is reached: `numeric(5)[]` is stored as `numeric(5,0)[]`.

An earlier revision also folded `integer[3]` and `integer[][]` to `integer[]`. That is dropped. PostgreSQL accepts those for SQL standard compatibility and enforces neither, so a column declared that way has already lost what it was trying to say, and none of the sample schemas declares one. LIMITATIONS records the drift that leaves.

## Dropping a partition emits a `DROP CONSTRAINT` for an inherited key

A partition carries a copy of its parent's foreign key, and PostgreSQL refuses to drop that copy on its own:

```
ERROR: cannot drop inherited constraint "logs_user_id_fkey" of relation "logs_2024"
```

The copy goes with the partition anyway. `diffForeignKeys` already skipped an inherited key; the drop path does too now. Dropping the whole partitioned table failed the same way, and the parent's own key still goes out there.

This one reaches a dump. Removing one partition from `pista dump` output is enough, since the copy is on the current side rather than in the file.

## A domain's base type is compared as written

`format_type` writes a base type on the search path unqualified, so a domain written over `public.status` compares `status` against `public.status` and reads as a base type change, which cannot be altered:

```
pista: error: failed to diff domains: cannot change base type of domain public.status_dom
```

It goes through `equalTypeName` now, the way a column's type does. Both objects are in the same schema, so this is not the cross-schema case in LIMITATIONS.

## An inherited collation is read as a declared one (BREAKING)

A domain or a composite attribute typed over a collated domain inherits that collation without declaring it, and the catalog stores declared and inherited alike. Reading the inherited one as declared made the desired side, which writes none, look like a collation change. A domain's collation cannot be altered, so the plan stopped, and nothing else in the schema could be planned either.

Both queries compare against the base type's collation now, as `columns.go` already did. Checked against a domain that sets its own collation, one that overrides its base type's, and one that only inherits: the first two are still reported.

`dump` stops writing the inherited clause, which is what `pg_dump` does. The old dump round-tripped only because it wrote the same redundant clause on both sides.

That is what breaks. A file an earlier `pista dump` wrote restates the inherited clause, and the catalog reports none for it now, so the file reads as a collation change:

```sql
CREATE DOMAIN public.ctext2 AS ctext COLLATE pg_catalog."C";
```

A domain fails with `cannot change collation of domain`, which stops the whole plan; a composite attribute drifts on every run instead. Running `pista dump` again rewrites the file, and the new output plans clean. Only a domain or an attribute over a collated type is affected. `plan/domain_restating_inherited_collation.yml` pins the failure.

## A directive on a file with CRLF line endings is silently ignored

The carriage return was taken as trailing content, so `ignore`, `concurrently`, `bulk-alter` and `execute` turned off with no warning: the name is still a known one, so `validateDirectives` raised nothing either. A table marked `-- pista:ignore` was created as if the line were not there.

Git for Windows converts LF to CRLF on checkout by default, and this repository sets no `.gitattributes`, so a schema file committed with LF reaches a Windows working tree as CRLF. pistachio already reads the SQL in such a file correctly; only the directives fell out.

## Tests

Ten fixtures and three unit test groups. Each was checked by reverting its fix:

- `plan/no_diff_alter_table_add_primary_key.yml`, with a composite key so a fix reaching only the first column fails it
- `plan/no_diff_serial_without_not_null.yml`, over all three serial spellings
- `plan/no_diff_numeric_precision.yml`
- `plan/drop_partition_with_inherited_foreign_key.yml`, `apply/drop_partition_with_inherited_foreign_key.yml`, `apply/drop_partitioned_table_with_foreign_key.yml`
- `plan/no_diff_domain_over_schema_qualified_enum.yml`
- `plan/no_diff_domain_over_collated_domain.yml`, covering a domain and an attribute that inherit, one that overrides and one that declares its own, so a predicate reporting too little fails it as well as one reporting too much
- `plan/domain_restating_inherited_collation.yml`, the shape the old dump wrote
- `TestDirectivesAcrossLineEndings`, seven directive forms against four line endings, plus `TestValidateDirectivesAcrossLineEndings` for the argument check; reverting the patterns fails 13 of its cases
- `TestNormalizeTypeName`, `TestSplitTypeSuffix`, `TestFillNumericScale` through `export_test.go`, and `TestDiffTables_dropTableWithInheritedForeignKey` over both drop policies

CI measures coverage per package, so the root package fixtures credit nothing to `parser` and `diff`; the unit tests cover that code where it lives. Every changed line is covered, and project coverage is 96.70%, up from 96.69% on `main`.

`make lint`, `make vet`, `make test`, `make test-scenario` and `make test-fidelity` pass against PostgreSQL 16.

## Also

`LIMITATIONS.md` gains two entries this work turned up: the array dimension drift left by the narrowed type fix, and `PRIMARY KEY USING INDEX`, which does not converge at all and takes the `DROP NOT NULL` with it.

https://claude.ai/code/session_018iZFWSeiGCSNGi6zT4R81Q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 * Read a `serial` column as `NOT NULL`. The type expands to a `NOT NULL` column, and the catalog reports it as one, so a declaration that left the words off planned `DROP NOT NULL` right after the table was created. Applying it took the constraint away.
 
-* Read a column type the way PostgreSQL stores it. A `numeric` with only a precision gains a scale of zero, and an array forgets its dimensions and bounds, so `numeric(5)`, `integer[][]` and `integer[3]` each re-issued the same `ALTER` on every plan.
+* Read a `numeric` written with only a precision. PostgreSQL stores `numeric(5)` as `numeric(5,0)` and `format_type` prints both digits, so the declaration and the catalog never compared equal and every plan re-issued the same `ALTER`.
 
 * Drop a partition that carries a copy of its parent's foreign key. The copy was dropped first, which PostgreSQL rejects with `cannot drop inherited constraint`, and the copy goes with the partition anyway. Dropping the whole partitioned table failed the same way.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-* Read a primary key added by `ALTER TABLE` as implying `NOT NULL`. PostgreSQL sets the flag whichever way the key arrives, but only the inline form was read that way, so a schema file that adds the key in a later statement planned `DROP NOT NULL` on every run, and applying it failed with `column is in a primary key`.
+* Read a primary key added by `ALTER TABLE` as implying `NOT NULL`. PostgreSQL sets the flag whichever way the key arrives, and only the inline form was read that way. A file that adds the key in a later statement planned `DROP NOT NULL` on every run, and applying it failed with `column is in a primary key`.
 
 * Read a `serial` column as `NOT NULL`. The type expands to a `NOT NULL` column, and the catalog reports it as one, so a declaration that left the words off planned `DROP NOT NULL` right after the table was created. Applying it took the constraint away.
 
@@ -10,11 +10,11 @@
 
 * Drop a partition that carries a copy of its parent's foreign key. The copy was dropped first, which PostgreSQL rejects with `cannot drop inherited constraint`, and the copy goes with the partition anyway. Dropping the whole partitioned table failed the same way.
 
-* Compare a domain's base type the way a column's type is compared. `format_type` writes a base type on the search path unqualified, so a domain written over `public.status` was read as a base type change and every plan failed rather than reporting no changes.
+* Compare a domain's base type the way a column's type is compared. `format_type` writes a base type on the search path unqualified, so a domain written over `public.status` read as a base type change, which cannot be altered, and every plan failed.
 
-* Read only the collation a domain or a composite attribute sets itself. One typed over a collated domain inherits that collation, and reading the inherited one as declared made the desired side look like a collation change, which cannot be altered, so every plan failed.
+* Read only the collation a domain or a composite attribute sets itself. One typed over a collated domain inherits that collation without declaring it, and reading the inherited one as declared made the desired side, which writes none, look like a collation change. A domain's collation cannot be altered, so every plan failed. `dump` no longer writes the inherited clause either, which is what `pg_dump` does.
 
-* Read a `-- pista:` directive on a file written with CRLF line endings. The carriage return was taken as trailing content, which turned `ignore`, `concurrently`, `bulk-alter` and `execute` off silently, since the name is still a known one and nothing was reported.
+* Read a `-- pista:` directive on a file written with CRLF line endings. The carriage return was taken as trailing content, so `ignore`, `concurrently`, `bulk-alter` and `execute` turned off silently: the name is still a known one, so nothing was reported.
 
 * Read `CREATE INDEX IF NOT EXISTS` as the index it describes. The clause was kept in the stored definition, which `pg_get_indexdef` never writes, so an index a schema file spelled that way was dropped and created again on every plan.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@
 
 * Compare a domain's base type the way a column's type is compared. `format_type` writes a base type on the search path unqualified, so a domain written over `public.status` read as a base type change, which cannot be altered, and every plan failed.
 
-* Read only the collation a domain or a composite attribute sets itself. One typed over a collated domain inherits that collation without declaring it, and reading the inherited one as declared made the desired side, which writes none, look like a collation change. A domain's collation cannot be altered, so every plan failed. `dump` no longer writes the inherited clause either, which is what `pg_dump` does.
+* **BREAKING**: Read only the collation a domain or a composite attribute sets itself. One typed over a collated domain inherits that collation without declaring it, and reading the inherited one as declared made the desired side, which writes none, look like a collation change. A domain's collation cannot be altered, so every plan failed. `dump` no longer writes the inherited clause either, which is what `pg_dump` does.
+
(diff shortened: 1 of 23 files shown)


winebarrel (commented):
Reviewed a6ae241. The seven fixes each read correctly and I could not fault any of them on their own terms. One thing about the collation change is worth a decision before this merges; everything else below is confirmation.

## The inherited collation fix breaks a schema file an earlier `pista dump` wrote

The new catalog reading is the right one, and matching `pg_dump` is clearly correct. The transition is the problem. As the description says, the old dump round-tripped only because it wrote the same redundant clause on both sides, which means the clause is sitting in files that earlier versions produced. Those files no longer plan.

For a domain the failure is fatal, not drift. Given a database holding

```sql
CREATE DOMAIN public.ctext AS text COLLATE "C";
CREATE DOMAIN public.ctext2 AS public.ctext;
```

the previous version dumped `ctext2` with the inherited clause restated, so the file on disk reads `CREATE DOMAIN public.ctext2 AS public.ctext COLLATE pg_catalog."C";`. The catalog now reports no collation for it, the file declares one, and `diffDomain` turns any mismatch into an error rather than a statement:

```
cannot change collation of domain public.ctext2: PostgreSQL does not support this
```

`DiffDomains` returns that error, so the whole plan stops and nothing else in the schema can be planned either, which is the same failure mode this PR fixes coming from the other direction.

A composite attribute takes the softer half. The same restated clause gives perpetual drift instead:

```
ALTER TYPE public.addr ALTER ATTRIBUTE zip TYPE ctext COLLATE pg_catalog."C";
```

Applying it leaves the catalog reading unchanged, so it comes back on every plan.

`no_diff_domain_over_collated_domain.yml` covers a domain and an attribute that inherit, one that overrides and one that declares its own, but not one that restates what it inherits, which is exactly the shape the old dump wrote.

Re-running `pista dump` fixes an affected file, so this may not be worth code. But a CHANGELOG line saying that a file written by an earlier version can carry a redundant `COLLATE` on a domain or attribute over a collated type, and that re-dumping clears it, would save someone a confusing hard failure. Your call whether that is enough or whether the equality check should accept a desired collation that matches the base type's.

## Confirmed

- The primary key helper reaches every key column through `ALTER TABLE`: a composite key marks both columns and leaves the rest nullable. `PRIMARY KEY USING INDEX` carries no column list and marks nothing, which matches the LIMITATIONS entry.
- `splitTypeSuffix` always rejoins to its input, so the refactor is a no-op for everything but `numeric`, including the `timestamp(3) with time zone` shape whose suffix carries trailing text. `fillNumericScale` only fires on an all-digit modifier, so `geometry(Polygon,4326)` and a qualified `myschema.numeric(5)` are left alone.
- Routine parameters strip the modifier after `deparseTypeName`, so filling the scale does not reach them and `numeric(5)` still compares as `numeric` there.
- The `\r?` patterns are safe against the with-args checks because Go's `\S` excludes `\r`, so a bare CRLF directive is not read as carrying an argument. `TestDirectivesAcrossLineEndings` runs through `validateDirectives`, so its trailing-space-plus-CRLF case already proves it.
- The inherited foreign key skip matches what `diffForeignKeys` already did at both of its drop sites. Inherited CHECK constraints need no equivalent: a partition child returns early from `diffTable` before constraints are diffed, and the dropped-table loop only emits foreign key drops.
- The new domain query's join to the base type is inner, but `typtype = 'd'` guarantees a non-zero `typbasetype`, so no domain is dropped from the result.

`go vet` and the `parser`, `diff` and `model` package tests pass locally. CI is green on PG 15 through 18, samples, fuzz and CodeQL; only `windows` was still running when I looked.

---

winebarrel (commented):
Re-reviewed 38e10b8. LGTM.

Documenting the break rather than softening the comparison is a reasonable call: re-running `pista dump` clears an affected file, and only a domain or an attribute over a collated type can be one. The CHANGELOG note says which shape breaks, how it fails on each side, and what to run, and `**BREAKING**:` matches the marker the entry at 1.x already uses.

`domain_restating_inherited_collation.yml` pins the shape rather than just describing it. Its `desired` writes the base type unqualified and the collation as `pg_catalog."C"`, which is what the old dump wrote, so the fixture reaches the collation check through the base type fix in this same PR rather than stopping short of it. The plan harness compares with `assert.Contains`, so the shortened message in the fixture matches the full `failed to diff domains: ... : PostgreSQL does not support this`.

The composite attribute half is documented but not pinned. Its failure is drift rather than a stopped plan, so leaving it to the CHANGELOG is defensible; a fixture asserting the repeated `ALTER TYPE ... ALTER ATTRIBUTE` would close the pair if you want it later.

Nothing in Go changed since my last pass, so the earlier verification stands. CI on this head is green for build, fuzz, CodeQL, keywords and ascii, with the PG matrix, samples and windows still running.

---

PR #550, opened 2026-09-09
Add --explain to plan

winebarrel:
`plan --explain` (`$PISTA_EXPLAIN`) writes a comment before each statement that reads or rewrites a table that already holds data, so a plan pasted into a review says which lines cost time in proportion to the table.

```
-- rewrite, blocks reads and writes: public.events (~20000 rows, 872 kB, 2 indexes rebuilt)
ALTER TABLE public.events ALTER COLUMN amount SET DATA TYPE numeric(12,2);
-- scan, blocks reads and writes: public.events (~20000 rows, 872 kB)
ALTER TABLE public.events ALTER COLUMN note SET NOT NULL;
-- scan, blocks writes: public.orders (not analyzed), public.customers (not analyzed)
ALTER TABLE ONLY public.orders ADD CONSTRAINT orders_customer_id_fkey FOREIGN KEY (customer_id) REFERENCES public.customers (id);
```

`rewrite` copies the table into a new file and rebuilds every index on it. `scan` reads every row once. The lock is folded to what a reader or a writer of the table sees: `blocks reads and writes` is ACCESS EXCLUSIVE, `blocks writes` SHARE or SHARE ROW EXCLUSIVE, `blocks nothing` SHARE UPDATE EXCLUSIVE or weaker.

A statement that only changes the catalog takes no comment, whatever lock it holds for the instant it runs, and neither does one on a table the same plan creates. A partitioned table shows the sum of its partitions and their number, an `INHERITS` parent its own rows plus its children's. A foreign key names the referenced table next to the referencing one, and a domain change names every table with a column of the domain.

## How it decides

The classification is a fixed table applied to the parse tree of the emitted statement, drawn from the `ALTER TABLE` reference and `tablecmds.c`. The diff is not involved, so no existing code path changes.

Three catalog reads back it, each skipped when the plan has nothing that needs it:

- `pg_class.reltuples` / `relpages` for the row and byte estimates, TOAST included. These are what VACUUM and ANALYZE last wrote, so the comment costs no read of the table itself; a table neither has visited reads as `not analyzed`.
- `pg_cast` for whether a column type change is a binary-coercible relabel or a conversion.
- `pg_proc` for whether an `ADD COLUMN` default calls a volatile function.

The rewrite table was checked against a live PostgreSQL 16 by watching `relfilenode`, which corrected two entries: `varchar` to `text` rebuilds nothing, and `ADD GENERATED AS IDENTITY` on an existing column is not a rewrite.

## Known gaps

- A binary-coercible type change that still changes an index's operator class (`int4` to `oid`) rebuilds its indexes but takes no comment. Rare enough to leave out of the table.
- An `ALTER TABLE` without `ONLY` on an `INHERITS` parent is treated as recursing even for `ADD FOREIGN KEY` and `ADD PRIMARY KEY`, which do not. pistachio writes foreign keys with `ONLY`, so no emitted statement reaches this.

## Tests

19 plan fixtures under `testdata/plan/explain_*.yml` and Go tests for the pure helpers. `go test -p 1 ./...` passes against PostgreSQL 16, `go vet` and `golangci-lint` are clean. The sample schemas from `make schema` were not exercised; the CLI output was checked against a small hand-built schema instead.

https://claude.ai/code/session_01SASkfVcTb6EwcnZuTYRNY2

---
_Generated by [Claude Code](https://claude.ai/code/session_01SASkfVcTb6EwcnZuTYRNY2)_


explain.go
```diff
@@ -0,0 +1,829 @@
...
+
+	eff := explainEffect{touch: touchScan, block: blockWrites}
+	for key, t := range ex.current.All() {
+		for _, col := range t.Columns.CollectValues() {
+			typ := strings.TrimSuffix(col.TypeName, "[]")
+			if typ == domain || typ == bare {
+				eff.addTarget(ex.target(key, t, t.Partitioned))
+				break
+			}
+		}
```
winebarrel, line 719: `current` here is `CurrentTables` before the filters, so it holds every partition, and a partition keeps its inherited columns (`TestTables_partitionChildKeepsInheritedColumns`). For a partitioned table with a column of the domain this loop adds the parent with `recurse = t.Partitioned` (rows summed over the partitions, `N partitions`) and then each partition again on its own line, so the comment counts every partition twice, e.g.

```
-- scan, blocks writes: public.events (~4 rows, 16 kB, 2 partitions), public.events_2024 (~2 rows, 8192 bytes), public.events_2025 (~2 rows, 8192 bytes)
```

Skipping `t.IsPartitionChild()` in the loop keeps the parent line alone. An INHERITS child is fine as is, since the parent is not recursed into there and each side reports its own rows. None of the `explain_alter_domain*` fixtures has a partitioned table, so a small one would pin this down.

---

docs/guides/explaining-plans.md
```diff
@@ -0,0 +1,52 @@
...
+A plan says what will run, not what it costs. `--explain` writes a comment before each statement that reads or rewrites a table that already holds data.
+
+```
+$ pista plan --explain schema.sql
+-- rewrite, blocks reads and writes: public.events (~120000000 rows, 9629 MB, 5 indexes rebuilt)
+ALTER TABLE public.events ALTER COLUMN amount SET DATA TYPE numeric(12,2);
+-- scan, blocks writes: public.orders (~2000000 rows, 210 MB), public.customers (~50000 rows, 6280 kB)
+ALTER TABLE ONLY public.orders ADD CONSTRAINT orders_customer_id_fkey FOREIGN KEY (customer_id) REFERENCES public.customers (id);
+-- scan, blocks nothing: public.events (~120000000 rows, 9629 MB, 24 partitions)
+CREATE INDEX CONCURRENTLY events_at_idx ON public.events USING btree (at);
```
winebarrel: Nit: PostgreSQL does not support `CREATE INDEX CONCURRENTLY` on a partitioned table (`cannot create index on partitioned table "events" concurrently`), so this line of the example cannot come out of a real plan. A plain `CREATE INDEX` on the partitioned table, or a concurrent one on a regular table, would show the same thing.

---

explain.go
```diff
@@ -0,0 +1,829 @@
...
+		for _, f := range defaultFuncsOf(parsed[i]) {
+			if !seenFunc[f] {
+				seenFunc[f] = true
+				funcs = append(funcs, f)
+			}
+		}
+	}
+
+	var err error
+	if ex.stats, err = cat.TableStats(ctx); err != nil {
```
winebarrel: Nit, optional: `TableStats` runs for any plan with at least one statement, including one that only creates or drops tables and gets no comment at all. The PR description says each of the three reads is skipped when nothing needs it; the guide says that only of the last two, which is what the code does. It is one cheap query, so fine either way, but classifying first and reading `pg_class` only when some effect has a target would make the description true as well.

---

catalog/stats.go
winebarrel, line 1: Nit, optional: every other reader in this package has a happy-path test next to it (`TestTables`, `TestDomains`, ...). The three here are exercised only through the root `explain_*` fixtures plus the closed-connection cases, and since `make test` runs without `-coverpkg`, codecov shows this file at 0% patch coverage. A `catalog/stats_test.go` with one case per function would match the package convention and make the report honest.

---

winebarrel (commented):
Reviewed the whole diff on `3b3aaf5`. The design is sound: the classification stays on the parse tree of the emitted statement, the diff is untouched, and the three catalog reads happen after the diff and before the `search_path` change, so the type names resolve the way they were printed. I checked the rewrite table against `tablecmds.c` and the ALTER TABLE reference and found nothing wrong: the widening cases (`varchar(n)` up, `numeric` more precision at the same scale, `timestamp` more precision, any type to no modifier, binary-coercible relabel without a modifier), the rewrites (`text[]`, narrowing, constrained domain, volatile default, serial, identity, generated, `SET LOGGED`), and the lock folding all match. Rename directives, INHERITS children, partition sums, the `NOT VALID` skips and the "table this plan creates" case are each covered by a fixture. Build, vet and the CI matrix are green.

One thing to fix before merge, in an inline comment: `ALTER DOMAIN` on a domain used by a partitioned table lists the parent with its partitions summed and then every partition again, so the sizes double up. The other three comments are nits (a doc example that PostgreSQL would reject, `pg_class` read on plans that get no comment, and happy-path tests for the new catalog readers) and can go either way.

LGTM once the `ALTER DOMAIN` partition case is handled.

---

explain.go
winebarrel: Confirmed and fixed in de5d6d4. A new fixture, `explain_alter_domain_partitioned.yml`, reproduced it exactly as you described:

```
-- scan, blocks writes: public.events (~3 rows, 16 kB, 2 partitions), public.events_2024 (~2 rows, 8192 bytes), public.events_2025 (~1 row, 8192 bytes)
```

The loop now skips `t.IsPartitionChild()`, so the parent line stands alone. An INHERITS child is untouched, for the reason you give: its parent is not summed, so each side reports its own rows and nothing is counted twice.

---

docs/guides/explaining-plans.md
winebarrel: Right, the server rejects it. The example had a second problem too: `public.events` was partitioned on that line but not on the rewrite line above it, so the same table read two ways.

The example now uses a plain `CREATE INDEX` on the partitioned table and a concurrent one on a regular table, which shows both lock levels without an impossible statement.

---

explain.go
winebarrel: Taken. The classification does not need the sizes, only the type and volatility answers, so those two reads now come first, the statements are classified, and `pg_class` is read only when some effect has a target. A plan that only creates and drops tables reads nothing extra now, which is what the description claimed.

---

catalog/stats.go
winebarrel: Added `catalog/stats_test.go` with a case per function: an analyzed table, one never analyzed, a partitioned table and its partition, a view that is not a table, the type-change shapes including a domain and an unresolvable name, and volatility by name.

One expectation is worth recording: `pg_cast` holds no row for a type to itself, so `Binary` is false whenever both sides resolve to one base type. `alterTypeTouch` reads `SameBase` first and never reaches `Binary` there, so the behavior is right; the test says so now.

---

winebarrel (commented):
Re-reviewed `de5d6d4`. All four points are addressed:

- `classifyAlterDomain` skips `IsPartitionChild()`, so a partitioned parent is listed once with its partitions summed, and an INHERITS child still gets its own line. `explain_alter_domain_partitioned.yml` pins the output.
- `TableStats` now runs after classification and only when some effect has a target, so a plan that only creates and drops tables reads no more of the catalog than it did before. The guide's "What it reads" paragraph matches the code now.
- The guide example moves `CREATE INDEX CONCURRENTLY` to a regular table and shows the partitioned one with a plain `CREATE INDEX`.
- `catalog/stats_test.go` covers the three readers the way the rest of the package does, including the `-1` rows case, a partition's own entry, the domain and constrained-domain answers, and the by-name volatility lookup.

Build and vet pass locally, and the test jobs for PostgreSQL 15 through 18 are green on this head.

LGTM.

---

PR #557, opened 2026-09-10
Date the size an --explain comment prints

winebarrel:
`--explain` prints the row and byte estimates from `pg_class`, which VACUUM and ANALYZE write, so the numbers say nothing about how old they are. A table loaded months ago and never analyzed since reads as small as it was then.

The comment now carries the date the estimate was written:

```
-- rewrite, blocks reads and writes: public.events (~120,000,000 rows, 9629 MB, as of 2026-09-08, 5 indexes rebuilt)
ALTER TABLE public.events ALTER COLUMN amount SET DATA TYPE numeric(12,2);
```

- The date is the newest of the last vacuum, autovacuum, analyze and autoanalyze times, in the local time zone.
- Over several relations, a partitioned table or an `INHERITS` parent, the oldest of them is printed: the sum is no fresher than its stalest part.
- It is left out when the server no longer has a time, after a statistics reset or a `pg_upgrade`. `CREATE INDEX` writes the row estimate without a time of its own, so an estimate is sometimes newer than the date says; the guide notes this.

## Load

The times come from the `pg_stat_get_last_*_time` functions, not from `pg_stat_all_tables`. The view groups over every table in the database, and a join condition cannot narrow it to the managed schemas first. Measured with `EXPLAIN ANALYZE` on a 2000-table database, five runs each:

| | Execution time |
| --- | --- |
| before | 2.9 - 3.8 ms |
| after, function calls | 3.8 - 4.4 ms |
| `pg_stat_all_tables` joined | 8.9 - 10.4 ms |

The four calls read shared memory for the managed tables alone, and the query still runs only when the plan holds a statement that wants a size.

## Tests

`make test` and `make lint` pass. The plan fixtures write `{{today}}` where the date goes and the harness expands it. `testdata/plan/explain_stats_time_reset.yml` covers a size the server can no longer date, and `catalog/stats_test.go` gains a VACUUM-only case.


winebarrel:
Reviewed the diff and ran `TestTableStats` and the `TestPlan/explain*` fixtures against a local PostgreSQL 16. All pass, including the new `explain_stats_time_reset` case and the VACUUM-only case.

The implementation reads well:

- `GREATEST` over the four `pg_stat_get_last_*_time` calls returns NULL only when all four are NULL, so the `*time.Time` scan maps cleanly to the zero-time `StatsAt`. Using the functions instead of `pg_stat_all_tables` keeps the lookup scoped to the managed tables, and the measurements in the description back that up.
- Taking the oldest time across partitions and `INHERITS` children is the right call for a summed estimate.
- `.Local()` on the rendered date and `time.Now()` in `expandToday` agree on the zone, so the `{{today}}` fixtures are consistent.

One small doc nit, not blocking: `docs/guides/explaining-plans.md` under "What it reads" now says "one read of `pg_class` joined to `pg_stat_all_tables`", but the code deliberately avoids the view and calls the `pg_stat_get_last_*_time` functions (as the comment in `catalog/stats.go` and the PR description say). Worth aligning the wording so the guide does not contradict the code.

LGTM.

---

winebarrel:
Fixed in 3a9bc3b. The guide said `pg_class` joined to `pg_stat_all_tables`, which is what the first draft did; it now reads "one read of `pg_class`, with the vacuum and analyze times the server keeps in its statistics", without naming the view the query avoids.

winebarrel:
Re-reviewed the follow-up commit. It only touches `docs/guides/explaining-plans.md`, and the "What it reads" sentence now describes the read as `pg_class` plus the vacuum and analyze times from the server's statistics, which matches the query in `catalog/stats.go`. No code changes since the last review, so the earlier test results still stand.

LGTM.

---

PR #559, opened 2026-09-10
Read ADD CONSTRAINT ... USING INDEX

winebarrel:
## Problem

`ALTER TABLE ... ADD CONSTRAINT ... UNIQUE USING INDEX` (and the `PRIMARY KEY` form) is the recipe the ALTER TABLE reference gives for adding a constraint without a long lock: build the index with `CREATE UNIQUE INDEX CONCURRENTLY`, then promote it. A desired schema written this way broke in three places:

1. The form never compared equal to the constraint it created. The catalog renders that constraint as `UNIQUE (col)`, so every plan re-issued `DROP CONSTRAINT` and `ADD CONSTRAINT`.
2. The index the constraint takes over was planned as a `DROP INDEX`. Under `--allow-drop index` the drop ran before the `ADD CONSTRAINT`, which then failed with `index "..." does not exist`.
3. `ADD UNIQUE USING INDEX` without a constraint name was auto-named the usual way instead of taking the index's name, which is how PostgreSQL names it.

## Fix

The promotion renames the index to the constraint's name, so once the constraint exists there is nothing left to compare the written index name against. The form is treated as transitional, like `NOT VALID`: an existing constraint of the same name, type and deferral satisfies it, and the file plans clean until it is rewritten the way `dump` writes it. The index a desired `USING INDEX` constraint names is kept out of the index drops and adds, so the recipe works both with and without the `CREATE UNIQUE INDEX` line left in the file. An unnamed form takes the index's name.

`--explain` already classified the statement as catalog-only; only its stale comment changed.

## Testing

- Plan fixtures: promotion, the `PRIMARY KEY` form, the unnamed form, no drift after apply with and without the index declaration left in place.
- Apply fixtures re-plan and require no drift; one reproduces the `dist_id_temp_idx` -> `distributors_pkey` example from the ALTER TABLE reference, including the rename.
- Package-local tests for the parser, and for the diff's equality and index bookkeeping.
- Coverage is at the same level as main (diff 97.7%, parser 94.5%, catalog 93.5%).

Not covered on purpose: applying the transitional file to an empty database fails in PostgreSQL with `index does not exist`. Constraint statements are emitted before index statements, so even a file that keeps the `CREATE UNIQUE INDEX` line cannot provision from scratch; the form is transitional and assumes the index is already there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Unreleased]
+
+* Read a constraint added with `UNIQUE USING INDEX` or `PRIMARY KEY USING INDEX`. The form never compared equal to the constraint it created, so every plan re-issued DROP and ADD, and the index it takes over was planned as a DROP, which ran first under `--allow-drop index` and broke the ADD. The index now stays, the file plans clean, and a constraint written without a name takes the index's name, the way PostgreSQL names it.
+
 ## [1.49.2] - 2026-09-10
 
 * Date the size an `--explain` comment prints, `~120,000,000 rows, 9629 MB, as of 2026-09-08`. It is when a VACUUM or an ANALYZE last wrote the estimate, so a plan says how stale the numbers are. A size the server can no longer date is printed without it.
--- a/diff/tables.go
+++ b/diff/tables.go
@@ -166,7 +166,7 @@ func diffTable(current, desired *model.Table, dc DropChecker) (*tableDiffResult,
 	// auto-inherit them), so they're still diffed here, mirroring how
 	// indexes and FKs work.
 	if desired.IsPartitionChild() {
-		idxResult, err := diffIndexes(current.Indexes, desired.Indexes, dc)
+		idxResult, err := diffIndexes(current.Indexes, desired.Indexes, usingIndexNames(desired.Constraints), dc)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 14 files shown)


winebarrel (commented):
LGTM.

Reviewed the diff against the surrounding code and ran `go vet` plus the `diff`, `parser` and `model` package tests locally; all pass. The database-backed CI jobs were still in progress when I wrote this.

What I checked:

- `sameConstraintDef` only relaxes the comparison when the desired side carries `IndexName`, which only the parser sets, so catalog-vs-file comparisons for every other constraint are unchanged.
- `usingIndexNames` is keyed by bare index name, which matches how `USING INDEX` works (the index must live in the table's schema) and how the index maps are keyed.
- Both `diffIndexes` call sites (partition child and regular table) pass the consumed set, and rename detection is name-based only, so the new form does not interact with it.
- The unnamed form taking the index's name matches the ALTER TABLE reference.
- The fuzz self-diff stays clean: both sides carry the same `IndexName`, and an index declared next to the constraint is either present on both sides or skipped on both loops.

One non-blocking note, since it is already called out as out of scope: for a file that keeps the `CREATE UNIQUE INDEX` line next to the `USING INDEX` constraint, planning against an empty database still emits only the `ADD CONSTRAINT`, because the add loop skips the consumed index. Constraint statements were already emitted before index statements, so this is not a regression, and the transitional-form framing covers it. The PR description's "nothing to synthesize it from" is slightly stronger than the with-index-declared case, but that is a wording nit, not a code issue.

---

winebarrel:
Thanks for the review. Reworded the last paragraph of the description: with the index declared next to the constraint, the definition is there but statement order (constraints before indexes) still keeps the file from provisioning an empty database, so the transitional-form framing stands. No code change.

winebarrel (commented):
Re-reviewed. LGTM.

The head commit is unchanged, so the earlier code review stands. The reworded last paragraph of the description now matches the behavior: with the index declared next to the constraint, the definition is there, but constraints are emitted before indexes, so the file still cannot provision an empty database. That is the right framing for a transitional form.

CI has since finished the database-backed jobs: `test (15)`, `test (16)`, `test (17)`, `test (18)` and `samples` are all green, as are both fuzz targets, build and CodeQL. Only the `windows` job was still running at the time of this comment.

---

PR #565, opened 2026-09-11
Add pista dump --json

winebarrel:
`pista parse` writes a JSON document for a set of schema files, and the schema published for it carries no command in its name because a dump would write the same document. It now does.

```
$ pista dump --json | jq -r '.tables | keys[]'
public.items
public.users
```

## The document

`DumpResult.Document` builds a `parser.ParseResult` out of what the accessors return, so the document type is the one the schema is reflected off rather than a second struct that could drift from it. Two things follow from that:

* `--omit-schema` and the filters reach the JSON the way they reach the SQL, since the accessors already carry them. `--manage-routine` decides whether `routines` holds anything, as it decides whether they are read at all.
* A database holds no `-- pista:execute` statements, so `execute_stmts` is empty. The key is written all the same, which is what makes the one schema describe both commands.

The fields `parse` leaves at their zero value are the ones a database fills: `oid`, a column's `serial_sequence` and `type_storage`, a constraint's `inherited`.

No header precedes the document. The comment naming the connection and the object count would not parse, and the count is in the document itself.

## The flag

`--json` rather than `--format json`. SQL is what `dump` writes, not one option among several, and `--format` sitting next to the existing `--no-format` would let `--format json --no-format` be written.

`--split`, `--sort-by-deps` and `--no-format` lay SQL out and have nothing to change in a JSON document, so they are refused rather than ignored. The check runs before the database is read.

The JSON is written through one function that `parse` shares, so the encoder options, `Deterministic` among them, cannot come apart between the two commands.

## Tests

`cmd/command` covers the document, validation against the committed schema, `--omit-schema`, the three refused flags, the filters with `--manage-routine`, and a schema holding nothing. The root package covers `Document` for every object kind, `--omit-schema`, an empty result and two schemas.

`make test` (12 packages) and `make lint` (0 issues) pass. Coverage is 96.0% overall, unchanged in the root package and up in `cmd/command` (95.5% -> 95.9%); `Document`, `checkJSON`, `writeJSON` and both `Run` methods are at 100%.

Beyond the tests, I dumped all 60 schemas `make schema` loads and validated each against the committed schema: none invalid, with 1443 objects from gitlab and 518 from musicbrainz among them.

## Docs

`docs/reference/commands.md` documents the flag under `dump`, and the parse guide gains a short section saying the same document comes from a database.


winebarrel:
Reviewed `08e733b`. The approach is right: building `Document()` out of the existing accessors means `--omit-schema`, the filters, `--manage-routine` and `--skip-partition-child` reach the JSON without a second code path to keep in step, and sharing `writeJSON` with `parse` keeps `Deterministic` from drifting between the two. Refusing `--split`, `--sort-by-deps` and `--no-format` before the connection is the right call over ignoring them.

I ran this against a real PostgreSQL 16 cluster rather than reading it. `make test` passes for all 12 packages, `go vet ./...` is clean. By hand: `dump --json` validates against the committed schema for a schema with a serial, an identity column, a domain, an enum, a composite type, a standalone sequence, a partitioned parent and child, a restrictive policy, a trigger, a matview and a plain view with a check option; `--omit-schema` empties `schema` and keys the tables bare; `-n public,other` qualifies both; `--exclude 'post*'` drops the table; `routines` is empty without `--manage-routine` and holds `public.noop()` with it; `--skip-partition-child` leaves only the parent; `execute_stmts` is `[]`, not null; and all three refused flags error instantly against an unreachable database, with `--sort-by-deps --no-format` naming both.

Then I checked the claim the guide now makes about the two documents. For each of the 33 schemas under `test/fidelity/schemas`, I loaded it, took `dump --json`, took `parse` of that same dump's SQL, and diffed the two. Every document validated. Two things came out of the diff.

## `storage_params` is `null` from dump and `{}` from parse

`Dump` calls `clearStorageParams` when `--manage-storage-param` is off, which is the default, so `Table.StorageParams` is nil and the document says:

```json
"storage_params": null
```

while `parse` always writes `{}`, since the parser allocates the map. 47 tables and 5 materialized views across the corpus. Both validate, because the field is a pointer and `markNullable` widened it, so nothing fails; it is just the same fact written two ways depending on which command produced the document.

Reading `null` as "this dump did not read them" is defensible and arguably better than `{}`. But nothing says so, and it is not how the other unmanaged case is written: without `--manage-routine`, `routines` is `{}`, not null. So a consumer meets two spellings of "pistachio was not managing this". Worth either a line in the guide or normalizing the two.

## The list of differing fields is not exhaustive

The new section says the fields `parse` leaves at their zero value are `oid`, `serial_sequence`, `type_storage` and `inherited`. Counted over the corpus, the differences are:

| field | count | listed? |
| --- | --- | --- |
| `oid` (all kinds) | 149 | yes |
| `columns/*/type_storage` | 112 | yes |
| `columns/*/storage_type` | 111 | **no** |
| `storage_params` (null vs `{}`) | 52 | **no** |
| `constraints/*/columns` length | 8 | **no** |
| `columns/*/not_null` | 2 | **no** |
| `columns/*/serial_sequence` | 3 | yes |
| `definition`, `default`, `using`, `with_check`, `partition_bound` | 34 | n/a |
| a partition child's inherited columns and constraints | 26 | n/a |

The last two rows are the existing catalog-versus-parser gap (`pg_get_viewdef` reformats, `pg_get_triggerdef` spells a trigger differently, a child's inherited columns are not in the file) and not this PR's business. The first four are.

`storage_type` is the one I would fix: it sits one letter from `type_storage`, which is listed, and it differs on essentially every column. The catalog reports the effective storage for all of them, while the parser fills it only where the file wrote `SET STORAGE`:

```
public.docs.body  dump: storage_type "external"   parse: "external"   (the file writes it)
public.docs.note  dump: storage_type "extended"   parse: ""           (the file does not)
public.docs.id    dump: storage_type "plain"      parse: ""
```

Same shape for a constraint's `columns`, which the catalog fills for a check or exclusion constraint and the parser leaves empty, and for an identity column's `not_null`, which is true in the catalog and unwritten in the file.

An exhaustive list is probably the wrong form for this. Saying that the catalog knows things a file does not state, naming the ones a consumer will actually hit, and pointing at the rendering gap would hold up better than four names that a reader will take as the whole difference.

## Nit

`--split` and `--sort-by-deps` exclude each other through kong's `xor`, so `dump --help` shows it, while `--json` excludes its three through `checkJSON`, which only speaks at run time. `xor` cannot express this one (`--no-format` is fine with `--split`), so the runtime check is the right mechanism; it is just worth knowing that the conflict is invisible in `--help`, where the reference page carries it.

---

winebarrel:
Both points confirmed and fixed in 99b81cd. The nit was already handled in af9b427, which you had not seen.

## `storage_params` null vs `{}`

Reproduced, and normalized to `{}`. You are right that the two spellings are the problem rather than which one is chosen: `routines` is `{}` when `--manage-routine` did not read them, so `storage_params` should be `{}` when `--manage-storage-param` did not.

`clearStorageParams` empties the map instead of dropping it. Nothing downstream can tell: `Table.StorageParamsSQL` and `View.StorageParamsSQL` both test `nil || Len() == 0`, and the diff runs both sides through `nonNilParams` first. `make test` passes, and a test now pins the empty object for a table and a materialized view whose parameters were not read, alongside the same table with the flag on so the empty object is the unmanaged case rather than a relation that carries none.

That also takes 52 of the differences off the table you counted.

## The list was not exhaustive

It was not, and `storage_type` was the bad miss: one letter from `type_storage`, which was listed, and different on essentially every column.

I re-ran your comparison after the `storage_params` fix, counting by field name across the 33 fidelity schemas:

```
  149  oid
  112  type_storage
  111  storage_type
   26  definition
    8  columns
    4  default
    3  serial_sequence
    2  not_null
    2  partition_bound
    1  using
    1  with_check
   26  a partition child's inherited columns and constraints
```

Rather than fix the list, the section now says what the difference is: `parse` fills a field from what the file writes and leaves the rest at zero, while the catalog reports the effective value whether or not anyone wrote it. `storage_type` is named as the clearest case, with the rest of that group after it, and the rendering gap and the partition case get a paragraph of their own. A test pins the `storage_type` claim, since a guide that explains a mechanism should have the mechanism under test.

## The nit, in af9b427

`--json` no longer checks anything at run time. `Split`, `SortByDeps` and `NoFormat` each share a `xor` group with it -- three groups, since a flag can sit in several -- so kong refuses the combination, `--help` shows it, and `--split --no-format` still works because those two share no group. `checkJSON` is gone. The message is kong's own (`--no-format and --json can't be used together`) and the exit code is 80 rather than 1, so the test moved to `cmd/pista`.

7b807a2 writes the rule down: a flag's type, default, allowed values and the combinations it refuses go in its tag, not in code that reads the options back.

`make test` (12 packages) and `make lint` (0 issues) pass.

winebarrel:
LGTM at `99b81cd`. All three points are answered, and the way the first two were answered is better than what I asked for.

## `storage_params`

Emptying the map instead of dropping it (`99b81cd`) is the right half of the choice: `{}` now means the same thing for unread storage parameters as it does for unread routines, and the document no longer spells one fact two ways.

The part worth checking is that `clearStorageParams` is shared with `plan` and `apply` through `diff_all.go:192`, so this is not a dump-only change. It holds, for a reason already in the code: `StorageParamsSQL` and its view twin test `== nil || Len() == 0`, and `storageParamChanges` runs both sides through `nonNilParams`, so nil and empty were already the same input everywhere.

I checked it rather than trusting the reading. Building both `7b807a2` and `99b81cd` and running them over 70 schemas (the fidelity corpus plus every `test/scenario/testdata` file), with `dump` under no flags, `--manage-storage-param`, `--manage-routine --manage-storage-param`, `--no-format` and `--sort-by-deps`, and `plan` of each schema against itself with and without `--manage-storage-param`: **zero SQL differences** across all of it. The CHANGELOG's "The SQL is unaffected" is accurate.

Re-running the document sweep, `storage_params` is gone from the differences entirely, and no document out of 140 (`dump --json` over the same 70 schemas, with and without `--manage-storage-param`) carries a null for it. All 140 validate against the committed schema.

## The guide

Replacing the four-name list with the shape of the difference is the better fix. Everything the corpus still turns up falls under the new text: `type_storage` 112, `storage_type` 111, a partition child's columns 24, a view's `definition` 12, a constraint's `columns` 8, a trigger's `definition` 7, a column's `default` 4, `serial_sequence` 3, `not_null` 2, a policy's `using` and `with_check`, plus `oid` everywhere.

Two that a literal reader might not map onto "a view's definition, a trigger's, a column's `default` and a policy's `using`": a foreign key's `definition` (`REFERENCES users(id)` against `REFERENCES users (id)`, 6) and `partition_bound` (`FOR VALUES WITH (modulus 2, remainder 0)` against `(MODULUS 2, REMAINDER 0)`, 2). The sentence reads as illustrative, so this is only if you want them named.

## The flags

`af9b427` moving the exclusion into three `xor` groups is cleaner than the helper I was only asking you to document, and `TestRun_DumpJSONExclusiveFlags` going through `runCLI` tests the thing users meet rather than a method. Confirmed all three combinations exit 80 with `can't be used together`, and that `--no-format --sort-by-deps` without `--json` still runs, so the new groups did not over-restrict.

One leftover, in the spirit of the AGENTS.md line you just wrote: `--sort-by-deps` still spells "Cannot be used with --split" in its help prose, which its `xor` tag has carried since before this PR. `--json` names none of its three. Either is fine; having one flag do it and not the other is the odd part.

Checks on `99b81cd`: `go vet ./...` clean and `make test` green for all 12 packages against a real PostgreSQL 16.

---

winebarrel:
LGTM at `76726ba`. Nothing left from me.

"Shape" rather than "document" (`880b099`) is the right correction, and it is the word the rest of the PR needed: the two commands cannot write the same document, and the CHANGELOG, the reference page, the contributing page and `Document`'s own comment now all say the thing that is true.

`7d37e4d` went further than the nit. I only pointed at `--sort-by-deps`; taking `--force-index-concurrently`'s "Cannot be combined with --with-tx" out as well applies the AGENTS.md line where it was already being broken before this PR. I checked that nothing was lost with the prose: `--with-tx --force-index-concurrently` and `--disable-index-concurrently --force-index-concurrently` still stop at kong, `--split --sort-by-deps` still does, and `--try-tx --force-index-concurrently` still runs, which is what the old sentence implied by naming only `--with-tx`. The two edited help blocks in `docs/reference/commands.md` match `pista apply --help` and `pista dump --help` byte for byte, so the hand edit did not drift from the output.

The shortened "From a database" (`76726ba`) still lands, because the `Fields` section above it already names `oid`, `serial_sequence`, `type_storage` and `inherited`, so the short version adds the principle and two cases rather than replacing a list with less. I would not take a fourth pass at it.

Re-checked on `76726ba`: `go vet ./...` clean, `make test` green for all 12 packages against PostgreSQL 16, and `dump --json` over `storage_params.sql` still validates and still writes `{}` without `--manage-storage-param` and the four parameters with it. The JSON behavior is untouched since `99b81cd`, which I verified in full then (70 schemas, zero SQL differences against `7b807a2`, 140 documents valid). I also ran the two suites I had skipped: `make test-fidelity` 33 passed, `make test-scenario` all scenarios passed.

One aside, not this PR and not asked for: `docs/reference/commands.md` carries a `--help` block for `plan`, `apply`, `dump` and the root command, `fmt`'s block omits the connection flags the binary prints, and `parse` has no block at all. Both predate this branch.

I could not run `make lint` here (the installed golangci-lint is built with go1.25 and refuses a go1.27 target), and I only tested against PostgreSQL 16, not the 15/17/18 of the matrix. Nothing in this branch reads a version-dependent catalog path, so I would not expect the others to differ, but CI is the word on both.

---

PR #568, opened 2026-09-12
Move the connection options to the commands that connect

winebarrel:
The CLI embedded `pistachio.Options` at the root, so kong offered the connection flags under every command. `fmt` reads no database and used none of them; `parse` uses `--schemas` alone, to qualify unqualified names.

`plan`, `apply` and `dump` now carry `Options` themselves and build the client in `Run`; `parse` carries `--schemas`; `fmt` carries neither.

**BREAKING**: the flags are written after the command as a result.

```bash
pista -n myschema -c postgres://user@host/db dump   # before
pista dump -n myschema -c postgres://user@host/db   # after
```

`-C` / `--config`, `--pager` and `--version` stay at the root, and the environment variables and the config file are unchanged, the config file because the YAML resolver matches on flag name alone.

The Commands page carries the `--help` output of every command in a details block. `parse` had none and `fmt` had one that was hand-trimmed, so both are regenerated here along with the rest.

Verified against a local PostgreSQL 16: `make test`, `make test-scenario`, `make test-fidelity`, `make lint` and `make deadcode` all pass, and a `dump` of the sample schemas plans back clean. One scenario step passed `-c` before the subcommand and is updated.


winebarrel (commented):
LGTM.

Reviewed the full diff and checked it out locally (`go build`, `go vet`, and the database-free tests in `cmd/pista` and `cmd/command` pass; lint and the PostgreSQL suites I leave to CI, which is green on 046b980).

What I verified beyond reading:

- kong still runs both `AfterApply` hooks now that `Options` and `FilterOptions` sit on the same command node: `pista plan -m a=x -m b=x` fails with the duplicate schema-map error and `pista dump --include '/[/'` with the invalid-pattern error, both before any connection is opened.
- `PISTA_CONN_STR` in the environment and `conn-string:` in a `-C` config file are ignored by `fmt` without error, since the YAML resolver's `Validate` collects flag names from every child node. `schemas:` in the config file and `PISTA_SCHEMAS` both reach `parse`.
- The six `--help` blocks in `docs/reference/commands.md` are byte-identical to the current binary's output.
- No remaining flag-before-command invocations in the scenario, sample, fidelity scripts or the docs; the `test/samples/run.sh` and `sample-db.mk` hits are prose comments that stay correct.

The CHANGELOG entry, the README and docs index notices, and the `Run` signature change in the `command` package all match what the code does.

---

PR #570, opened 2026-09-12
Write a table's columns as a JSON array

winebarrel:
`parse` and `dump --json` wrote a table's `columns` as an object keyed by the
column name. The order is the physical column order, which a JSON object does
not promise to keep, and the key is already the column's `name` field, so the
object shape carried nothing the array does not.

```json
"columns": [
  { "name": "z_id", "type": "bigint", ... },
  { "name": "a_col", "type": "text",  ... }
]
```

Only `columns` changes. Everything else an ordered map holds stays an object:
the catalog gives it a name order, and the parser a declaration order, that a
consumer can sort back to from what the document already carries.

## How

The structs are unchanged. `model.JSONMarshalers` holds a json/v2 marshaler for
`*orderedmap.Map[string, *Column]`, the one ordered map that holds columns, and
`writeJSON` passes it, so `parse` and `dump` write the same shape from one
place. None of the 36 call sites that read `Table.Columns` moved.

The JSON Schema reflector answers for the same type with an array, and
`schema-1.0.json` is regenerated. Per the request this breaks 1.0 in place
rather than publishing a 2.0.

## Tests

- `model`: the `Table` golden, and a new case pinning the array and its order.
- `internal/jsonschema`: `columns` reflects as an array of `Column` while
  `indexes` stays an object, and a document with a map where the array belongs
  now fails validation. `parseDocument` marshals with the same marshalers the
  command uses, so schema validation still measures what `parse` writes.
- `cmd/command`: the `parse` and `dump --json` assertions read the array, with
  a `columnsByName` helper for the cases that want one column.

`make vet`, `make lint` and `make test` pass. Coverage holds: `model` 99.8%,
`internal/jsonschema` 92.4% -> 92.5%, `cmd/command` 95.9%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.53.0] - 2026-09-12
+
+* **BREAKING**: `parse` and `dump --json` write a table's `columns` as an array rather than an object keyed by name. The order is the column order, which a JSON object does not promise to keep, and the name is already a field of the column. The published JSON Schema at `schema-1.0.json` is updated in place.
+
 ## [1.52.0] - 2026-09-12
 
 * **BREAKING**: The connection options (`-c` / `--conn-string`, `-d` / `--dbname`, `--password`, `-n` / `--schemas`, `-m` / `--schema-map`, `--search-path`) belong to the command that opens a connection, so they are written after it rather than before: `pista dump -n myschema`, not `pista -n myschema dump`. `fmt` reads no database and no longer takes them at all, and `parse` takes `-n` alone, which it uses to qualify unqualified names. `-C` / `--config`, `--pager` and `--version` stay where they were, and the environment variables and the config file are unchanged.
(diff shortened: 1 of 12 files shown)


winebarrel (commented):
LGTM.

Reviewed the full diff and checked it out locally: `go build`, `go vet`, and the database-free tests in `model`, `internal/jsonschema`, `cmd/command` and `cmd/pista` pass on 36ca8ad. CI is green on the jobs that have finished.

What I verified beyond reading:

- `Table.Columns` is the only `*orderedmap.Map[string, *Column]` field in the model, so the type-keyed marshaler and the reflector's `Column` check cover exactly one property and cannot reshape anything else by accident.
- A nil `Columns` map marshals as `null` through `JSONMarshalers`, which is the second arm of the schema's `oneOf`. Both the parser and the catalog construct the map, so the document never carries that in practice, but the schema and the encoder agree either way.
- The `parse` output for the example in `docs/guides/parsing.md` is byte-identical to the JSON block the guide shows, and no other doc or fixture still shows `columns` as an object.
- `writeJSON` is the single marshal path for `parse` and `dump --json`, and the `jsonschema` tests marshal through the same marshalers, so the validation tests still measure the bytes the commands write. Nothing in non-test code unmarshals the document, so no decoder counterpart is needed.

Keeping `schema-1.0.json` in place rather than publishing a 2.0 is understood as intentional.

One non-blocking note: the CHANGELOG entry is written directly under `## [1.53.0] - 2026-09-12` rather than `## [Unreleased]`. The version is right for the next release, so this only matters if the release step expects to fill in the heading itself.

---

PR #572, opened 2026-09-13
Add pista diff

winebarrel:
Adds `pista diff <current> <desired>`: compares two schema SQL files and prints the DDL that takes the first to the second, without reading a database.

The first file plays the catalog's role, so the output carries the same statements, order and drop policy as `plan`. `pista plan` against an empty database and `pista diff` from an empty file produce identical output (verified against the pagila sample schema, which also self-diffs clean).

- The diff pipeline in `diff_all.go` is split off the catalog fetch (`diffObjects`); `plan` and `apply` behavior is unchanged.
- Routines are read only under `--manage-routine` and a column-owned sequence stays unmanaged, mirroring the catalog side.
- A `-- pista:execute` check SQL cannot be evaluated without a database, so the statement is kept with a note and `apply` decides.
- `--check` exits with code 2 when the diff contains executable changes, like `plan --check`.
- Tests: `testdata/diff/` fixture suite (21 cases, needs no server) plus Go tests for the error paths and the command output. Package coverage: root 97.7% -> 97.8%, cmd/command unchanged.
- Docs: `docs/guides/diffing.md`, the `diff` section and help block in `docs/reference/commands.md`, README and CHANGELOG.

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -1,0 +1,104 @@
+package pistachio
+
+import (
+	"strings"
+
+	"github.com/winebarrel/orderedmap/v2"
+	"github.com/winebarrel/pistachio/model"
+	"github.com/winebarrel/pistachio/parser"
+)
+
+type DiffOptions struct {
+	FilterOptions
+	DropPolicy
+	Current                  string `arg:"" help:"Path to the current schema SQL file."`
+	Desired                  string `arg:"" help:"Path to the desired schema SQL file."`
+	DisableIndexConcurrently bool   `xor:"index-concurrently" env:"PISTA_DISABLE_INDEX_CONCURRENTLY" help:"Ignore CONCURRENTLY opt-ins (directive and inline) and emit plain CREATE/DROP INDEX."`
+	ForceIndexConcurrently   bool   `xor:"index-concurrently" env:"PISTA_FORCE_INDEX_CONCURRENTLY" help:"Force CONCURRENTLY on every CREATE/DROP INDEX, including pure drops."`
+	BulkAlter                bool   `env:"PISTA_BULK_ALTER" help:"Combine consecutive ALTER TABLE actions on the same table into a single statement. FK changes, RENAME, VALIDATE CONSTRAINT, RLS toggles, and skipped DROPs stay separate."`
+	AssumeValidated          bool   `env:"PISTA_ASSUME_VALIDATED" help:"Treat every table constraint, domain constraint, and foreign key as validated: ignore NOT VALID and never emit VALIDATE CONSTRAINT."`
+}
+
+// Diff diffs two schema SQL files without a database: the first file stands in
+// for the current state the catalog gives Plan, the second is the desired
+// state. The output is the DDL that takes the first schema to the second.
+func (client *Client) Diff(options *DiffOptions) (*PlanResult, error) {
+	if err := client.validateSchemas(); err != nil {
+		return nil, err
+	}
+
+	// Not wrapped: the parser's message already names the file.
+	current, err := parser.ParseSQLFilesWithSchema([]string{options.Current}, client.Schemas[0])
+	if err != nil {
+		return nil, err
+	}
+	desired, err := parser.ParseSQLFilesWithSchema([]string{options.Desired}, client.Schemas[0])
+	if err != nil {
+		return nil, err
+	}
+
+	// The current file plays the catalog's role, and the catalog reads only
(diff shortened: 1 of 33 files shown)


diff.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/winebarrel/pistachio/model"
+	"github.com/winebarrel/pistachio/parser"
+)
+
+type DiffOptions struct {
+	FilterOptions
+	DropPolicy
+	Current                  string `arg:"" help:"Path to the current schema SQL file."`
+	Desired                  string `arg:"" help:"Path to the desired schema SQL file."`
+	DisableIndexConcurrently bool   `xor:"index-concurrently" env:"PISTA_DISABLE_INDEX_CONCURRENTLY" help:"Ignore CONCURRENTLY opt-ins (directive and inline) and emit plain CREATE/DROP INDEX."`
```
winebarrel, line 16: This flag does not reach the current side. `clearConcurrentlyDirectives` in `diffObjects` clears `desiredTables` / `desiredViews` only, which was enough for `plan` because the catalog never sets `Concurrently`. With `diff` the current side is a parsed file and can carry `-- pista:concurrently`, and a pure drop takes the flag from the current index (`diff/tables.go`, `useConcurrently := currentIdx.Concurrently`). So the flag whose help promises a plain DROP emits `DROP INDEX CONCURRENTLY`.

Reproduced on this head:

```
-- current.sql
CREATE TABLE public.users (id integer NOT NULL, name text);
-- pista:concurrently
CREATE INDEX idx_users_name ON public.users (name);
-- desired.sql
CREATE TABLE public.users (id integer NOT NULL, name text);

$ pista diff --allow-drop all --disable-index-concurrently current.sql desired.sql
DROP INDEX CONCURRENTLY public.idx_users_name;
```

Passing `filteredTables` / `filteredViews` to `clearConcurrentlyDirectives` as well is a no-op for `plan` and fixes `diff`. A `testdata/diff/disable_index_concurrently_drop.yml` fixture mirroring `force_index_concurrently_drop.yml` would pin it.

Without the flag the same input also emits `DROP INDEX CONCURRENTLY`, which `plan` against a database would not, since the catalog carries no directive. That one reads as what the old file asked for, so I would leave it; only the "same statements as plan" wording in the docs is not exact there.

---

winebarrel (commented):
Reviewed the whole diff and checked it locally on `7fa6f02`.

What I verified:

- `make build`, `go vet ./...` and the new `TestDiff*` suites in the root package and `cmd/command` pass without a database.
- The `pista diff --help` block in `docs/reference/commands.md` is byte-identical to the binary's output.
- Every `init.sql` under `test/scenario/testdata/` self-diffs to `-- No changes` with `--allow-drop all --manage-routine`.
- The `diffAll` split is a pure extraction: the catalog fetch fills the same seven maps and hands them to `diffObjects`, so `plan` and `apply` behave as before. Filters, `--skip-partition-child`, the storage-param clearing and `standaloneSequences` all run inside `diffObjects` or are applied to the current side explicitly, so the file-backed current side gets the same treatment the catalog side does.
- `filterDesiredBySchemas(current, ..., nil)` is right: the schema map is a desired-side concern, and the current names are already on the catalog side.
- The `-- pista:execute` handling mirrors `plan` (execute-first before, plain after, note when a check SQL exists), and `--check` / exit code 2 are wired the same way as `plan --check`.

One thing to fix, left inline: `--disable-index-concurrently` clears the directive on the desired side only, so a `-- pista:concurrently` in the current file still turns a pure drop into `DROP INDEX CONCURRENTLY`. It cannot happen with `plan` because the catalog never sets the flag, but `diff` makes the current side a parsed file. Clearing the current side too is a one-line change and a no-op for `plan`.

Nit, no action needed: the README line says `--check` exits 2 "when the two differ"; suppressed drops alone exit 0, as the guide says. The `plan` line above it makes the same simplification, so this matches.

Otherwise LGTM. The rest reads clean, the fixture suite covers the flags and directives, and the docs match the behavior.

---

diff.go
winebarrel: Fixed in 5deff97. clearConcurrentlyDirectives now runs over the current side as well, and testdata/diff/disable_index_concurrently_drop.yml pins the plain DROP. The concurrently-related plan/apply fixtures still pass, so it is a no-op there.

Agreed on leaving the no-flag behavior: a -- pista:concurrently in the current file reads as what that file asked for. Added a sentence to the guide so the docs say so.

PR #577, opened 2026-09-13
Leave execute statements out of diff

winebarrel:
`pista diff` printed a `-- pista:execute` statement with a note saying its check SQL could not be evaluated, leaving the decision to `apply`. Printing it at all was inferred from `plan` rather than asked for; `diff` now prints the schema DDL alone, and an execute-only difference is `-- No changes` with exit code 0 under `--check`. `apply` runs execute statements as usual.

- Fixtures replaced with ones pinning the exclusion (`execute_excluded.yml`, `execute_first_excluded.yml`).
- Docs updated, and the "same statements as plan" wording in the guide, `commands.md` and the README tightened to "same schema DDL", which stays exact.
- CHANGELOG entry under 1.54.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [1.54.1] - 2026-09-13
+
+* `diff` no longer prints `-- pista:execute` statements. They are not schema state and their check SQL cannot be evaluated without a database; `apply` runs them as usual.
+
 ## [1.54.0] - 2026-09-13
 
 * Add `pista diff`. It compares two schema SQL files and prints the DDL that takes the first to the second, with the same statements and ordering `plan` produces, without reading a database. `--check` exits with code 2 when the diff contains executable changes.
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ pista diff old.sql new.sql         # print the DDL that takes old.sql to new.sql
 pista diff --check old.sql new.sql # exit 2 when the two differ
 ```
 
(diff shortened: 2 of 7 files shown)


winebarrel (commented):
Reviewed the diff and checked it locally on `1a244f8`.

What I verified:

- `make build`, `go vet ./...` and the `TestDiff*` suites in the root package and `cmd/command` pass.
- The code change is a straight removal: `Diff` now returns `result.Stmts` as is, and `HasChanges` follows the same slice, so the `HasChanges == (SQL != "")` invariant the fixture harness asserts still holds. The `parser` and `strings` imports are still in use.
- An execute-only difference behaves as described: `pista diff --check` prints `-- No changes` and exits 0.
- No stale wording is left behind. The only remaining "apply will decide" references are in `plan`, its tests and `directives.md`, where they still apply. The renamed fixtures pin both the plain and the execute-first exclusion.
- The "same statements" to "same schema DDL" tightening in the guide, `commands.md` and the README is exact now, and it also covers the current-side `-- pista:concurrently` case from the previous review.

One optional nit: the guide's `--check` section says exit 2 "when the diff contains executable changes". Since an execute statement is arguably an executable change from the reader's side, one sentence there saying an execute-only difference exits 0 would save a CI user a surprise. The Execute directives section implies it, so this is fine as is.

Note the CHANGELOG entry goes under a dated `1.54.1` heading rather than `[Unreleased]`. Your call as the releaser; mentioning it only because CLAUDE.md asks for `[Unreleased]`.

LGTM.

---

PR #589, opened 2026-09-15
Keep an ARRAY constructor on one line

winebarrel:
## What

`format.parenDepth` counted a parenthesis but not a square bracket, so every
token inside an `ARRAY[...]` constructor carried the depth of the definition
list around it. A comma there read as the end of a column and the line broke
after it. `fmt` wrote that into the user's file and `dump` into its output:

```
CREATE TABLE public.t2 (
    id integer NOT NULL,
    arr text[] DEFAULT ARRAY['x'::text,
    'y'::text],
    CONSTRAINT t2_pkey PRIMARY KEY (id)
);
```

`pg_get_expr` returns the constructor on one line, so the break was pistachio's
own.

## How

`parenDepth` counts `[` and `]` the way it counts `(` and `)`. The commas of a
constructor now sit one level below the ones that separate columns, so
`listElem` no longer matches them. `needsSpace` and `closesUp` already knew the
brackets, so nothing else moves. The change is two `case` arms and a doc
comment.

A break the input itself put inside a constructor is still kept, lined up with
the definition list.

## Testing

Seven cases in `format/format_test.go`: a constructor on one line, a definition
list collapsing around one, a break the input made, a constructor inside
another, a subscript and a dimension, a constructor in a CHECK, and one a
comment breaks. Three fail without the fix; the rest guard the brackets that
are not constructors.

`testdata/dump/array_default.yml` covers the path from the catalog through the
layout. It fails when `format.go` is reverted.

`make test`, `make test-scenario`, `make test-fidelity`, `make lint` (0 issues)
and `go vet` pass against PostgreSQL 16, and `FuzzFormat` ran three minutes
without a failure.

Coverage is 96.0% before and after, unchanged in every package. The two changed
lines were already covered. What is left uncovered in `format` is the error
paths: a `split` that fails after `scan` succeeded, a parenthesis with no match,
a `deparse` that fails on a statement already parsed. Valid SQL does not reach
them.

## Also found, not fixed

`LIMITATIONS.md` gets an entry for a second bug found next to this one.
libpg_query's deparse drops the parentheses of a subscripted `ARRAY`
constructor, so a column declared `DEFAULT (ARRAY[1,2,3])[1]` reads back from
the catalog as `(ARRAY[1, 2, 3])[1]`, deparses to `ARRAY[1, 2, 3][1]`, drifts on
every plan and emits a statement PostgreSQL rejects:

```
ALTER TABLE public.t ALTER COLUMN e SET DEFAULT ARRAY[1, 2, 3][1];
```

It breaks a `pista dump` round trip rather than only drifting, so it is not
`Priority: low`. Closing it means putting the pair back on each of the four
paths that render an expression, which is not the small fix this PR is, and an
array constructor is a constant so `(ARRAY[1,2,3])[1]` is a longer way to write
`1`. Happy to take it in a follow-up.

https://claude.ai/code/session_01XwUxTCudDYx45jEpyUPukD


winebarrel:
LGTM.

I reviewed this against a local PostgreSQL 16 and reproduced every claim in the description.

## The bug and the fix

Confirmed the break is pistachio's own. On `main`, `pista dump` of the table in the description writes:

```
    arr text[] DEFAULT ARRAY['x'::text,
    'y'::text],
```

and on this branch:

```
    arr text[] DEFAULT ARRAY['x'::text, 'y'::text],
```

while `pg_get_expr` returns the expression on one line either way.

Counting `[` and `]` in `parenDepth` is the right place to fix it. `listElem` is the only rule that asks whether a comma separates columns, and depth is what it asks with, so the constructor's commas had to move a level down. Everything else that reads the array survives:

- `matchParen` finds the `)` at `depth[open]`. Square brackets are always balanced in an input that parses, and `Format` fails on a parse error before `render` runs, so the added levels cancel out and the same `)` is still found.
- `viewBodies` looks for `AS` at the statement's base depth, unaffected for the same reason.
- `render`'s indent stack is still pushed only for `(`, so a `[` opens no frame and the hanging indent is unchanged.

The narrow scope is the nice part: no new state, and `needsSpace`/`closesUp` already knew the brackets.

## What I ran

All against PostgreSQL 16.13:

- `make test` - green, 0 failures.
- `make test-scenario` - 11 passed, 0 failed.
- `make test-fidelity` - 33 passed, 0 failed.
- `go vet ./...` - clean.
- `FuzzFormat` for 90s, ~1.7M execs, 241 new interesting inputs, no failures. Idempotency and the token check both hold.

I also checked `testdata/dump/array_default.yml` is a real regression guard: reverting `format/format.go` to `main` while keeping the fixture fails `TestDump/array_default` with exactly the broken layout, so the fixture would catch a relapse.

Spot checks that stay correct: `numeric(10,2)`, `b[1]` in a CHECK, a subscript in an index expression, a `jsonb` literal containing `[1, 2]`, `text[][]`, and a slice `(ARRAY['x', 'y'])[1:2]`.

## On the limitation

Reproduced exactly as documented. The catalog returns `(ARRAY[1, 2, 3])[1]`, the desired side deparses to `ARRAY[1, 2, 3][1]`, the plan never converges, and PostgreSQL rejects the emitted statement:

```
ERROR:  syntax error at or near "["
LINE 1: ...TABLE public.t ALTER COLUMN e SET DEFAULT ARRAY[1, 2, 3][1];
```

So it does break the round trip rather than only drift, and the `LIMITATIONS.md` entry is accurate. Agreed it does not belong in this PR: it is a deparse problem on four separate render paths, and this one is a two-arm layout fix. Documenting it now and taking it separately is the right call.

## One observation, not for this PR

A comma followed directly by `[` loses its space, so a hand-written `ARRAY[[1, 2], [3, 4]]` comes out as `ARRAY[[1, 2],[3, 4]]`. This is pre-existing in `closesUp`, which returns true for `cur.orig == "["` before `needsSpace` ever looks at `cur.space`; I confirmed `main` produces the identical output for `SELECT ARRAY[[1, 2], [3, 4]];`.

It does not reach `dump`, because `pg_get_expr` writes a multidimensional default as nested constructors, not as `[[...]]`:

```
ARRAY[ARRAY[1, 2], ARRAY[3, 4]]
```

That form round-trips clean on this branch with the space intact, and on `main` it was broken across four lines, so this PR strictly improves it. Only `pista fmt` over a hand-written `[[...]]` is affected, which is cosmetic and arguably the corner case the guidelines say not to chase. Noting it only so it is on record.

---

PR #590, opened 2026-09-15
Read the diff sides out of git

winebarrel:
`diff --git` takes a revision range and reads the files on the command line out of the repository, so two revisions of a schema no longer have to be written out with `git show` first.

```bash
pista diff --git HEAD^..HEAD schema.sql
pista diff --git HEAD^ schema.sql                    # revision against the working tree
pista diff --git origin/main...HEAD schema.sql       # what this branch changes
pista diff --check --git origin/main...HEAD schema.sql
```

| Range | Current side | Desired side |
| --- | --- | --- |
| `A..B` | `A` | `B` |
| `A...B` | the merge base of `A` and `B` | `B` |
| `A` | `A` | the working tree |

An omitted side is `HEAD`. Also `-g` and `$PISTA_GIT`.

## How it works

`internal/gitfile` shells out to git rather than reading the object store, so every revision spelling git accepts works, worktrees and submodules included. Both ends of the range are verified with `git rev-parse` first, which separates a revision that does not exist from a file that is not in it; the contents then come from `git cat-file -t` and `git show`. Paths go to git as `rev:./path`, so they resolve against the working directory rather than the repository root.

A file one side does not hold is empty there, so a file added between the two revisions reads as a create and one removed reads as a drop. A path neither side holds is an error, and so is a directory: either would otherwise read as an empty schema and plan a drop of everything.

The two positional arguments become a file list, since a side can now span more than one file. `diff` without `--git` still takes exactly two.

`parser` grows `ParseSQLSourcesWithSchema`, which takes SQL already in memory, so an error keeps the file name it came from:

```
HEAD^:schema.sql:1:8: failed to parse SQL: syntax error at or near "TABEL"
```

The working tree side carries the path alone, as a file on the command line does.

## Tests

`make test`, `make vet`, `make lint` and `make test-scenario` pass. Coverage per package, measured with `make test`:

| Package | before | after |
| --- | --- | --- |
| root | 97.74% | 97.80% |
| `cmd/command` | 95.85% | 96.31% |
| `internal/gitfile` | - | 98.61% |
| `parser` | 94.54% | 94.55% |

Two blocks in the new code stay uncovered: the `diffObjects` error return, which is uncovered on `main` as well, and `git show` failing after `git cat-file -t` reported a blob, which needs the repository to change between two execs.


winebarrel:
Reviewed against a local PostgreSQL 16 and a set of scratch git repositories. The feature works as described and the design is sound. Nothing blocking; one doc suggestion and a few nits below.

## What I verified

Behaviour, all confirmed:

| Case | Result |
| --- | --- |
| `A..B`, `A...B`, bare `A`, omitted sides | correct sides, merge base resolved |
| file added / removed between revisions | create / drop |
| working tree side, file deleted on disk | drop |
| path resolved from a subdirectory | correct |
| absolute path | correct |
| `HEAD@{0}`, tags, branches | accepted |
| worktree, and running inside a submodule | correct |
| `--check` | 2 on changes, 0 on none, 1 on error |
| parse error naming | `HEAD:bad.sql:1:1` on the git side, bare `bad.sql` on the working tree side |
| duplicate path on the command line | clean `duplicate table` error |

Error paths are all distinct and none of them degrades into an empty schema:

```
git revision nope: Needed a single revision
ghost.sql is in neither HEAD^ nor HEAD
sub is not a file in HEAD
not a git repository (or any of the parent directories): .git
```

The "in neither side" guard and the directory check are the two that matter, since either would otherwise plan a drop of everything, and both are covered by tests that say so in their comments.

`--end-of-options` genuinely holds. I tried a revision and a path that begin with `-`, through both `--git X` and `--git=X`:

```
$ pista diff --git='--output=/tmp/pwned..HEAD' schema.sql
pista: error: git revision --output=/tmp/pwned: Needed a single revision
$ pista diff --git='--upload-pack=touch /tmp/pwned2' schema.sql
pista: error: git revision --upload-pack=touch /tmp/pwned2: Needed a single revision
```

Nothing was written in either case. Passing the path as `rev:./path` rather than `rev:path` is the right call too, and `TestRead_RelativeToWorkingDirectory` pins it.

Suites, on PostgreSQL 16.13: `make test` green with 0 failures, `make test-scenario` 11 passed, `go vet ./...` clean. Coverage matches the table: `internal/gitfile` 98.6%, `parser` 94.5%. `make lint` cannot run in my environment either, for the same go1.25 vs 1.27 reason.

`pista diff --help` matches `docs/reference/commands.md` byte for byte, including the new `-g` block. `-g` collides with no other flag, and `current`/`desired` were never accepted as config keys, so `-C` files are unaffected.

The `parser` change is a faithful extraction: `ParseSQLFilesWithSchema` now builds `[]Source` and delegates, with the `-` to `<stdin>` mapping preserved and the span arithmetic unchanged.

## One thing worth a line in the docs

The file list comes from the command line, but the shell expands a glob against the working tree, so a file deleted on the branch is never named and its drop goes unreported:

```
$ git rm schema/b.sql && git commit -m 'drop b.sql'
$ pista diff --git HEAD^..HEAD schema/*.sql
-- No changes

$ pista diff --allow-drop all --git HEAD^..HEAD schema/a.sql schema/b.sql
DROP TABLE public.b;
```

This is shell semantics rather than a bug, and the file list has to come from somewhere. But `--check --git origin/main...HEAD` in CI is the use case the guide promotes, and there it exits 0 and the build passes over a dropped table. The multi-file examples in `diffing.md` and `commands.md` invite a glob. "Every file is read at both ends of the range" already hints at it; one sentence saying the list itself is not, so a file that only one side holds still has to be named, would close it.

## Nits

`ParseRange` verifies the left revision of `A...B` only through `merge-base`, so that one case gets a different wording from the other three:

```
$ pista diff --git 'nope...HEAD' schema.sql
pista: error: git merge-base nope HEAD: Not a valid object name nope
$ pista diff --git 'nope..HEAD' schema.sql
pista: error: git revision nope: Needed a single revision
```

Not wrong, and `TestParseRange_UnknownRevision` passes either way since it only asserts the revision name appears. But the doc comment says the revisions are verified here, and in that branch `verify` instead runs on the merge base's own output, a full SHA that cannot fail. Verifying `left` and `right` before calling `merge-base` would give one wording everywhere and drop a guaranteed-success `git rev-parse`.

`gitSources` builds the current-side name inline as `revs.Current + ":" + path` while the desired side goes through `sourceName`. Equivalent, since `revs.Current` is never empty, but both sides could use the helper.

`objectType` reports every `git cat-file` failure as "the revision does not hold the path". The comment justifies it and the "in neither side" guard catches the case that matters, so I think it is the right trade for the corner cases it avoids chasing. Recording it only because a git failure on one side alone, with the path present on the other, would read as a drop rather than an error.

## Behaviour change to be aware of

The wrong-argument-count case moves from kong to `diffInput`, so the exit code changes:

```
# main
$ pista diff schema.sql
pista: error: expected "<desired>"          # exit 80

# this branch
$ pista diff schema.sql
pista: error: diff takes a current and a desired schema SQL file, got 1 file(s); --git reads them from git instead   # exit 1
```

Unavoidable, since kong cannot express "exactly two unless `--git`", and it reads better and matches the documented "1 on error" for `--check`. `pista diff old.sql new.sql` is untouched, so the CLI stays compatible and the additive CHANGELOG entry is accurate. Mentioning it only in case the `80` is worth a note.

---

winebarrel:
Thanks for the thorough pass. I reproduced all three points and pushed 253503d.

**The glob gap.** Confirmed, and it is the one that mattered:

```
$ git rm schema/b.sql && git commit -m 'drop b.sql'
$ pista diff --allow-drop all --check --git HEAD^..HEAD schema/*.sql
-- No changes
exit=0
$ pista diff --allow-drop all --check --git HEAD^..HEAD schema/a.sql schema/b.sql
DROP TABLE public.b;
exit=2
```

`docs/guides/diffing.md` now says the list itself is not read from git, with the `--check` case spelled out, and `commands.md` carries the short form. I left it at that rather than deriving the list from git: the list has to come from somewhere, and a directive the user writes beats an inference.

**The `A...B` wording.** Fixed. `ParseRange` splits the range first, verifies both endpoints, and only then asks for the merge base, so the rev-parse on the merge base's own output is gone too:

```
$ pista diff --git 'nope...HEAD' schema/a.sql
pista: error: git revision nope: Needed a single revision
$ pista diff --git 'nope..HEAD' schema/a.sql
pista: error: git revision nope: Needed a single revision
```

`TestParseRange_UnknownRevision` now asserts the wording across all five forms rather than just the revision name.

**The `sourceName` nit.** Both sides go through it now.

**`objectType`.** Left as it is, for the reason you gave.

**The exit code.** I did not add a CHANGELOG line. The entry already says `diff` takes a file list, and the exit code of a usage error is not something the docs promise; 80 is kong's, not ours. Say the word if you want it noted.

`make test`, `make test-scenario`, `make vet` and `make lint` are green on the new commit.

winebarrel:
Re-reviewed at `253503d`. All three points are answered, and the `ParseRange` rework is better than what I suggested: pulling the split out into `splitRange` leaves one verification loop that every form goes through, rather than a check bolted onto each branch.

The error wording is uniform now, `HEAD...nosuchrev` included, which the old code did not reach at all:

```
nosuchrev              -> git revision nosuchrev: Needed a single revision
nosuchrev..HEAD        -> git revision nosuchrev: Needed a single revision
HEAD..nosuchrev        -> git revision nosuchrev: Needed a single revision
nosuchrev...HEAD       -> git revision nosuchrev: Needed a single revision
HEAD...nosuchrev       -> git revision nosuchrev: Needed a single revision
```

Good catch adding `HEAD...nosuchrev` to the table test, and asserting the full `git revision nosuchrev` rather than just the name, which is what actually pins the fix.

I re-ran the range matrix to confirm the restructure changed no behaviour: `A..B`, `A...B`, `..B`, `A..`, `A...`, bare `A` against a clean and a dirty working tree, and `..` / `...` alone all give the same answers as before. `--end-of-options` still holds after the reorder, on the right-hand side of a three-dot range too, where verification now catches it before `merge-base` runs at all. `make test` 0 failures, `make test-scenario` 11 passed, `go vet` clean, `--help` still matches `commands.md`.

The `diffing.md` paragraph says it exactly right, and putting `--check` in the example is the part that matters, since that is where it costs someone a green build.

## One thing the fix cost

Verifying before `merge-base` means `nosuchrev...HEAD` no longer reaches the `merge-base` error return, which used to cover it by accident. `internal/gitfile` is 95.9% now, not the 98.61% in the description:

```
ParseRange   93.8%
Read         92.3%
mergeBase    75.0%
```

The uncovered blocks are `gitfile.go:62-64` and `gitfile.go:154-156`, both on that path. It is still reachable, by two roots that share no history:

```
$ pista diff --git main...lonely schema.sql
pista: error: git merge-base main lonely: exit status 1
```

That is a natural test rather than one written for unreachable defensive code, so it seems worth having:

```go
// Two root commits share no history, so git has no merge base to report.
func TestParseRange_NoMergeBase(t *testing.T) {
	initRepo(t)
	commit(t, "main", map[string]string{"schema.sql": "CREATE TABLE t (id int);"})
	git(t, "checkout", "-q", "--orphan", "lonely")
	commit(t, "lonely", map[string]string{"schema.sql": "CREATE TABLE u (id int);"})

	_, err := gitfile.ParseRange("main...lonely")
	require.Error(t, err)
	assert.Contains(t, err.Error(), "git merge-base main lonely")
}
```

I ran it: `internal/gitfile` goes back to 98.6%, which is the number already in the description's table, so that table would be right again with no edit.

While you are there: `git merge-base` writes nothing to stderr when there is no merge base, so `run` falls back to the bare exec error and the message ends in `exit status 1`. Naming both revisions carries most of it already, and it is pre-existing rather than something this commit introduced, so only worth touching if the test above makes it easy.

Everything else from the first pass stands. Nothing blocking.

---

winebarrel:
Right on both counts, pushed c8ee38f.

`TestParseRange_NoMergeBase` is in, close to your version. `internal/gitfile` is 98.7%, so the table in the description holds.

I took the stderr note too, since the test made the path reachable. The exit status 1 is rewritten into what it means, in the same return the other failures take, so the wording is better without a second branch to leave uncovered:

```
$ pista diff --git 'main...lonely' schema.sql
pista: error: git merge-base main lonely: the two share no history
```

The only block left uncovered in the package is `git show` failing after `git cat-file -t` reported a blob, which needs the repository to change between the two execs.

`make test`, `make test-scenario`, `make vet` and `make lint` are green.

winebarrel:
LGTM.

Re-reviewed at `c8ee38f`. The coverage gap is closed and you took the message with it, which turns the last opaque error in the package into something that says what happened:

```
$ pista diff --git main...lonely schema.sql
pista: error: git merge-base main lonely: the two share no history
```

The narrowing is exactly right, and worth spelling out since it is the part that could have gone wrong. `run` returns a plain `errors.New` whenever git wrote to stderr, so `errors.As(err, &exitErr)` is false on that path and git's own words survive untouched. Only the silent exit-1, which is the one status `git merge-base` reserves for "no merge base", gets renamed. Any other failure keeps its message. `assert.NotContains(t, err.Error(), "exit status")` pins the rewrite, and asserting the wrapped `git merge-base main lonely` alongside it pins the part that was already there.

`internal/gitfile` is at 98.7%. The one uncovered block left is `gitfile.go:101-102`, `git show` failing after `cat-file -t` reported a blob, which is the case the description already calls out as needing the repository to change between two execs. Nothing else in the package is unreached.

Re-ran the sweep on this head and nothing regressed:

- All five unknown-revision forms still report `git revision nosuchrev: Needed a single revision`, so the endpoint check still wins over `merge-base`.
- `A..B`, `A...B`, `..B`, `A..`, `A...`, bare `A`, and `..` / `...` alone give the same answers as before.
- `--end-of-options` still holds, including on the right-hand side of a three-dot range.
- `make test` 0 failures, `make test-scenario` 11 passed, `go vet` clean, all on PostgreSQL 16.13.

No further comments. `make lint` is the only thing I could not run, for the go1.25 vs 1.27 reason that is not this PR's.

---

PR #594, opened 2026-09-17
Compare an index column's operator class the way PostgreSQL writes it

winebarrel:
pg_get_indexdef writes an operator class on the search_path without its
schema, and writes the class's options through get_reloptions, which
quotes a value that does not read as an identifier. A file that
qualified the class, or wrote siglen=32 for the catalog's siglen='32',
never compared equal, so the index was dropped and created on every
plan. An exclusion constraint's elements are index elements and take the
same fold.

A default operator class a file writes out still drifts: deciding that
it is the default takes a lookup the diff does not thread.
LIMITATIONS.md covers it, and the COLLATE clause omitted on the same
terms.

Claude-Session: https://claude.ai/code/session_01JijnaPDBfe8mFGJyNMWVot


winebarrel:
Reviewed at `013c568`. **LGTM** - one non-blocking note below.

What I checked:

- `normalizeIndexElem` is reached only from `equalIndexDef` (via `normalizeIndexStmt`) and `equalConstraintDef` (via `normalizeExclusion`), and both parse a fresh tree per call, so the dropped qualifier never leaks into emitted DDL.
- `IndexElem.Opclassopts` is the same `[]*Node` of `DefElem` as `IndexStmt.Options`, so reusing `normalizeStorageParams` is right, and its in-place `SortStableFunc` works through the slice header the caller passes.
- `INCLUDE` columns (`IndexIncludingParams`) cannot carry an operator class or a collation, so leaving them out of the walk is correct.
- The `LIMITATIONS.md` claim about an element with opclass options holds: `pg_get_indexdef_worker` passes `has_options ? InvalidOid : keycoltype` to `get_opclass_name`, so a class carrying options is written even when it is the default, and both sides name it.
- `go vet ./diff/` is clean and the new `TestEqualIndexDef_*` / `TestEqualConstraintDef_*` cases pass here. No database in this environment, so I did not run the three new plan fixtures; `test (15)`, `test (16)`, `test (17)`, `samples` and both fuzz targets are green on CI.

The one gap, which may well be a deliberate "not worth it" under the corner-case rule:

`normalizeStorageParams` folds an `Integer` and a single-name `TypeName` to a `String`, but not a `Float`. That was safe while it saw only an index `WITH` clause, which the doc comment says outright, but an operator class option can take a fractional value: BRIN's bloom classes have `false_positive_rate`, as in `USING brin (a int8_bloom_ops (false_positive_rate = 0.05))`. `get_reloptions` quotes it, so the catalog hands back `false_positive_rate='0.05'` while the file's bare `0.05` parses as a `Float`, and the index is dropped and created on every plan - the same shape as the `siglen` case this PR fixes. Confirmed against this branch with a scratch test:

```go
equalIndexDef(
    "CREATE INDEX idx ON t USING brin (a int8_bloom_ops (false_positive_rate='0.05'))",
    "CREATE INDEX idx ON t USING brin (a int8_bloom_ops (false_positive_rate=0.05))",
) // false
```

Folding a `Float` to its `Fval` next to the `Integer` arm would cover it. If it is not worth covering, the line to amend is "no index storage parameter takes a fractional value" in that comment, since the function now also serves opclass options, where one does.

Wording nit, take it or leave it: "an exclusion constraint's included" in the changelog entry is hard to parse. "an exclusion constraint's elements too" reads more directly.

---

winebarrel:
Both taken, pushed as `8d06da8`.

The `Float` gap reproduces: `TestEqualIndexDef_opclassOptionFractional` fails on `013c568` and passes with a `Float` arm folding `Fval` next to the `Integer` one. Added a differing-value case alongside it so the fold cannot swallow a real change. The comment line you pointed at now reads that a fractional value folds too, and names the BRIN bloom class as the parameter that takes one.

Changelog wording changed to "an exclusion constraint's elements too".

`go vet`, `golangci-lint` and `go test ./diff` are clean; coverage on the package is unchanged at 97.7% with `normalizeStorageParams` at 100%. The three plan fixtures still have not run here, so CI is the first to exercise them.

---

winebarrel:
Re-reviewed at `8d06da8`. **LGTM.**

Both notes are addressed:

- The `Float` arm folds to `Fval`, which is the literal as the file wrote it, and reloptions keep the text as given, so `false_positive_rate=0.05` and the catalog's `'0.05'` now compare equal while `0.02` still differs. Both new cases pass here, and the comment is accurate again.
- The changelog entry reads cleanly now.

`go vet ./diff/` is clean and every check on the new head is green, including `test (18)` and `windows`, which were still running last time.

---

PR #595, opened 2026-09-17
Manage comments on view columns

winebarrel:
`COMMENT ON COLUMN` naming a column of a view or a materialized view was dropped by the parser without a warning, and the catalog did not read one. `apply` never set such a comment and `dump` never wrote it.

The view now keeps its column comments. The diff sets a new or changed one, clears one the file drops, and writes them again after the view is dropped and created. A rename keeps them, and a recreate the drop policy skips leaves them alone.

The fidelity schemas `comments.sql` and `object_comments.sql` gain a view column comment and a materialized view column comment; both fail on main.


winebarrel:
Reviewed at `62c296e`. **LGTM.**

What I checked:

- The catalog's `CROSS JOIN LATERAL` always yields one row, since an ungrouped aggregate does, and both `array_agg`s order by `attnum`, so the two arrays line up and a view without column comments gets two empty arrays rather than dropping out of the join. The `classoid` filter is right for a column description.
- Every path in `DiffViews` lands where the fixtures say. A new view or a recreated one starts from a nil current map, so all desired comments are emitted after the `CREATE`; a denied recreate hits the `continue` before the column comments, so nothing half-applied goes out; a rename compares against the old view under the new key, since `detectViewRenames` clones the map and keeps the pointer. I also ran a rename that needs `DROP` + `CREATE` (materialized, column removed), which the fixtures do not cover: the `DROP` names the old view, the `CREATE` and the re-emitted comment name the new one, and nothing is cleared.
- `CREATE OR REPLACE VIEW` can only append columns, so a comment on a kept column survives and only the diff goes out, which is what `view_column_comment_append_column.yml` asserts.
- `viewColumnCommentStmts` and `CommentSQL` both tolerate a nil map, so the many `model.View` literals in existing tests that leave `ColumnComments` unset keep working, and `null` is allowed by the JSON schema.
- `omit_schema` works through `Ident` skipping the empty schema part after `dump.go` blanks it, so no extra handling was needed there.
- The parser change is placed before the composite type lookup, so a table still wins for a same-named table; and `COLUMN` was already in `commentTargetSupported`, so the warning path is unchanged.
- `go vet ./...` is clean; the `diff`, `model`, `parser` and `gen-json-schema` tests pass here. No database in this environment, so I did not run the fixtures or the fidelity schemas; `fuzz`, `build`, `Analyze` are green on CI and `test (15..18)`, `samples` and `windows` are still running as I write this.

Nothing to change. The fidelity additions are the right place for this, since `pg_dump` is the oracle for whether `dump` writes the comment back.

---

PR #596, opened 2026-09-17
Fold ONLY on a partitioned table's index and leave attached indexes to it

winebarrel:
`pg_get_indexdef` writes `ON ONLY` for every index on a partitioned table, however it was created. A file that wrote the index without `ONLY` dropped and created it on every plan, and the plan also carried a `DROP INDEX` for the index PostgreSQL attached on each partition. `ONLY` now takes no part in the comparison.

An attached index goes away with the parent's, and PostgreSQL rejects a `DROP INDEX` on one. The plan still emitted it, so dropping or changing a partitioned index failed on apply, a `dump`-written file included. With `--allow-drop all` the failure came after the parent's index was already gone. The catalog now marks an attached index from `relispartition`, and the diff emits no `DROP` for it. A rename of one still goes out.

Not covered here: a file in the `dump` form (`ON ONLY` on the parent, a separate `CREATE INDEX` on each partition) creates the parent's index without attaching the partition's, so it stays invalid, and the plan reports no changes.


winebarrel:
Reviewed at `e20d481`. **LGTM** - two non-blocking notes below.

What I checked:

- `Inh = true` is the right direction: pg_query parses a plain `ON t` as `Inh: true` and `ON ONLY t` as `Inh: false`, so forcing it folds `ONLY` away. `normalizeIndexStmt` is reached only from `equalIndexDef`, which parses fresh trees, so the fold never reaches emitted DDL and a file that writes `ON ONLY` still goes out as written.
- `relispartition` on the index's own `pg_class` is exactly "attached to a partitioned index", and the `Attached` skip sits before the drop-policy branch, so a denied drop reports only the parent's index, which `drop_partitioned_index_denied.yml` and `TestDiffIndexes_dropAttached` (with `denyAllDrops`) both pin.
- The rename path is untouched by the skip, since `detectIndexRenames` runs first and `ALTER INDEX ... RENAME` is legal on an attached index. The apply fixture re-plans clean after the parent is recreated without `ONLY`: the new `logs_2025_at_id_idx` comes back attached and is left alone.
- The existing `LIMITATIONS.md` entry from #459 (psql load hits `already exists`) and this PR's "not covered" note (the parent's index stays invalid) do not contradict each other. They differ in statement order: psql runs the dump in file order, so the partition is created after the parent's `ON ONLY` index and gets its copy from `CREATE TABLE ... PARTITION OF`; `pista apply` creates both tables first, so the later `ON ONLY` index clones nothing and the partition's own `CREATE INDEX` is never attached.
- `go vet ./...` is clean; `diff`, `model` and `gen-json-schema` tests pass here. No database in this environment, so I did not run the fixtures; `build`, `ascii` and `keywords` are green and `test (15..18)`, `samples`, `windows` and both fuzz targets are still running as I write this.

Notes, neither blocking:

1. The "not covered" case in the description is a real gap on the `dump` round trip: a `dump`-written file applied to an empty database leaves the parent's index invalid for good, since nothing ever attaches the partition's, and every later plan is clean. It reads as a candidate for its own `LIMITATIONS.md` entry (or a paragraph on the #459 one), with `ALTER INDEX ... ATTACH PARTITION` named as the fix, so it is recorded somewhere other than this PR's text.

2. A definition change on an attached index alone, with the parent's left as it is, now plans a bare `CREATE INDEX` under the existing name, which fails at apply with `relation already exists`. Before this PR the same edit failed on the `DROP`, so nothing regresses, and the edit is a user error since the attached index mirrors the parent's. Just noting that the failure moved rather than went away, in case a one-line mention next to the new docs paragraph is wanted.

---

winebarrel:
Thanks. On the two notes:

1. I ran a `dump`-written file against an empty database on this branch, and the order is not what the review describes. `pista apply` runs each table's indexes right after that table, so the parent's `ON ONLY` index is created before the partition, `CREATE TABLE ... PARTITION OF` makes and attaches the partition's copy, and the next `CREATE INDEX` fails with `relation "c1_v_idx" already exists`. When it stops, the parent's index is valid and the copy is attached. I tried three pairs of names (`p`/`c1`, `logs`/`logs_2025`, `a_parent`/`z_child`) so the sort order of names would not decide it, and all three behaved the same. So `pista apply` hits the same failure as the psql load in #459.

   The parent's index is left invalid in a different case: a database where the partition already exists gets a new index in the `dump` form. The parent's `ON ONLY` index clones nothing, the partition's `CREATE INDEX` is never attached, and every later plan is clean. That is what the "not covered" paragraph means.

   I agree both belong in `LIMITATIONS.md`. I will record them together with the decision on how to handle this, not in this PR.

2. Confirmed: on main the edit fails on the `DROP`, and on this branch on the `CREATE`. An attached index has to match its parent's, so this is an error in the file, and I would leave the docs as they are.

winebarrel:
Thanks for checking both. You are right on 1: I inferred the statement order instead of reading it, and `DiffTables` emits a new table's indexes through `newTableExtras` right after its `CREATE TABLE`, so the parent's `ON ONLY` index comes before the partition and the empty-database apply fails on `already exists` exactly like the psql load in #459. The invalid-parent case is the one you describe, a new index in the `dump` form on a database whose partition already exists. Recording the two together in `LIMITATIONS.md` with the decision, outside this PR, sounds right. Nothing further from me; LGTM stands.

---

PR #602, opened 2026-09-17
Leave out the constructors of a range type

winebarrel:
CREATE TYPE ... AS RANGE creates constructor functions for the range and
its multirange, recorded in pg_depend as part of the type. The routine
reader read them, so dump wrote them as LANGUAGE internal functions and a
desired schema without them planned DROP FUNCTION, which PostgreSQL
refuses. Leave out a routine with an internal dependency, as one an
extension owns already is.

Found while checking the Hatchet schema as a sample.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


winebarrel:
Reviewed at `f9e1b8a`. **LGTM** - one non-blocking note below.

What I checked:

- `deptype = 'i'` with `classid = pg_proc` is the right key. `makeRangeConstructors` and `makeMultirangeConstructors` (`typecmds.c`) record each constructor with `DEPENDENCY_INTERNAL` on the range or multirange type, and I could not think of another path that makes a `pg_proc` row the dependent side of an internal dependency: I/O, cast, operator, handler and aggregate support functions are all on the referenced side with a normal dependency. So the filter takes exactly the constructors, both the range's and the multirange's, and the new `classid` guard is a correctness improvement for the `'e'` case too.
- The catalog test is the right shape: a user function under the constructor's name and a different signature is still read, so the skip is by dependency, not by name.
- The dump fixture is consistent with range types being unmanaged: nothing but the table comes out, and the plan fixture pins that a file without the constructors no longer proposes a `DROP FUNCTION` PostgreSQL would refuse.
- `go vet` is clean here; no database, so I did not run the two fixtures. `test (15..18)`, `samples`, both fuzz targets, `build` and `Analyze` are green on CI; `windows` is still running.

Note, non-blocking:

A file `dump` wrote before this change carries the constructors as `LANGUAGE internal` functions, and the parser still reads those (I checked with `pista diff --manage-routine` on this branch: an empty file against one holding `floatrange(double precision, double precision) ... LANGUAGE internal AS $$range_constructor2$$` plans the `CREATE OR REPLACE FUNCTION`). Since the catalog no longer reads them back, such a file plans that statement on every run, and it fails at apply without superuser. The people it hits are the ones this fix is for, anyone who ran `dump --manage-routine` over a range type. A sentence in the changelog entry saying those lines are to be deleted from an existing file would cover it cheaply; the alternative is the parser leaving out a `LANGUAGE internal` routine the way it leaves out a `SUPPORT` one, with the same warning, since pistachio cannot create one without superuser anyway. Either works; I would not hold the PR for it.

---

winebarrel:
Thanks. Took the changelog route: the entry now says to delete the constructors from a file an earlier `dump` wrote. Skipping `LANGUAGE internal` in the parser would hide a line the user should remove, so I left the parser alone.

PR #613, opened 2026-09-17
Add --explain to dump

winebarrel:
`pista dump --explain` (`$PISTA_DUMP_EXPLAIN`) writes the size of each table and materialized view after its name in the header comment, and the size of each index in a comment above it.

```sql
-- public.users (~1,000 rows, 48 kB, as of 2026-09-17)
CREATE TABLE public.users (
    id integer NOT NULL,
    name text NOT NULL,
    CONSTRAINT users_pkey PRIMARY KEY (id)
);
-- 40 kB
CREATE INDEX users_name_idx ON public.users USING btree (name);
```

The sizes are the `pg_class` estimates `plan --explain` reads, and the row and byte summing is shared with it. Two catalog reads run only under the flag: `TableStats`, which now also reads materialized views, and a new `IndexSizes`.

A partitioned table sums its partitions at any depth, including ones a filter leaves out of the dump, and an index on it sums the indexes on the partitions. An INHERITS parent counts its own rows alone. An index has no date, since VACUUM and ANALYZE write it with the table's. An index a constraint owns is written inside `CREATE TABLE` and has no comment. The estimates are attached before `--schema-map` renames the schemas. `--json` cannot be used with the flag.

Only comments are added, so the dump still plans clean and is a fixed point of `pista fmt`.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


winebarrel:
Reviewed at `54dd9a8`. **LGTM** - one non-blocking note below.

What I checked:

- `sizeDetails` is a faithful extraction: the analyzed and not-analyzed branches keep the same detail order, `renderTarget` appends `rebuilt` and the descendant count after it exactly as before, and the partitioned-parent skip and oldest-date logic are unchanged. `TableStats` taking in `'m'` cannot change `plan --explain`, since `ex.stats` is only ever read by key.
- `IndexSizes` is right for the partitioned case: `pg_partition_tree` accepts a partitioned index, `isleaf` keeps the sum to real `'i'` relations, and a nested partitioned index gets its own sum over its leaves. `GREATEST(relpages, 0)` matches `TableStats`.
- `explainDump` runs before `remapTableSchemas` / `remapViewSchemas` and before the filters, so `--schema-map` still finds the estimates and a filtered-out partition still counts toward its parent; both have tests. `IsPartitionChild()` (not `PartitionOf != nil`) is the right test there, since the explainer's own `children` map includes `INHERITS` children for `ALTER TABLE` recursion, while the dump's documented behavior is that an `INHERITS` parent counts its own rows alone. The materialized view lookup through a throwaway `model.Table{Schema, Name}` works because `FQTN` and `FQVN` build the same key.
- `Size` is `json:"-"` on all three structs and `--explain` is in the `--json` xor group, so the JSON shape and `schema-1.0.json` are untouched; `TestRun_DumpJSONExclusiveFlags` covers the refusal.
- `CLAUDE.md` is a symlink to `AGENTS.md`, so the fixture-field list is updated in the one place it lives.
- `go vet ./...` is clean; `model` tests pass here. No database, so I did not run the eight dump fixtures or the catalog tests; CI's `test (15..18)` and `samples` were still running when I wrote this, with `ascii` and `keywords` green.

Note, non-blocking:

The table sum and the index sum draw the partition boundary differently. `explainDump` sums a partitioned table over the tables the catalog read, so a partition in a schema `-n` does not name is left out, which the guide says. `IndexSizes` sums a partitioned index through `pg_partition_tree` with no schema filter, so the same partition's attached index does count toward the parent index's size. A cross-schema partition is rare, so this is an observation rather than an ask; if it is worth closing, joining `pg_namespace` on the leaf and adding `AND ln.nspname = ANY(@schemas)` inside the `'I'` subquery lines the two up, and otherwise a clause in the guide's sentence ("...but not when it sits in a schema `-n` does not name; an index on it still counts toward the parent's") records it.

---

winebarrel:
Addressed the note in fda9b5e: `IndexSizes` now joins the leaf's namespace and keeps `nspname = ANY(@schemas)`, so a partitioned index sums the same partitions a partitioned table does. `TestIndexSizes` has a case with a partition in another schema.

winebarrel:
Re-reviewed at `fda9b5e`. The schema filter on the leaf sum is right and `TestIndexSizes` pins it with the cross-schema partition; the `1.58.0` heading is just the release. That part is done.

One blocker, not from this push: **`TestDump/explain` and `TestDump/explain_sort_by_deps` fail on PostgreSQL 16, 17 and 18**, which is what the Codecov comment on the first run was reporting. I reproduced it here against a local 16:

```
--- Expected
+++ Actual
-SELECT users.name
+SELECT name
```

and the same for `SELECT users.id` in `explain_sort_by_deps.yml`. The `--explain` comments themselves come out exactly as the fixtures expect; the two fixtures just assert the PostgreSQL 15 spelling of a one-table view body, and 16 and later print it without the table qualifier, so they need the `dump_pg16` / `dump_pg17` / `dump_pg18` variants the other view fixtures carry (`view_column_comment.yml` is the model). The other six explain fixtures have no view body, so they pass on every version. `fda9b5e` did not touch these files, so the run in progress will go red on `test (16)`, `test (17)` and `test (18)`.

Applying this locally makes all eight `TestDump/explain*` cases pass on 16:

```yaml
# testdata/dump/explain.yml, appended
dump_pg16: &dump_pg16plus |
  -- public.logs (not analyzed)
  CREATE TABLE public.logs (
      id integer NOT NULL,
      body text
  );

  -- public.users (~1,000 rows, 48 kB, as of {{today}})
  CREATE TABLE public.users (
      id integer NOT NULL,
      name text NOT NULL,
      CONSTRAINT users_pkey PRIMARY KEY (id)
  );
  -- 40 kB
  CREATE INDEX users_name_idx ON public.users USING btree (name);

  -- public.user_counts (~1 row, 8192 bytes, as of {{today}})
  CREATE MATERIALIZED VIEW public.user_counts AS
  SELECT count(*) AS cnt
     FROM users;
  -- 16 kB
  CREATE UNIQUE INDEX user_counts_cnt_idx ON public.user_counts USING btree (cnt);

  -- public.user_names
  CREATE OR REPLACE VIEW public.user_names AS
  SELECT name
     FROM users;
dump_pg17: *dump_pg16plus
dump_pg18: *dump_pg16plus
```

```yaml
# testdata/dump/explain_sort_by_deps.yml, appended
dump_pg16: &dump_pg16plus |
  -- public.users (~3 rows, 8192 bytes, as of {{today}})
  CREATE TABLE public.users (
      id integer NOT NULL
  );
  -- 16 kB
  CREATE INDEX users_id_idx ON public.users USING btree (id);

  -- public.user_ids (~3 rows, 8192 bytes, as of {{today}})
  CREATE MATERIALIZED VIEW public.user_ids AS
  SELECT id
     FROM users;
dump_pg17: *dump_pg16plus
dump_pg18: *dump_pg16plus
```

Nothing else changes. `TestTableStats`, `TestIndexSizes` and the three Go-level explain tests also pass here on 16. Once those two fixtures carry the variants, LGTM.

---

winebarrel:
Re-reviewed at `50c758b`. **LGTM.**

Swapping the view bodies to `count(*)` is a cleaner fix than the per-version variants I sketched: with no column reference there is nothing for 15 to qualify, so one expected dump holds on every version and the fixtures stay about `--explain` rather than about view deparsing. The `~1 row` update in `explain_sort_by_deps.yml` follows from the new body.

Against a local 16, all eight `TestDump/explain*` cases, `TestDumpResult_Files_Explain`, `TestDump_ExplainSchemaMap`, `TestTableStats` and `TestIndexSizes` pass. On CI, `test (16)`, `test (17)`, `test (18)`, `windows` and `samples` are green on this head; `test (15)` was still running as I wrote this, and the `count(*)` body has no version-specific spelling for it to trip on.

---

PR #629, opened 2026-09-19
Support serial2, serial4, and serial8 type aliases

winebarrel:
## Summary

Add support for PostgreSQL's serial type aliases (`serial2`, `serial4`, `serial8`) by normalizing them to their canonical forms (`smallserial`, `serial`, `bigserial`) during parsing, matching how integer aliases (`int2`, `int4`, `int8`) are already handled.

## Changes

- **Parser**: Added three new type aliases to the `typeAliases` map:
  - `serial2` → `smallserial`
  - `serial4` → `serial`
  - `serial8` → `bigserial`

- **Test fixtures**: Added comprehensive test coverage:
  - Parser normalization test (`serial_aliases.yml`)
  - Plan tests for adding/altering serial columns with aliases (`add_column_serial8.yml`, `alter_column_serial4_to_serial8.yml`, `no_diff_serial_aliases.yml`)
  - Apply test for creating tables with serial aliases (`create_table_with_serial_aliases.yml`)

## Implementation Details

This fix resolves a round-trip fidelity issue where columns defined with serial aliases would be read correctly but then cause spurious diffs on subsequent runs. Without this normalization, a column written as `serial4` would be parsed as-is, but when planning a change to `serial8`, PostgreSQL would refuse the `SET DATA TYPE serial8` command (it requires the base type name). Additionally, `DROP NOT NULL` would be incorrectly planned on every run.

By normalizing these aliases at parse time to match PostgreSQL's canonical forms, the behavior aligns with how integer type aliases are already handled, ensuring `pista dump` output fed back as desired schema plans clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Correctly interprets PostgreSQL `serial2`, `serial4`, and `serial8` aliases as `smallserial`, `serial`, and `bigserial`.
  - Prevents unnecessary data-type changes and repeated `DROP NOT NULL` operations.
  - Correctly handles serial alias additions and widening migrations, including associated sequences.

- **Tests**
  - Added coverage for serial alias parsing, table creation, migration planning, and no-change scenarios.


winebarrel:
LGTM. The three lines are the whole fix, and they land where the rest of the code already expects a canonical name: `parseColumnDef` reads `serialTypes[col.TypeName]` (`parser/parser.go:748`) and the diff keys `serialBaseTypes`, `alterTypeName` and `equalTypeName` the same way, so an alias missed every one of them. That is both bugs the description names. The set is complete too, since `serial2`, `serial4` and `serial8` are all PostgreSQL has.

What I checked around it:

- Case. The map lookup is case-sensitive, so I ran `pista diff` on main with `INT4` and `VARCHAR(10)` against `integer` and `character varying(10)`, which came back `-- No changes`. The type name is already lowercased by the time it reaches the map, so `SERIAL8` is covered as well.
- `fmt`. It keeps the token stream and does not touch spelling, so a file that writes `serial4` still reads `serial4` after formatting. The normalization stays inside the model.
- `explain.go:595` has listed `serial2`, `serial4` and `serial8` all along, which says the parser was the one place that missed them. It reads raw pg_query nodes rather than the model, so leaving the aliases there is right.
- The fixtures. `add_column_serial8` and `alter_column_serial4_to_serial8` expect exactly what `add_column_bigserial` and `alter_column_serial_to_bigserial` expect, which is the way to show the aliases take the same path rather than a path of their own. `testdata/parser` runs from `parser/parser_test.go`, so the package the change is in keeps its coverage.

No apply counterparts needed for the new plan fixtures, as far as I can see: an alias differs only on the input side and is the canonical name by the time anything is emitted, and `apply/alter_column_serial_to_bigserial.yml` already runs those statements. Nothing to add to the docs either, which name no type aliases anywhere.

## Minor

`TestNormalizeTypeName` in `parser/typename_test.go` is where the map's behavior is written out case by case, so `{"serial alias", "serial8", "bigserial"}` would sit well in that table. The parser fixture covers it end to end, so this is a nicety rather than a gap.

PR #630, opened 2026-09-19
Recreate a constraint trigger whose definition changes

winebarrel:
## Summary

PostgreSQL has no `CREATE OR REPLACE CONSTRAINT TRIGGER`. The diff fell back to `DROP TRIGGER` and `CREATE TRIGGER` only when a change crossed the constraint trigger line, so a constraint trigger that stayed one, with a new function say, planned a statement that `apply` then failed on:

```
ERROR: CREATE OR REPLACE CONSTRAINT TRIGGER is not supported (SQLSTATE 0A000)
```

A change to or from a constraint trigger now takes the drop and create path, under `--allow-drop trigger` like the switch already did. With the drop denied, the trigger keeps its current definition and the plan shows only `-- skipped: DROP TRIGGER ...`.

## Changes

- `diff/triggers.go`: recreate when a constraint trigger is on either side of the change, not only when the kind switches.
- `diff/triggers_test.go`: a unit test for a constraint trigger changing its function.
- Fixtures: `plan/constraint_trigger_change`, `plan/constraint_trigger_change_disallowed`, `apply/constraint_trigger_change`. All three failed before the fix.
- `docs/reference/objects.md` and `CHANGELOG.md`.

## Verification

- `go test -p 1 ./...` on PostgreSQL 16, `go vet`, `gofmt`, `make lint` (0 issues), `make deadcode`.
- `test/scenario/trigger.test.sh`: 12 steps pass.
- diff package coverage is 97.7% before and after.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XMS6sseiLmbzPpYZ1JKuEN)_

## Summary by CodeRabbit

- **Bug Fixes**
  - Constraint trigger definition changes are now applied by safely dropping and recreating the trigger when permitted.
  - Changes remain untouched when the configured drop policy does not allow trigger removal.
  - Recreated triggers retain their intended settings, including deferred and per-row behavior.
  - Plans now identify dependencies that prevent dropping referenced views or materialized views.
  - View operations are ordered to preserve valid dependency relationships when needed.


winebarrel:
LGTM. One wording suggestion on the docs sentence, nothing else.

Reproduced the failure on PostgreSQL 15 as well, not only 16:

```
CREATE OR REPLACE CONSTRAINT TRIGGER ct AFTER INSERT ON rev_tmp.t
    DEFERRABLE INITIALLY DEFERRED FOR EACH ROW EXECUTE FUNCTION rev_tmp.g();
ERROR:  CREATE OR REPLACE CONSTRAINT TRIGGER is not supported
```

So the fixtures need no `min_pg`, and the statement is refused outright rather than on some condition.

Going from `!=` to `||` makes the condition a superset of the old one, so nothing that took the `CREATE OR REPLACE TRIGGER` path before changes path now, and only the case that always failed at apply moves. That is the shape a fix like this should have.

`replaceTriggerSQL` can still build the invalid statement, since it sets `Replace` on whatever it is handed, but the call at `diff/triggers.go:134` is its only one and the new condition is the gate in front of it. A guard inside it would be dead code with no way to test it, so leaving the invariant at the condition reads right to me.

I also checked what the wider drop path takes with it, since `DROP TRIGGER` drops a trigger's comment: `COMMENT ON TRIGGER` is not modeled, the parser warns and drops it, so nothing pistachio manages is lost. And the denied branch skips the rename, the drop, the create and the state change alike, which is what the `plan: ""` plus one skipped drop in the disallowed fixture is saying.

The fixtures match the existing `trigger_constraint_switch` trio, and the apply fixture's dump has the same shape as `constraint_trigger_idempotent.yml`. No `applied_sql` needed there, since apply runs the statements against the server and the old statement would fail on its own.

## Minor

In `docs/reference/objects.md`, "a change to or from a constraint trigger" reads as the transition alone, plain to constraint or back, which is the rule that was already there. The case this fixes is a constraint trigger that stays one, so something like "a definition change with a constraint trigger on either side" says it without the reader having to work it out. The CHANGELOG entry already puts it plainly.

PR #631, opened 2026-09-19
Run DROP NOT NULL after the constraint statements

winebarrel:
PostgreSQL refuses DROP NOT NULL while a primary key holds the column,
and the column statements ran before the constraint ones, so dropping a
key and the NOT NULL on its column in one plan failed at apply with
"column is in a primary key". diffColumns now hands the DROP NOT NULL
statements back on their own and diffTable appends them after the
constraint statements. They stay ahead of the index statements, so the
ALTER TABLE run that --bulk-alter merges is still one piece, with the
DROP CONSTRAINT listed before the DROP NOT NULL.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01XMS6sseiLmbzPpYZ1JKuEN

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed PostgreSQL migration failures when removing a primary key and the associated column’s `NOT NULL` constraint in the same change.
  - `DROP NOT NULL` now runs after table constraint changes, ensuring the correct operation order.

- **Tests**
  - Added coverage for standard and bulk alterations involving primary key and `NOT NULL` constraint removal.


winebarrel:
LGTM. Two suggestions at the bottom, neither blocking.

Checked the premise on PostgreSQL 15, one table per case:

```
=== 1. DROP NOT NULL first (the old plan order)
ERROR:  column "id" is in a primary key
=== 2. DROP CONSTRAINT then DROP NOT NULL, separate statements
ALTER TABLE
ALTER TABLE
=== 3. merged, constraint first
ALTER TABLE
=== 4. merged, reversed
ERROR:  column "id" is in a primary key
```

Case 4 is the one I expected to go the other way, since `AT_PASS_DROP` runs ahead of the column passes, but the subcommand order does decide it. So the comment in `plan/bulk_alter_drop_primary_key_and_not_null.yml` is right as written, and the fix is needed in both the plain and the merged form rather than only one.

`dropNotNullSQL`'s guard is the old condition unchanged: `desIsIdent` in `alterColumnSQL` is `desired.Identity.IsIdentityColumn()` (`diff/tables.go:439`), so the statement moves and nothing else about when it is emitted changes. It is also the only table-level emitter left, `diff/domains.go:98` being a domain's own `DROP NOT NULL` with no key over it, so there is no second path still running early.

No existing fixture's order shifts, which I wondered about since a fixture with both a `DROP NOT NULL` and a constraint statement on one table would have had to move. The three that mention both, `alter_column_drop_not_null.yml`, `explain_set_not_null.yml` and `no_diff_alter_table_add_primary_key.yml`, keep the key untouched, so no constraint statement reaches their plan.

This also finishes what `applyPrimaryKeyNotNull` started: that one stopped the spurious `DROP NOT NULL` while the key stays, and this one orders the real one when the key goes. Same server error from two directions, now closed from both.

## Minor

`diffColumns` returns five values now, and the fifth cost about forty `_` edits across the tests. `diffIndexes` and `diffTable` both hand back a result struct instead; a `columnDiffResult` would take the next slot without another sweep. Worth it only if a sixth looks likely.

The `SET NOT NULL` branch now reads as a condition with an empty path:

```go
if current.NotNull != desired.NotNull && !desIsIdent {
	if desired.NotNull {
```

`if !current.NotNull && desired.NotNull && !desIsIdent` says the same thing, since the `else if` below needs both sides NOT NULL and so is unreachable in the case the outer condition currently swallows. The comment already explains the absence either way.

PR #632, opened 2026-09-19
Name unnamed UNIQUE/EXCLUDE constraints with expressions and INCLUDE

winebarrel:
## Summary

PostgreSQL names an unnamed `UNIQUE` or `EXCLUDE` constraint from the index behind it: the key elements, then the `INCLUDE` columns, with an expression element named the way an index element is (`lower(a)` reads `lower`, a cast reads what it casts, a repeated name takes a number). The parser stood `expr` in for every expression and left the `INCLUDE` list out, so `EXCLUDE (lower(a) WITH =)` was `t_expr_excl` where the server has `t_lower_excl`, and `UNIQUE (a) INCLUDE (b)` was `t_a_key` where it has `t_a_b_key`. Every plan then added the constraint under the wrong name and offered to drop the server's.

This affects a hand-written schema that leaves the constraint name off. `dump` writes `CONSTRAINT <name>`, so its output never depended on the naming.

## Changes

- `parser/parser.go`: the index naming already followed the server, so its body is now `chooseIndexElemNames` over element lists, and `constraintNameParts` (formerly `constraintKeyCols`) feeds the constraint's keys, exclusions and `INCLUDE` columns through it. A primary key keeps `{table}_pkey`.
- `parser/parser_test.go`: six naming cases, verified against PostgreSQL 16.
- Fixtures `plan/unnamed_exclusion_expression` and `plan/unnamed_unique_include`: the unnamed constraint matches the server's and the plan is empty. Both failed before the fix.
- `docs/reference/objects.md` and `CHANGELOG.md`.

## Verification

- `go test -p 1 ./...` on PostgreSQL 16, `go vet`, `gofmt`, `make lint` (0 issues), `make deadcode`.
- parser package coverage is 94.6% before and after.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL-compatible naming for unnamed `UNIQUE` and `EXCLUDE` constraints.
  * Constraint names now correctly account for expressions, function calls, casts, and `INCLUDE` columns.
  * Schema planning no longer proposes unnecessary constraint renames when PostgreSQL generates a different name.

* **Documentation**
  * Updated reference documentation to clarify how unnamed constraints are named when `INCLUDE` columns are present.


winebarrel:
LGTM. Two small notes at the end, neither blocking.

Ran every shape the tests name past PostgreSQL 15 and read the names back out of `pg_constraint`:

```
 tbl |       conname
-----+----------------------
 t1  | t1_a_b_key            -- UNIQUE (a) INCLUDE (b)
 t2  | t2_c_b_excl           -- EXCLUDE (c WITH =) INCLUDE (b)
 t3  | t3_lower_excl         -- EXCLUDE (lower(a) WITH =)
 t4  | t4_lower_lower1_excl  -- EXCLUDE (lower(a) WITH =, lower(b) WITH =)
 t5  | t5_a_excl             -- EXCLUDE ((a::text) WITH =)
 t6  | t6_pkey               -- PRIMARY KEY (a) INCLUDE (b)
 t7  | t7_expr_excl          -- EXCLUDE ((a + b) WITH =)
 t8  | t8_a_b_excl           -- EXCLUDE (a WITH =) INCLUDE (b)
```

Every expectation in `parser_test.go` matches, including the two I would not have predicted: the cast resolving to the column under it rather than to its target type, and the numbering in `lower_lower1`. Sharing `chooseIndexElemNames` is what gets those for free, and it is the same list order the server uses, key elements then `INCLUDE`.

The scope looks right too:

- `constraintKeyCols` has one caller and it feeds the name alone. `model.Constraint.Columns` is still built straight from `con.Keys` a few lines down, so the `INCLUDE` columns do not leak into the key list. That one matters, since `applyPrimaryKeyNotNull` reads that field and PostgreSQL leaves a primary key's `INCLUDE` column nullable. Checked `PRIMARY KEY (a) INCLUDE (b)` on the server: `a` is `NOT NULL`, `b` is not.
- The `ALTER TABLE ... ADD CONSTRAINT` path at `parser/parser.go:2130` goes through the same `parseTableConstraint`, so an unnamed `ADD UNIQUE (a) INCLUDE (b)` is named correctly as well.
- A constraint written on a column keeps its own `autoNameColumnConstraint`, where there is no `INCLUDE` list and no expression to get wrong.

## Minor

`constraintKeyCols` no longer returns the columns a constraint keys on: it returns the parts of the name, `INCLUDE` columns and expression names among them. The doc comment says exactly that now, so only the identifier is behind; something like `constraintNameParts` would carry it.

The description says this fixes the `dump` round trip, but `dump` writes `CONSTRAINT <name>` explicitly, so its output never depended on the naming. What was broken is a hand-written schema that leaves the name off, which is what both fixtures do. The CHANGELOG entry already puts it that way.

winebarrel:
LGTM. Both notes are closed and there is nothing else from me.

`constraintNameParts` now says what it returns, and the rename reached all three places: the doc comment, the function and the one call site. Nothing else referred to it, the tests going through the parse path instead. No behavior rides on the commit.

The description carries the scope now too, that a hand-written schema leaving the name off is what this fixes and that `dump` writes `CONSTRAINT <name>` regardless.

PR #634, opened 2026-09-19
Read exclusive-wait: 0 from the config file

winebarrel:
## Summary

YAML reads a bare `0` as a number, and kong transcodes a resolved value into the flag through JSON, so a number reaching a flag that wants text fails. `exclusive-wait: 0` in the config file failed where `--exclusive-wait 0` worked:

```
pista: error: --exclusive-wait: 0 -> *string: json: cannot unmarshal number into Go value of type string
```

`UnsignedDuration` now implements kong's `MapperValue`: `Decode` takes the token and reads a number as its text before the existing `UnmarshalText`. The command line and an environment variable already hand over text, so nothing changes there.

The same class reaches any text flag given a YAML number, `conn-string: 5432` or `schemas: [2024]`, and those stay as they are: `exclusive-wait` is the one flag whose documented value, `0`, is spelled as a number, and the others take a quoted string.

## Changes

- `exclusive.go`: `Decode` on `UnsignedDuration`.
- `config_test.go`: `exclusive-wait: 0` and `exclusive-wait: 5m` from the file, `--exclusive-wait 5m` from the command line, `PISTA_EXCLUSIVE_WAIT=0`, and a negative value, a missing value and a number without a unit refused. The test CLI gains an `apply` command for it.
- `CHANGELOG.md`.

## Verification

- `go test -p 1 ./...` on PostgreSQL 16, `go vet`, `gofmt`, `make lint` (0 issues).
- The built CLI reads a config file holding `exclusive-wait: 0` and proceeds to the connection.
- Root package coverage is 97.9% before and after.


winebarrel:
The fix itself looks right to me, and `MapperValue` is the hook kong means for this. One correction to the scope, then what I checked.

## The other flags are not immune

The description says every other flag is a bool, a string, a list or a map, "which kong decodes from YAML as it is". That is not what I see. kong transcodes a resolved value into the target through JSON whatever the target is, so a YAML number reaching a string or a list fails the same way. On main:

```
$ cat c.yml
conn-string: 5432
$ pista dump --config c.yml
pista: error: --conn-string: 5432 -> *string: json: cannot unmarshal number into Go value of type string

$ cat c.yml
search-path: 0
$ pista dump --config c.yml
pista: error: --search-path: 0 -> *string: json: cannot unmarshal number into Go value of type string

$ cat c.yml
schemas: [2024]
$ pista dump --config c.yml
pista: error: --schemas: []interface {}{2024} -> *[]string: json: cannot unmarshal number into .0 of type string
```

So the class is a YAML scalar that is not a string reaching a flag that wants text, not a type that parses its own value. `exclusive-wait` is the flag where the numeric spelling is the natural one, `0` being a documented value, so fixing that one alone is a fair call. But the CHANGELOG entry and the description read as if the rest cannot hit it, and a schema or table named `2024` in `schemas`, `include` or `exclude` does.

Closing the class instead means rendering a non-string scalar as text in `yamlResolver.Resolve`, element-wise for a sequence. That happens to be safe today because no flag is numeric, every one being a bool, a string, a list, a map or this duration, so it would want to read `flag.Target.Type()` rather than rely on that staying true. Your call which way; if the per-flag hook stays, dropping the "as it is" claim would keep the record straight.

## What I checked

- `UnsignedDuration` is the only type in the tree with its own text parsing, so nothing else needs the same hook today.
- `fmt.Sprint(token.Value)` is what kong's own string decoder does with a token, so a string token is unchanged and a number arrives as its digits. `ParseDuration` accepts `0` without a unit and rejects `30` with "missing unit", which is why the file and the command line now agree on every spelling rather than only on `0`. That agreement is the property worth having, and the two tests for it say so.
- The "missing value" test is the one I would have asked for: implementing `MapperValue` is where a flag can accidentally stop requiring a value, and `--exclusive-wait` with nothing after it still errors.
- An environment variable goes through `Decode` too, where the token is already a string, and `PISTA_EXCLUSIVE_WAIT=0` covers it.
- Adding `Apply` to `commandCLI` does not disturb `TestYAMLConfig_MetaFlagsAreNotConfigurable`, which builds the other CLI through `parseWithConfig`.

One aside, not this PR's business: `Resolve` returns nil for a key written bare, so `exclusive-wait:` with no value is silently not set rather than an error. Every flag behaves that way, `search-path:` included, which I confirmed by watching it fall back to `public`.

winebarrel:
Right, the "as it is" claim was wrong: the class is a YAML scalar that is not a string reaching a flag that wants text, and `conn-string: 5432` and `schemas: [2024]` fail on main the same way. The description now says so. The per-flag hook stays, since `0` is the one value whose numeric spelling is the documented one and the others take a quoted string; closing the class in `Resolve` is a separate change if it is wanted.

---

PR #637, opened 2026-09-19
Warn about CREATE TABLE AS, LIKE clauses, and ALTER on non-tables

winebarrel:
## Summary

Add warnings for three categories of unsupported SQL statements that were previously dropped silently:

1. `CREATE TABLE ... AS` statements (which share a statement type with `CREATE MATERIALIZED VIEW ... AS`)
2. `LIKE` clauses in `CREATE TABLE` statements (which copy columns the parser does not expand)
3. `ALTER INDEX`, `ALTER VIEW`, and `ALTER MATERIALIZED VIEW` statements (which share a statement type with `ALTER TABLE`)

These statements were silently dropped because they shared statement types with supported statements, preventing them from reaching the `ignored unsupported statement` warning. This caused schema drift: tables created with `AS` were absent from the desired schema, and columns copied via `LIKE` clauses planned as `DROP COLUMN`.

## Key Changes

- **`parser.go`**: 
  - Added `warnIgnoredCreateTableLike()` function to warn about `LIKE` clauses in `CREATE TABLE`, rebuilding the warning statement to show only the `LIKE` clauses (since the parser reads the explicitly declared columns)
  - Modified `CREATE TABLE AS` handling to detect and warn about non-materialized-view cases
  - Modified `ALTER TABLE` handling to detect and warn about `ALTER INDEX`, `ALTER VIEW`, and `ALTER MATERIALIZED VIEW` statements that reference non-existent tables
  - Call `warnIgnoredCreateTableLike()` for non-ignored `CREATE TABLE` statements

- **`parser_test.go`**: Added comprehensive test cases covering:
  - `CREATE TABLE ... AS` warning
  - `LIKE` clause warning with explicit columns
  - Multiple `LIKE` clauses in one statement
  - Ignored tables with `LIKE` clauses (no warning)
  - `ALTER INDEX`, `ALTER VIEW`, and `ALTER MATERIALIZED VIEW` warnings

- **`location_test.go`**: Added test verifying the `LIKE` clause warning carries the correct file position

- **`CHANGELOG.md`**: Documented the new warnings

- **`docs/reference/objects.md`**: Updated documentation to clarify which statements are parsed

## Implementation Details

- The `LIKE` clause warning is built from a rebuilt `RawStmt` containing only the `LIKE` clauses, preserving the original statement's location information for accurate error reporting
- `ALTER` statements on non-existent objects only warn if the object type is not `OBJECT_TABLE`, since `ALTER TABLE` on undeclared tables is intentionally silent
- Ignored tables (marked with `-- pista:ignore`) do not trigger `LIKE` clause warnings

## Summary by CodeRabbit

* **New Features**
  * Added warnings for unsupported SQL statements and clauses, including `CREATE TABLE ... AS`, `CREATE TABLE ... LIKE`, and several `ALTER` statements.
  * Unsupported `CREATE TABLE` clauses are now excluded from parsed column definitions.

* **Documentation**
  * Updated supported-object documentation to describe ignored clauses and warning behavior.

* **Tests**
  * Added coverage for warning content, source locations, ignored tables, and unsupported statement handling.


winebarrel:
LGTM. One question at the end about what stays silent, otherwise notes only.

Confirmed the premise on main with the six statements in one file:

```sql
CREATE TABLE public.a AS SELECT 1 AS x;
CREATE TABLE public.b (id integer, LIKE public.a);
ALTER INDEX public.i SET (fillfactor = 70);
ALTER VIEW public.v SET (security_barrier = true);
ALTER MATERIALIZED VIEW public.mv SET (fillfactor = 70);
ALTER TABLE public.typo ADD CONSTRAINT typo_chk CHECK (id > 0);
```

`pista parse` writes nothing to standard error, and the model it prints holds `public.b` with the single column `id`. So `public.a` is gone and the copied columns are gone, with nothing said, which is the drift the description names.

A few things I checked that are less obvious than they look:

- The discriminator in the `ALTER` case is exact rather than a heuristic. Tables, indexes, views and materialized views share one namespace in PostgreSQL, so a name `tables` holds cannot also be an index or a view, and an `ALTER INDEX` naming a declared table is SQL the server itself refuses. "The lookup failed and the object type is not `OBJECT_TABLE`" therefore can only be an `ALTER` on a non-table.
- The rebuilt `RawStmt` works because `ignoredStmtSnippet` deparses the statement while the position comes from the preserved `StmtLocation`, so the warning shows the `LIKE` clauses alone and still points at the real line. A deparse that ever failed falls back to the original region, which is only less precise.
- `break` in the `CREATE TABLE AS` case behaves as the `continue` in the `ALTER` case does, nothing running after the switch inside the loop.
- Nothing this warns about is written by `dump`, so a dump fed back stays quiet. `ALTER VIEW`, `ALTER INDEX` and `ALTER MATERIALIZED VIEW` come out of `model/view.go` and the storage-param diff, which are plan output, and no dump fixture carries one.

## What stays silent

`ALTER TABLE` on a table the file does not declare stays silent, and the comment calls that deliberate. It does match a convention that is already there: `COMMENT ON TABLE` naming an unknown table, an index on an undeclared table, and `ALTER TABLE ... DISABLE TRIGGER` naming an unknown trigger are each tested as silently skipped.

But the argument in the description covers them just as well. `ALTER TABLE public.typo ADD CONSTRAINT typo_chk ...` above says nothing today, and the constraint is then absent from the desired side, so a plan offers to drop the real one. The file-order requirement makes it reachable without a typo, too, since the docs ask for the relation to be declared in an earlier file. Is the silence there deliberate for a reason the shared-statement-type cases did not have, or is it the next one to look at? Not asking for it in this PR either way.

winebarrel:
The silence on `ALTER TABLE` for an undeclared table predates this PR, and the record treats it as a decision rather than an oversight: the 1.31.0 changelog entry that added the ALTER TABLE action warning closes with "An `ALTER TABLE` naming a relation the file does not declare, or one marked `-- pista:ignore`, is still skipped without one", and `TestParseSQL_NoWarnForAlterTableOnUndeclaredRelation` pins it with the CREATE INDEX parallel. I did not find a reason written down beyond that parallel.

The argument in the description does reach it: `ALTER TABLE public.typo ADD CONSTRAINT ...` says nothing, the constraint is absent from the desired side, and the plan offers to drop the real one. A first draft of this PR warned there too, and was cut back when the changelog line turned up. So it is the next one to look at, as its own change, since it would reverse a stated behavior and also covers `COMMENT ON` and `DISABLE TRIGGER` on an unknown target.

---

PR #651, opened 2026-09-20
Name a statement's object from the parse tree, not from a list of prefixes

winebarrel:
orderStatements placed each statement by matching it against a list of
47 statement prefixes. A list has to carry every form a statement can be
written in, and the form it does not carry sorts as unknown in silence.
CREATE UNIQUE INDEX was such a form once, and CREATE UNLOGGED SEQUENCE
still was: the plan ran it before every other create, where the logged
spelling of the same statement ran in the place its dependencies gave
it.

The name now comes from pg_query's parse tree, which takes the spelling
out of the question and drops the hand-written identifier scanner along
with the list. extractFirstIdentifier stays, moved next to
parseMergeableAlterTable, which needs the identifier as the statement
spells it rather than as the tree holds it.

The index statements keep the place they had: a rename and a drop that
name an index alone stay unplaced, so both still run before the
statements that follow them.

orderStatements and fallbackOrder took the two sides of the diff as
fourteen maps, seven pairs a reader counts out by position. They take
one struct per side and one for the diff results, so the call and its
tests name what they pass.

Claude-Session: https://claude.ai/code/session_01N1VuFiTkodFQzqrxnsBHuE

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved schema change ordering based on object dependencies, including unlogged sequences, tables, types, indexes, triggers, policies, and routines.
  - Corrected handling of `ALTER TABLE` statements involving quoted identifiers and escaped quotes.
  - Improved ordering for dropped objects and statements with circular or unknown dependencies.
  - Enhanced recognition of multiline statements, renames, comments, and other supported SQL changes during schema comparisons.


winebarrel:
LGTM. Taking the name from the parse tree is the right trade: the list could only ever hold the spellings someone remembered, and the one it missed sorted as unknown without saying so.

I checked the ordering out locally, building the branch and running `extractObjectName` over the statement forms `diff/` actually emits: `ALTER TABLE ONLY`, a bulk-merged multi-action `ALTER TABLE`, `ATTACH`/`DETACH PARTITION`, `SET LOGGED`, `RENAME` of a table, a column and a constraint, `ALTER TYPE ... RENAME VALUE`, `CREATE INDEX CONCURRENTLY`, `CREATE TABLE ... PARTITION OF`, `COMMENT ON COLUMN`, `DROP TABLE ... CASCADE`, and a statement led by a comment line or blank lines. Every one lands on the key the prefix list gave it, and `DROP INDEX CONCURRENTLY` and `ALTER INDEX ... RENAME` both stay unplaced, as the PR says they do. `go build`, `go vet` and the touched unit tests pass here; CI is green so far on the jobs that have finished.

One thing the description undersells: the name now comes back through `model.Ident`, so it matches the position map whatever quoting the statement carries. The old scanner copied the statement's own spelling and matched only because pistachio writes its DDL through `model.Ident` too.

Two notes, neither blocking and neither a reason to hold the PR.

`ALTER INDEX i SET (...)` parses as an `AlterTableStmt`, not a `RenameStmt`, so it now resolves to the index name and picks up the relation's position from `addIndexPositions`, where the prefix list left it at -1. `diff/rename.go:371` emits only the `RENAME` form, so nothing changes today.

`COMMENT ON CONSTRAINT c ON public.t` comes back as `public.t.c`, which is never a key in either map, so it is unplaced in silence. That is the same miss the PR is removing, arriving from the other side: a name that is returned but cannot match reads no differently from an empty one. Nothing emits a constraint comment today (only `parser_test.go` mentions the form), so this is only worth knowing if `diff/` ever starts to; `OBJECT_TABCONSTRAINT` and `OBJECT_DOMCONSTRAINT` would want `ownerIdent` at that point.

---

winebarrel:
Both notes check out here. Neither changes the plan today, so neither is fixed in code; bf01727 records the second one where the next reader will meet it.

`ALTER INDEX i SET (...)` does resolve to the index name now, and `addIndexPositions` turns that into the relation's position. That is deliberate: the branch that would have kept it unplaced was unreachable, since `diff/rename.go:371` emits the `RENAME` form alone, and an unreachable branch is not worth the line. If a `SET` ever is emitted, the relation's position is where it belongs anyway. The rename and the drop stay unplaced, which is all the comment on `stmtObjectName` claims.

The constraint comment is worse than a name that cannot match. The two forms take different shapes:

```
COMMENT ON CONSTRAINT users_pkey ON public.users         -> "public.users.users_pkey"
COMMENT ON CONSTRAINT dom_check ON DOMAIN public.pos_int -> "dom_check"
```

The table form wants `ownerIdent`, as you say. The domain form carries the domain as a `TypeName` node followed by the constraint name, so `nameIdent` reads the constraint alone and `ownerIdent` would return the empty string rather than the domain. Handling it means unwrapping the type name, for a statement nothing emits, so it stays out and the note says what it would take.

On the name coming back through `model.Ident`: agreed, and it is the part I would keep if the rest were reverted. The old scanner matched by copying the statement's spelling, which held only because `model.Ident` wrote both sides.

---

winebarrel:
Re-reviewed at bf01727. The new commit only adds the note on `commentObjectName`, so the earlier LGTM stands unchanged.

I checked the note against the trees themselves, since both halves of it are claims about shape. `COMMENT ON CONSTRAINT c ON public.t` arrives as `OBJECT_TABCONSTRAINT` with `List[String "public", String "t", String "c"]`, which is exactly what `ownerIdent` is written for. `COMMENT ON CONSTRAINT c ON DOMAIN public.d` arrives as `OBJECT_DOMCONSTRAINT` with `List[TypeName public.d, String "c"]`, so `nameIdent` reads the `TypeName` item as an empty string and the whole name comes back as `"c"` today; `ownerIdent` would hand it the `TypeName` alone and get `""`. The note is right that the domain form wants its own handling rather than the same call, which is the part a reader would otherwise have to find out the hard way.

Neither name is a key in either position map, so both forms sit at -1 either way, and nothing emits them. Build, vet and the touched unit tests pass on this head; CI is still running.

---

